PR #34, opened 2019-07-16
Truncate body and headers in webhook responses

ob-stripe:
### Reviewers
r? @aarongreen-stripe @tomer-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
Truncate webhook responses so the message stays under a certain size.

Messages sent to Stripe over the websocket connection must stay under 32 KiB, otherwise the server will reply with an unclear error message ("websocket frame too large"). This PR ensures we stay under the limit:
- the `body` field is truncated to 5000 bytes
- the `headers` map is limited to 25 entries, with keys (header names) truncated to 50 bytes and values truncated to 200 bytes

Even accounting for some overhead for JSON formatting and additional attributes, this should ensure we stay well under the limit.


tomer-stripe:
lgtm!

PR #35, opened 2019-07-16
Add verbose transport

ob-stripe:
### Reviewers
r? @brandur-stripe @tomer-stripe @aarongreen-stripe 
cc @stripe/dev-platform

 ### Summary
Adds a new `verboseTransport` type that acts like a regular HTTP transport but can optionally dump headers and other information (URL, status code).

This replaces the existing code in `requests/base.go` to output headers.

Behavior changes:
- headers are printed to stderr instead of stdout
- we don't print all headers, but just a few relevant ones
- API keys are redacted
- we also output the HTTP method and URL, and the response's status code

### DIFF
--- a/pkg/stripe/client.go
+++ b/pkg/stripe/client.go
@@ -29,6 +29,10 @@ type Client struct {
 	// empty, the `Authorization` header will be omitted.
 	APIKey string
 
+	// When this is enabled, request and response headers will be printed to
+	// stdout.
+	Verbose bool
+
 	// Cached HTTP client, lazily created the first time the Client is used to
 	// send a request.
 	httpClient *http.Client
@@ -68,7 +72,7 @@ func (c *Client) PerformRequest(method, path string, params string, configure fu
 	}
 
 	if c.httpClient == nil {
-		c.httpClient = newHTTPClient(os.Getenv("STRIPE_CLI_UNIX_SOCKET"))
+		c.httpClient = newHTTPClient(c.Verbose, os.Getenv("STRIPE_CLI_UNIX_SOCKET"))
 	}
 
 	resp, err := c.httpClient.Do(req)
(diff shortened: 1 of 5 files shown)


pkg/stripe/client.go
```diff
@@ -29,6 +29,10 @@ type Client struct {
 	// empty, the `Authorization` header will be omitted.
 	APIKey string
 
+	// When this is enabled, request and response headers will be printed to
+	// stdout.
+	Verbose bool
```
brandur-stripe: Just FYI that once/if you add linting, it'll complain about stuff like this. Generally comments on exported variables/functions are prefixed with the name of the variable/function. e.g. "Verbose enables verbose mode in the client, which causes it to print requests and responses to stdout."

pkg/stripe/verbosetransport.go
```diff
@@ -0,0 +1,80 @@
...
+	"io"
+	"net/http"
+	"regexp"
+	"strings"
+
+	"github.com/stripe/stripe-cli/pkg/ansi"
+)
+
+// inspectHeaders is the whitelist of headers that will be printed.
+var inspectHeaders = []string{
```
brandur-stripe: Good call on the whitelist (as opposed to blacklist) — these really don't change much.

May want to consider adding `Idempotent-Replayed` too.

pkg/stripe/verbosetransport.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, listed := range inspectHeaders {
+		for name, vv := range header {
+			if !strings.EqualFold(name, listed) {
+				continue
+			}
+			for _, v := range vv {
+				if v != "" {
+					r := regexp.MustCompile("(?i)^(basic|bearer) (.+)")
+					if r.MatchString(v) {
+						v = r.ReplaceAllString(v, "$1 [REDACTED]")
```
brandur-stripe: Might be a good idea to make `Authorization` one of the tested headers below, or have some test for this logic otherwise.

pkg/stripe/client.go
ob-stripe: We do have run both `go vet` and `golint`, but neither seem to complain here. Also the Go plugin for VSCode does show inline documentation for structs, but not for struct fields. I wonder if that's related.

brandur-stripe: Oh weird! Okay, I'm probably misremembering then — maybe it's just functions and globals.

ob-stripe:
Made the requested changes, self-approving.

PR #40, opened 2019-07-17
docs: Update supported payment_intent event types

ynnoj:
### Reviewers
r? @ob-stripe 

 ### Summary
As per https://stripe.com/docs/api/events/types, use singular `payment_intent` for event types and remove reference to `payment_intent.updated` as not currently supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ The webhook signing secret provided will not change between restarts to the `lis
 You can specify which events you want to listen to using `--events` with a comma-separated [list of Stripe events](https://stripe.com/docs/api/events/list).
 
 ```sh
-$ stripe listen --events=payment_intents.created,payment_intents.updated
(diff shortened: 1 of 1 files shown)


ynnoj:
Also, I couldn't find reference to a `payment_intent.updated` event. Should this be update to `payment_intent.amount_capturable_updated `?

https://stripe.com/docs/api/events/types

tomer-stripe:
Thanks for catching that, lgtm!

I think the `payment_intent.updated` event was added by mistake. We have't added support for `payment_intent.amount_capturable_updated`. Feel free to remove that or we can do that it :D

ynnoj:
> Thanks for catching that, lgtm!
> 
> I think the `payment_intent.updated` event was added by mistake. We have't added support for `payment_intent.amount_capturable_updated`. Feel free to remove that or we can do that it :D

I'll update the PR 🙃

PR #49, opened 2019-07-19
Embed OpenAPI spec in the CLI

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Embed the OpenAPI spec (specifically, the SDK variant) in the CLI.

This is done using [vfsgen](https://github.com/shurcooL/vfsgen). One cool feature of vfsgen is that it automatically compresses files. The spec is a 2.4 MB file, but embedding it into the CLI only increases the executable's size by ~250 KB.

I've added two new Makefile targets:
- `update-openapi-spec` will update the spec file to the latest version
- `build` will compile a `stripe` binary with the necessary generation steps

(Note that because the `spec` package is not used anywhere yet, merging this PR won't actually increase the size of the binary for now.)


tomer-stripe:
overall lgtm! Super cool you can get the spec to be compressed so much.

We're going to be using this to validate some of the ideas we discussed right?

Makefile
```diff
@@ -1,10 +1,19 @@
+export GO111MODULE := on
+
 all: test
 
 install-deps:
-	GO111MODULE=on go get
+	go get
+.PHONY: install-deps
```
brandur-stripe: Oh man, I have to admit I had no idea why `.PHONY` targets were used until RTFMing just now.

OOC, is there a reason to place the phony target below rather than above the body? Just a stylistic thing? (Asking because I noticed the manual example places above.

ob-stripe: No other reason than me realizing we were missing `.PHONY` targets when I looked at Goreleaser's Makefile and reusing their convention: https://github.com/goreleaser/goreleaser/blob/master/Makefile. If placing the annotation above the body is more common I'm happy to update the PR accordingly!

brandur-stripe: Ah, got it. Thanks!

> If placing the annotation above the body is more common I'm happy to update the PR accordingly!

Don't really mind either way, so whatever works.

brandur-stripe:
LGTM as well.

PR #53, opened 2019-07-23
Add status command

tomer-stripe:
### Reviewers
r? @ob-stripe @aarongreen-stripe @brandur-stripe @auchenberg-stripe 
cc @stripe/dev-platform

 ### Summary
Let people retrieve API status through the CLI. This pulls from our status site so that is the source of truth.

The command is `stripe status` and includes several flags with:
1. `--verbose`, lists out status of individual systems
2. `--format`, let's people render the status as json
3. `--poll`, continuously checks the status site
4. `--poll-rate`, let's people check more frequently than ever 60 seconds (up to 5 seconds per poll)
5. `--hide-spinner`,  hides the spinner that's shown when polling

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ go 1.12
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/briandowns/spinner v0.0.0-20190319032542-ac46072a5a91
-	github.com/coreos/etcd v3.3.10+incompatible
 	github.com/fatih/color v1.7.0 // indirect
 	github.com/golang/protobuf v1.3.1 // indirect
 	github.com/gorilla/websocket v1.4.0
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.12
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/briandowns/spinner v0.0.0-20190319032542-ac46072a5a91
+	github.com/coreos/etcd v3.3.10+incompatible
 	github.com/fatih/color v1.7.0 // indirect
 	github.com/golang/protobuf v1.3.1 // indirect
 	github.com/gorilla/websocket v1.4.0
--- a/pkg/status/status.go
+++ b/pkg/status/status.go
@@ -1,0 +1,126 @@
+package status
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"text/template"
+
+	"github.com/stripe/stripe-cli/pkg/ansi"
+)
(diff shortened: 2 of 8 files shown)


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.12
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/briandowns/spinner v0.0.0-20190319032542-ac46072a5a91
+	github.com/coreos/etcd v3.3.10+incompatible
```
ob-stripe: Was this intended? This isn't listed as an indirect dependency, but I can't find it actually being used anywhere.

tomer-stripe: Hmm, probably not. I'm not sure where that came from to be honest, removing.

tomer-stripe:
@ob-stripe yea that's a good idea. I wonder if we should do something more universal around emoji support that's built as part of the colors flag.

pkg/status/status.go
```diff
@@ -0,0 +1,126 @@
...
+			return "", err
+		}
+
+		return string(data), nil
+	}
+
+	statusString := fmt.Sprintf(`%s %s{{if .statuses}}
+%s API
+%s Dashboard
+%s StripeJS
```
brandur-stripe: Note a big deal, but I wonder if we should use the official product name/case/punctuation here, which I think is "Stripe.js" for StripeJS and just "Checkout" for CheckoutJS.

pkg/status/status.go
```diff
@@ -0,0 +1,126 @@
...
+	Time        string   `json:"time"`
+}
+
+type statuses struct {
+	API        string `json:"api"`
+	Dashboard  string `json:"dashboard"`
+	Stripejs   string `json:"stripejs"`
+	Checkoutjs string `json:"checkoutjs"`
+	Webhooks   string `json:"webhooks"`
+	Emails     string `json:"emails"`
```
brandur-stripe: Not sure if intentional or not, but "emails" doesn't seem to appear in the formatted output below.

pkg/status/status.go
tomer-stripe: Ah, good call! I'll change this.

pkg/status/status.go
tomer-stripe: I intentionally left it out. We don't report emails on status.stripe.com so I didn't want to cause confusion. Maybe I should remove it from here too?

brandur-stripe: Probably not a big deal either way. If you leave it in, I'd add a comment indicating that it's not currently exposed (and may be not reliable? not sure if there's a reason it's not).

PR #56, opened 2019-07-26
Remove `go generate` command from build step

tomer-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Running `go generate ./...` changes the `go.mod` and `go.sum` files, which prevents goreleaser from releasing new versions. I'm removing this for now since it looks like @ob-stripe added it in #49 and I don't think we need it for any existing functionality.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -3,7 +3,6 @@ env:
 before:
   hooks:
     - go mod download
-    - go generate ./...
 project_name: stripe
 builds:
   - main: ./cmd/stripe/main.go
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,6 @@ test: install-deps lint vet
 
 build:
 	go mod download
-	go generate ./...
 	go build -o stripe -ldflags "-s -w" cmd/stripe/main.go
 .PHONY: build
 


brandur-stripe:
I think you know already, but just pointing out that I think this will make OB's spec bundling implementation not work anymore (although I guess you could still run it manually).

OOC, why does `go generate` cause `go.mod` and `go.sum` to change? I wouldn't have expected that.

LGTM if you're blocked.

ob-stripe:
I don't have a problem with merging this for now to unblock new releases, but we need to find a proper fix ASAP. I'm quite confused by how `go mod` is supposed to work in conjunction with `go generate`.

https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module says that the best way is to have a Go file in your code with a build tag to import the generation tool so that it'll be added to `go.mod` but not actually imported by the main code base. In our case, I think `vfsgen.go` should fill that role.

However, it doesn't seem to be detected by `go mod`, and even if I manually add the dependency with `go install github.com/shurcooL/vfsgen`, it will be removed the next time `go mod tidy` is run.

@brandur-stripe Do you have any clue how this is supposed to work?

ob-stripe:
Actually, I think we can close this one in favor of #58.

tomer-stripe:
Closing this in favor of #58

PR #57, opened 2019-07-26
Introduce Log Tailing command

aarongreen-stripe:
### Reviewers
r? @tomer-stripe  @ob-stripe 
cc @stripe/dev-platform

 ### Summary
Users should be able to run `stripe logs tail` and receive API request logs in real-time.

With the anticipation of wanting to add future commands surrounding logs, there are two file structures we can consider.

The first one is continuing to have all commands in the top-level `cmd` dir and prefixed with their corresponding base command (e.g., `logs_tail.go`). I think this is very ugly and would lead to a lot of ugliness down the road.

As a result, in this PR, I propose having subdirectories for each base command, which will contain all of the subcommands for the base command. (e.g., `logs` directory with `tail.go` file for `stripe logs tail`). One consequence of this is that the `Config` object (responsible for profile configuration) initialized in `root.go` will not be in scope in the subcommands (learned this during implementation), and will instead need to be passed as a parameter down through the `AddCommand(NEW_CMD_FUNC(&Config))` process (which would now resemble a tree of commands, which I think is a nice -- and probably the intended -- design).

Additionally, this PR includes filters as flags that users can provide for filtering their request log feeds.

### Testing
Automated tests for what can be tested, manual testing for end-to-end flow.


pkg/cmd/logs/tail.go
```diff
@@ -0,0 +1,95 @@
+package cmd
```
ob-stripe: If it's in a subdirectory, it should be its own package. The problem is that you're already using `logs` as a package name.

Not sure what the best solution is but we might want to go with `cmd/logs_tail.go` for now.

cc @brandur-stripe in case there's a more elegant/idiomatic solution you can think of

pkg/cmd/logs.go
```diff
@@ -0,0 +1,40 @@
...
+func newLogsCmd(config *config.Config) *LogsCmd {
+	logsCmd := &LogsCmd{
+		cfg: config,
+	}
+
+	logsCmd.Cmd = &cobra.Command{
+		Use:   "logs",
+		Args:  validators.NoArgs,
+		Short: "Top-level package for logs commands with Stripe.",
+		Long: fmt.Sprintf(`
```
tomer-stripe: You don't need to use `fmt.Sprintf` here do you? Nothing's getting formatted.

pkg/cmd/root.go
```diff
@@ -21,11 +21,12 @@ var rootCmd = &cobra.Command{
...
-		"post":    "api",
-		"delete":  "api",
-		"trigger": "webhooks",
-		"listen":  "webhooks",
+		"get":       "api",
+		"post":      "api",
+		"delete":    "api",
+		"trigger":   "webhooks",
+		"listen":    "webhooks",
+		"logs tail": "logs",
```
tomer-stripe: I think this needs a template change as well. Does the help command print correctly?

pkg/logs/tailer.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/stripe/stripe-cli/pkg/ansi"
+	"github.com/stripe/stripe-cli/pkg/stripeauth"
+	"github.com/stripe/stripe-cli/pkg/websocket"
+)
+
+const outputFormatJSON = "json"
+
+// LogFilters contains all of the potential user-provided filters for log tailing
+type LogFilters struct {
+	FilterIPAddress      string `json:"filter_ip_address,omitempty"`
```
tomer-stripe: What does `omitempty` do? We don't do it in other places, what do we need it for here?

pkg/logs/tailer.go
```diff
@@ -0,0 +1,199 @@
...
+	FilterHTTPMethod     string `json:"filter_http_method,omitempty"`
+	FilterRequestPath    string `json:"filter_request_path,omitempty"`
+	FilterSource         string `json:"filter_source,omitempty"`
+	FilterStatusCode     string `json:"filter_status_code,omitempty"`
+	FilterStatusCodeType string `json:"filter_status_code_type,omitempty"`
+}
+
+// Config provides the cfguration of a Proxy
```
tomer-stripe: `cfguration` => `configuration`?

Is this something we could share with the webhook listener? It seems similar.

pkg/logs/tailer.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// Output format for request logs
+	OutputFormat string
+
+	// WebSocketFeature is the feature specified for the websocket connection
+	WebSocketFeature string
+}
+
+// Tailer is the main interface for running the log tailing session
+type Tailer struct {
```
tomer-stripe: Same with this, I think the webhook one is identical.

pkg/logs/tailer.go
```diff
@@ -0,0 +1,199 @@
...
+	signal.Notify(tailer.interruptCh, os.Interrupt, syscall.SIGTERM)
+
+	filters, err := jsonifyFilters(tailer.cfg.Filters)
+	if err != nil {
+		tailer.cfg.Log.Fatalf("Error while converting log filters to JSON encoding: %v", err)
+	}
+
+	session, err := tailer.stripeAuthClient.Authorize(tailer.cfg.DeviceName, tailer.cfg.WebSocketFeature, &filters)
+	if err != nil {
+		// TODO: better error handling / retries
```
tomer-stripe: Does this need to be done before we merge in?

pkg/logs/tailer.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	if tailer.cfg.OutputFormat == outputFormatJSON {
+		fmt.Println(ansi.ColorizeJSON(requestLogEvent.EventPayload, os.Stdout))
+		return
+	}
+
+	coloredStatus := colorizeStatus(payload.Status)
+
+	outputStr := fmt.Sprintf("%s [%d] %s %s %s", payload.CreatedAt, coloredStatus, payload.Method, payload.URL, payload.RequestID)
```
ob-stripe: Would be nice to linkify the request ID. Something like this:

```go
url := fmt.Sprintf("https://dashboard.stripe.com/test/logs/%s", payload.RequestID)
ansi.Linkify(payload.RequestID, url, os.Stdout)
```

pkg/logs/tailer.go
```diff
@@ -0,0 +1,199 @@
...
+		"webhook_id": requestLogEvent.RequestLogID,
+	}).Debugf("Processing request log event")
+
+	var payload EventPayload
+	if err := json.Unmarshal([]byte(requestLogEvent.EventPayload), &payload); err != nil {
+		tailer.cfg.Log.Warn("Received malformed payload: ", err)
+	}
+
+	// Don't show stripecli/sessions logs since they're generated by the CLI
+	if payload.URL == "/v1/stripecli/sessions" {
```
tomer-stripe: What would you think if we added a super verbose option that prints these out too? Doesn't need to happen now though.

pkg/websocket/request_log_messages.go
```diff
@@ -3,6 +3,6 @@ package websocket
 // RequestLogEvent represents incoming request log event messages sent by Stripe.
 type RequestLogEvent struct {
 	EventPayload string `json:"event_payload"`
-	Type         string `json:"type"`
 	RequestLogID string `json:"request_log_id"`
```
ob-stripe: Could you add a comment to clarify that this is a `resp_` ID and not the same as `EventPayload.RequestID`?

pkg/logs/tailer.go
aarongreen-stripe: Don't think so, this is also a comment on the Proxy implementation. I copy and pasted and decided to leave it in there. I think it should be addressed for both as a future ticket.

pkg/logs/tailer.go
aarongreen-stripe: We could...but I don't know if ever printing `/v1/stripecli/sessions` logs is useful for the user, since they have no way of triggering it themselves and have no idea it even exists -- it doesn't show up in their dashboard logs I believe.

pkg/cmd/logs/tail.go
aarongreen-stripe: Alternatively, we could rename the `logs` package to something like `logtailing`. Allowing us to name the sub-commands the `logs` package. I don't really like the idea of having sub-commands as top-level commands and think it's a bad pattern to start, even if we have the intention to improve it at some point.

pkg/logs/tailer.go
aarongreen-stripe: It is similar, but I don't like the idea of sharing the configs in case there are things that are special to each (in this case, the `LogFilters`). In a world where more specific configuration is added to each, this gets messy.

pkg/logs/tailer.go
aarongreen-stripe: This is similar as well, but does not contain `endpointClients  []*endpoint.Client`, which the webhook one does contain. Each also has their own specific config. It is similar code, but it's also not much, and for similar reasons to the above case, I think it's safer and cleaner to leave them separate.

pkg/cmd/root.go
aarongreen-stripe: Good catch -- it does not. Will fix this.

aarongreen-stripe:
Feedback should be addressed. ptal @ob-stripe  @tomer-stripe

pkg/cmd/logs.go
```diff
@@ -24,17 +22,17 @@ func newLogsCmd(config *config.Config) *LogsCmd {
 	logsCmd.Cmd = &cobra.Command{
 		Use:   "logs",
 		Args:  validators.NoArgs,
-		Short: "Top-level package for logs commands with Stripe.",
-		Long: fmt.Sprintf(`
-The logs package contains the sub-command 'tail', which allows you to tail your API request logs
-in real-time from Stripe.
+		Short: "Top-level command for interacting with API request logs from Stripe. Run `stripe logs --help` to learn more about different logs commands.",
```
tomer-stripe: This is too long, we should keep it to once sentence.

What about "Interact with Stripe request logs"?

aarongreen-stripe:
lgtm

aarongreen-stripe:
All filters should be included in the CLI now. Everything should be good for a final review. ptal @tomer-stripe  @ob-stripe

pkg/cmd/logs.go
```diff
@@ -0,0 +1,38 @@
...
+func newLogsCmd(config *config.Config) *LogsCmd {
+	logsCmd := &LogsCmd{
+		cfg: config,
+	}
+
+	logsCmd.Cmd = &cobra.Command{
+		Use:   "logs",
+		Args:  validators.NoArgs,
+		Short: "Interact with Stripe request logs",
+		Long: `
```
tomer-stripe: nit: first sentence should be on the same line otherwise it'll be a new line.

pkg/cmd/logs.go
```diff
@@ -0,0 +1,38 @@
...
+The logs command allows you to interact with your API request logs from Stripe. The first supported feature is log tailing, which allows you
+to view your API request logs in real-time.
+
+Invokable via:
```
tomer-stripe: Could you bold this?

pkg/cmd/logs/tail.go
```diff
@@ -0,0 +1,185 @@
...
+	tailCmd := &TailCmd{
+		cfg:        config,
+		LogFilters: &logTailing.LogFilters{},
+	}
+
+	tailCmd.Cmd = &cobra.Command{
+		Use:   "tail",
+		Args:  validators.NoArgs,
+		Short: "Listens for API request logs sent from Stripe to help test your integration.",
+		Long: fmt.Sprintf(`
```
tomer-stripe: nit: remove new line

pkg/cmd/logs/tail.go
```diff
@@ -0,0 +1,185 @@
...
+func NewTailCmd(config *config.Config) *TailCmd {
+	tailCmd := &TailCmd{
+		cfg:        config,
+		LogFilters: &logTailing.LogFilters{},
+	}
+
+	tailCmd.Cmd = &cobra.Command{
+		Use:   "tail",
+		Args:  validators.NoArgs,
+		Short: "Listens for API request logs sent from Stripe to help test your integration.",
```
tomer-stripe: What about, "Tails request logs created by making API requests to Stripe."

pkg/cmd/logs/tail.go
```diff
@@ -0,0 +1,185 @@
...
+	'DASHBOARD' - Requests that came through the Stripe Dashboard`,
+	)
+	tailCmd.Cmd.Flags().StringSliceVar(&tailCmd.LogFilters.FilterStatusCode, "filter-status-code", []string{}, "Filter request logs by status code")
+	tailCmd.Cmd.Flags().StringSliceVar(
+		&tailCmd.LogFilters.FilterStatusCodeType,
+		"filter-status-code-type",
+		[]string{},
+		`Filter request logs by status code type
+Acceptable values:
+	'200' - All 200 status codes
```
tomer-stripe: Should these be "All 2xx status codes", 4xx, 5xx or is it specifically 200, 400, and 500?

pkg/logtailing/tailer.go
```diff
@@ -0,0 +1,213 @@
...
+	signal.Notify(tailer.interruptCh, os.Interrupt, syscall.SIGTERM)
+
+	filters, err := jsonifyFilters(tailer.cfg.Filters)
+	if err != nil {
+		tailer.cfg.Log.Fatalf("Error while converting log filters to JSON encoding: %v", err)
+	}
+
+	session, err := tailer.stripeAuthClient.Authorize(tailer.cfg.DeviceName, tailer.cfg.WebSocketFeature, &filters)
+	if err != nil {
+		// TODO: better error handling / retries
```
tomer-stripe: Can't remember if I asked, but anything we should do before we merge? If not, could you create a ticket to track work here?

aarongreen-stripe: I don't think so. This was also commented in the corresponding location in `proxy.go` so I just copied it over here as well. I'm not 100% sure what it was commented for originally, perhaps just auth retrying, but we can remove it if desired since it doesn't really seem to be an action item right now.

pkg/cmd/logs/tail.go
aarongreen-stripe: 2XX, 4XX, 5XX. I'll update the comment to make it more clear. Should we also make it so the user specifies `2XX` instead of `200`?

pkg/logtailing/tailer.go
```diff
@@ -0,0 +1,212 @@
...
+
+	stripeAuthClient *stripeauth.Client
+	webSocketClient  *websocket.Client
+
+	interruptCh chan os.Signal
+}
+
+// EventPayload is the mapping for fields in event payloads from request log tailing
+type EventPayload struct {
+	CreatedAt string `json:"created_at"`
```
ob-stripe: Are we sending `created_at` as a human-readable string? I would have expected it to be a timestamp so we could do the conversion client-side, using the local timezone.

aarongreen-stripe: We do the conversion on the backend so it's sent as human-readable but in UTC, and it's noted as UTC in the output to the user.

I did this originally since I figured it's easier to just have everything that arrives on the client be in the proper form of output, so it's just `receive -> display`, as opposed to `receive -> any conversions -> display`, since doing the conversion would assumably be easier in Ruby than in Go.

If it's a strong preference to do the conversion on the client-side for the timezone to be in local time, I can make that change, but since it's also a backend change, it may delay the release of log tailing by a day.

Thoughts? @ob-stripe

ob-stripe: If it's not too much toil I would prefer we abide by the existing API conventions and send all times as Unix timestamps, and do any conversion client-side.

It's fairly easy to do the conversion in Go: https://play.golang.org/p/W18lIi9wDvH

aarongreen-stripe:
Feedback should be addressed. ptal @ob-stripe

PR #58, opened 2019-07-27
Add tools package

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Adds a `tools` package used to import vfsgen, as advised in https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md.

Also removes the `vfsgen.go` script and replaces it with a direct invocation of `vfsgendev`: https://github.com/shurcooL/vfsgen#vfsgendev-usage.

Replaces #56.


ob-stripe:
Opportunistically updated all dependencies with `go get -u ./... && go mod tidy`.

tomer-stripe:
lgtm!

I'll close out the other pr.

brandur-stripe:
Nice OB! It might be a good idea to do a bit of a writeup in the `tools` package or somewhere that describes why it exists and how all of this works — just to remind our future selves when we need to debug something.

PR #60, opened 2019-07-29
Set GOBIN and PATH env vars

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Set `GOBIN` and `PATH` environment variables so that `go generate ./...` will work even if vfsgendev is not installed globally.

This is listed in https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md, but I missed it because `vfsgen` was installed globally in my environment so the command worked.

This is kind of difficult to test on Travis, but I've verified that it works locally:

```sh
$ which vfsgendev
vfsgendev not found
$ make build
go mod download
go generate ./...
writing fs_vfsdata.go
go build -o stripe -ldflags "-s -w" cmd/stripe/main.go
```


ob-stripe:
Going to self-approve and try a release 🤞

ob-stripe:
Didn't work, but fairly sure that's because the env vars need to be set in `.goreleaser.yml` as well 🤦‍♂

PR #66, opened 2019-07-31
Update installation instructions

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Add installation instructions for using the new apt/yum repositories.

I think the README is already getting a bit too big and we should consider splitting it into wiki pages, but this is probably good enough for now.


tomer-stripe:
I'm going to merge this in so we have the better instructions for people :D

PR #67, opened 2019-07-31
Tweak OpenAPI spec embedding (again)

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Sorry for all the back and forth... Hopefully this will work.

This is basically a revert of #58, except that:
- the build tag for the Go generation script is now `vfsgen` instead of `ignore`. `ignore` is a special value which causes the package to be completely ignored and thus the dependencies were omitted from `go.mod`. By setting it to a different value (`vfsgen`, to be consistent with the file name, but it could really be anything), the dependencies are properly included.
- we override the modification times to be the Unix epoch (using sample code provided by vfsgen's author [here](https://github.com/shurcooL/vfsgen/issues/66#issuecomment-463103788)). git does not manage mod times, so the mod times can change between different environments, which results in `go generate ./...` producing different results. This ensures that the generated `fs_vfsdata.go` file will only change when the file contents are actually updated.


brandur-stripe:
> we override the modification times to be the Unix epoch (using sample code provided by vfsgen's author here). git does not manage mod times, so the mod times can change between different environments, which results in go generate ./... producing different results. This ensures that the generated fs_vfsdata.go file will only change when the file contents are actually updated.

Would you mind putting this explanation in the file somewhere? Just reading through that code, it wasn't obvious to me at all why we'd bother with all this mod time stuff, and I think it's going to be easy to forget why this is done a few months from now.

Nice fix though! LGTM otherwise.

ob-stripe:
Added a longer comment: https://github.com/stripe/stripe-cli/pull/67/files#diff-ab73c2df8ed2edd405db26c9a8083c5dR18-R22

PR #72, opened 2019-08-02
[wip] Resource commands

ob-stripe:
### Reviewers
r? @
cc @stripe/dev-platform

 ### Summary
This is very much a WIP, but it's (barely) functional, so opening a PR for early feedback.

This implements three new categories of commands:
- namespace commands: top-level, hold resource commands
- resource commands: either top-level (for core resources like charges, customers, etc.) or nested under a namespace command, hold operation commands
- operation commands: nested under a resource command, actually execute an API request

Operation commands are simply wrappers over the existing `post`/`get`/`delete` commands right now, so parameters are provided using the `-d` flag.

API parameters that require IDs (retrieve, update, custom operations like charge capture...) expect the ID as an argument. Nested resources need two IDs, the parent resource's and the nested resource's.

A few examples:
```
$ stripe charges create -d amount=123 -d currency=usd -d source=tok_visa
$ stripe charges retrieve ch_123
$ stripe issuing cards update ic_123 -d "metadata[key]=value"
$ stripe subscriptions cancel sub_123
```

TODO:
- [ ] Write tests
- [x] Use pluralized resource names (`charges` instead of `charge`).
  ~This is actually difficult because the spec doesn't have a field for that and it's not as simple as adding an "s" to the resource name (some resources like `balance` need to stay singular, other use words with non-trivial pluralization like `summary`). I'll need to extract the pluralized name from the URL, but that is itself not trivial because there's a not a single format that we can apply to all URLs (e.g. `/v1/invoices/upcoming`: is `upcoming` a method on the `invoices` resource, or is `upcoming` a resource in the `invoices` namespace?).~
  There really is no good way to get the pluralized name from the spec, so I just went with a simple whitelist.
- [ ] Strip out HTML from descriptions
- [ ] Figure out short descriptions for methods and parameters
  The descriptions in the spec are verbose (they're the same as the API reference). Not sure there's a good solution here unfortunately. Maybe just include links to the API reference?
- [ ] Figure out how to deal with request parameters: flags vs. args? Nested parameters? Array parameters?
- [ ] Probably many other things

### DIFF
--- a/pkg/cmd/resource/namespace.go
+++ b/pkg/cmd/resource/namespace.go
@@ -1,0 +1,86 @@
+package resource
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/ansi"
+)
+
+//
+// Public types
+//
+
+// NamespaceCmd represents namespace commands. Namespace commands are top-level
+// commands that are simply containers for resource commands.
+type NamespaceCmd struct {
+	Cmd *cobra.Command
+
+	Name string
+
+	ResourceCmds map[string]*ResourceCmd
+}
+
+//
+// Public functions
+//
+
+// NewNamespaceCmd returns a new NamespaceCmd.
+func NewNamespaceCmd(rootCmd *cobra.Command, namespaceName string) *NamespaceCmd {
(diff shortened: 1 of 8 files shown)


ob-stripe:
(Closing in favor of #87.)

PR #73, opened 2019-08-02
Add -v flag for version

### LINKED ISSUE #37: stripe -v Support
I think it'd be awesome to see a shortcut to run `stripe -v` and have the same behavior as `stripe --version`.

tomer-stripe:
### Reviewers
cc @stripe/dev-platform
cc @fishcharlie

 ### Summary
Fixes #37, adds `-v` flag for version and cleans up version printing.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -6,7 +6,10 @@ import (
 	"github.com/stripe/stripe-cli/pkg/ansi"
 )
 
-// Version number of the CLI
+// Version of the CLI -- currently in beta
+// This is set to the actual version by GoReleaser, identify by the
+// git tag assigned to the release Versions built from source will
+// always show master.
 var Version = "master"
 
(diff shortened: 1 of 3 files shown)


pkg/version/version.go
```diff
@@ -1,4 +1,12 @@
 package version
 
+import (
+	"fmt"
+
+	"github.com/stripe/stripe-cli/pkg/ansi"
+)
+
 // Version number of the CLI
 var Version = "master"
```
brandur-stripe: It might be worthwhile to note here that this value may be overwritten by Goreleaser, and that's how the final version value is generated (IMO, the fact that Go's compiler let's you do this is not obvious at all). Here's the comment I had in stripe-mock for inspiration:

```
// This is set to the actual version by GoReleaser (using `-ldflags "-X ..."`)
// as it's run. Versions built from source will always show master.
```

tomer-stripe: Good call, that could definitely be confusing otherwise. Updating!

PR #74, opened 2019-08-02
Make the status icons more suble

tomer-stripe:
### Reviewers
r? @romain-stripe @aarongreen-stripe @brandur-stripe 
cc @stripe/dev-platform


brandur-stripe:
(You may have noticed already, but it looks like you'll have to update the test suite too.)

tomer-stripe:
Yep, caught that 😅

PR #76, opened 2019-08-06
--help formatting

aarongreen-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
With the present organization, `stripe --help` returns information about the `GET`, `POST`, and `DELETE` commands in a section called `API commands`, and `logs` is in its own section and feels empty all alone.

This proposes a new format for the help section, where the existing `API commands` is renamed to `HTTP commands` and there is a new `Stripe commands` section which contains the `logs` command and the `status` command.

Here's what the new help section would look like.
Screen Shot 2019-08-06 at 3 35 36 PM

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -21,12 +21,13 @@ var rootCmd = &cobra.Command{
 	SilenceUsage:  true,
 	SilenceErrors: true,
 	Annotations: map[string]string{
-		"get":     "api",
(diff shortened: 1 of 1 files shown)


tomer-stripe:
Could you post the output of `--help` in the description? I'd be curious to see what other people think.

aarongreen-stripe:
lgtm

PR #79, opened 2019-08-07
Replace endpoints map with list of endpoint routes

ob-stripe:
### Reviewers
r? @aarongreen-stripe @brandur-stripe @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
More refactoring. This replaces `EndpointsMap` (a map of URLs to event types) with `EndpointRoutes` (a list of `EndpointRoute` which is a struct containing a URL and a list of event types).

This allows for having more than one endpoint with the same URL, which will be needed for Connect support.

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -21,7 +21,16 @@ import (
 // Public types
 //
 
-// Config provides the cfguration of a Proxy
+// EndpointRoute describes a local endpoint's routing configuration.
+type EndpointRoute struct {
+	// URL is the endpoint's URL.
+	URL string
+
+	// EventTypes is the list of event types that should be sent to the endpoint.
+	EventTypes []string
+}
+
+// Config provides the configuration of a Proxy
 type Config struct {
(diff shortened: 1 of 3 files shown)


pkg/proxy/proxy.go
```diff
@@ -21,7 +21,14 @@ import (
 // Public types
 //
 
-// Config provides the cfguration of a Proxy
+// EndpointRoute
```
brandur-stripe: Mind throwing a fleshed out comment on here?

pkg/proxy/proxy.go
```diff
@@ -21,7 +21,14 @@ import (
...
+type EndpointRoute struct {
+	URL string
+
+	EventTypes []string
```
brandur-stripe: Nit, but consider either (1) commenting these, or (2) tightening up the struct by killing newlines.

brandur-stripe:
Left a couple minor comments, but LGTM.

ob-stripe:
Added comments as requested, self-approving.

Thanks Brandur!

PR #81, opened 2019-08-07
Create config command that lets users modify their configs

tomer-stripe:
### Reviewers
r? @aarongreen-stripe @brandur-stripe @ob-stripe 
cc @stripe/dev-platform

 ### Summary
This creates a config command that:
1. let's users define new config options: `stripe config color off`
2. let's users list all currently set configs: `stripe config --list`
3. let's users open an editor to modify the config file: `stripe config --edit`
4. let's users remove pre-defined config fields: `stripe config --unset color`

### DIFF
--- a/pkg/cmd/config.go
+++ b/pkg/cmd/config.go
@@ -1,0 +1,45 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/stripe/stripe-cli/pkg/config"
+)
+
+type configCmd struct {
+	cmd    *cobra.Command
+	config *config.Config
+
+	list  bool
+	edit  bool
+	unset bool
+}
+
+func newConfigCmd() *configCmd {
+	cc := &configCmd{
+		config: &Config,
+	}
+	cc.cmd = &cobra.Command{
+		Use:   "config",
+		Short: "Manually change the config values for the CLI",
+		RunE:  cc.runConfigCmd,
+	}
+
+	cc.cmd.Flags().BoolVar(&cc.list, "list", false, "list configs")
+	cc.cmd.Flags().BoolVarP(&cc.edit, "edit", "e", false, "open editor to the config file")
+	cc.cmd.Flags().BoolVar(&cc.unset, "unset", false, "unset a specific config field")
+
+	return cc
+}
+
+func (cc *configCmd) runConfigCmd(cmd *cobra.Command, args []string) error {
+	if len(args) == 2 {
+		return cc.config.Profile.WriteConfigField(args[0], args[1])
+	} else if len(args) == 1 && cc.unset {
+		return cc.config.Profile.DeleteConfigField(args[0])
+	} else if cc.list {
+		cc.config.Profile.PrintConfig()
+	} else if cc.edit {
+		return cc.config.EditConfig()
+	}
+	return nil
+}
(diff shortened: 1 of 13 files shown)


pkg/cmd/config.go
```diff
@@ -0,0 +1,45 @@
...
+
+	cc.cmd.Flags().BoolVar(&cc.list, "list", false, "list configs")
+	cc.cmd.Flags().BoolVarP(&cc.edit, "edit", "e", false, "open editor to the config file")
+	cc.cmd.Flags().BoolVar(&cc.unset, "unset", false, "unset a specific config field")
+
+	return cc
+}
+
+func (cc *configCmd) runConfigCmd(cmd *cobra.Command, args []string) error {
+	if len(args) == 2 {
```
aarongreen-stripe: Does this work even if the user specifies `--project-name` in the command?

pkg/cmd/config.go
```diff
@@ -0,0 +1,45 @@
...
+		return cc.config.Profile.WriteConfigField(args[0], args[1])
+	} else if len(args) == 1 && cc.unset {
```
aarongreen-stripe: Same as above

pkg/config/config.go
```diff
@@ -22,6 +25,8 @@ type Config struct {
 	LogLevel     string
 	Profile      Profile
 	ProfilesFile string
+
+	location string
```
aarongreen-stripe: How is `location` different from `ProfilesFile`? It appears that they store the same thing, can we just re-use one of them for all purposes?

aarongreen-stripe:
Just a couple small Q's, otherwise lgtm

pkg/cmd/config.go
tomer-stripe: yep! that's handled at the profile level

pkg/config/config.go
tomer-stripe: `ProfilesFile` currently only stores the profile location when it's user supplied, not the implicit location used by the CLI. I'd be worried about overriding that behavior.

pkg/config/profile.go
```diff
@@ -45,10 +40,15 @@ func (p *Profile) GetDeviceName() (string, error) {
...
-	key := viper.GetString("secret_key")
+// GetAPIKey will return the existing key for the given profile
+func (p *Profile) GetAPIKey() (string, error) {
+	// If the user doesn't have an api_key field set, they might be using an
+	// old configuration so try to read from secret_key
+	if !viper.IsSet(p.GetConfigField("api_key")) {
+		p.RegisterAlias("api_key", "secret_key")
+	}
+
+	key := viper.GetString(p.GetConfigField("api_key"))
```
ob-stripe: I think this is incorrect, it should be:
```
key := viper.GetString("api_key")
```

Otherwise you're only reading using the profile prefix, which we're already doing below (L62).

pkg/config/config.go
```diff
@@ -110,11 +115,58 @@ func (c *Config) InitConfig() {
...
+		cmd := exec.Command(os.Getenv("EDITOR"), c.ProfilesFile)
+		// Some editors detect whether they have control of stdin/out and will
+		// fail if they do not.
+		cmd.Stdin = os.Stdin
+		cmd.Stdout = os.Stdout
+		return cmd.Run()
+	case "windows":
+		// As far as I can tell, Windows doesn't have an easily accesible or
+		// comparable option to $EDITOR, so default to notepad for now
+		err = exec.Command("notepad", c.ProfilesFile).Run()
```
ob-stripe: This might be fixed in Windows 10, but in the past notepad did not handle LF line endings (only CR+LF Windows-style line endings). We should test to check if it displays the config file correctly.

PR #87, opened 2019-08-09
Add namespaces, resources and operations subcommands

ob-stripe:
### Reviewers
r? @aarongreen-stripe @brandur-stripe @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
This PR adds a new `resource` package (nested under the `cmd` package) and three new structs:
- `NamespaceCmd` represents namespace commands. Namespace commands are top-level commands and serve as holders for resource commands.
- `ResourceCmd` represents resource commands. Resource commands are nested under namespace commands and serve as holders for operation commands.
- `OperationCmd` represents operation commands. Operation commands execute API requests.

~Currently, `OperationCmd` simply wraps over `requests.Base` and uses all the same flags (e.g. request parameters are passed with the `-d` flag). The only difference is that users don't need to pass the path as it's a property of the operation command.~

~E.g. if `update` is an operation command under the `customers` resource command:~

```
$ stripe customers update cus_123 -d email=foo@example.com -d "metadata[key]=value"
```

~would send a `POST` request to `/v1/customers/cus_123` with `email=foo@example.com&metadata[key]=value` in the request's body.~

As of #98, request parameters are passed as arguments, not flags:

```
$ stripe customers update cus_123 email=foo@example.com "metadata[key]=value"
```

~Note that at the moment, no namespace/resource/operation commands are created at all. I plan on creating the commands using the OpenAPI spec, either dynamically or by generating Go code from the spec (probably the latter, because parsing the JSON spec is actually quite extensive due to the spec's large size).~

As of #90, commands are generated.


pkg/cmd/resource/resource.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+//
+// Public types
+//
+
+// ResourceCmd represents resource commands. Resource commands can be either
+// top-level commands or nested under namespace commands. Resource commands
+// are containers for operation commands.
+type ResourceCmd struct { //nolint:golint
```
brandur-stripe: OOC, why did this have to be nolinted?

pkg/cmd/resource/namespace.go
```diff
@@ -0,0 +1,88 @@
+package resource
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/ansi"
```
brandur-stripe: Sorry to bring up these minor things all the time, but unless you have a more cohesive convention for organizing these imports in mind, you might want to just stick to two groups: (1) built-ins (like `fmt`), and (2) everything else (like `github.com/...`).

(And there's a few other files below that could use the same treatment.)

pkg/cmd/resource/namespace.go
```diff
@@ -0,0 +1,88 @@
...
+		ResourceCmds: make(map[string]*ResourceCmd),
+	}
+}
+
+//
+// Private functions
+//
+
+func namespaceUsageTemplate() string {
+	return fmt.Sprintf(`%s{{if .Runnable}}
```
brandur-stripe: Lots of stuff in here. I assume that the new template is modeled after some default that you found inside Cobra?

pkg/cmd/resource/namespace_test.go
```diff
@@ -0,0 +1,29 @@
...
+func TestNewNamespaceCmd_EmptyName(t *testing.T) {
+	rootCmd := &cobra.Command{Annotations: make(map[string]string)}
+
+	nsc := NewNamespaceCmd(rootCmd, "")
+
+	assert.Equal(t, "", nsc.Name)
+	assert.False(t, rootCmd.HasSubCommands())
+	_, ok := rootCmd.Annotations[""]
+	assert.False(t, ok)
+}
```
brandur-stripe: All the template stuff is evaluated at runtime only, so you may want to add a test here that exercises that in at least some basic way to make sure that it's right (and same goes for resource and operation commands).

pkg/cmd/resource/namespace.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/stripe/stripe-cli/pkg/ansi"
+)
+
+//
+// Public types
+//
+
+// NamespaceCmd represents namespace commands. Namespace commands are top-level
+// commands that are simply containers for resource commands.
```
brandur-stripe: Would it be possible to specify an example of what a namespace command might be here? I assume it's something like "issuing" or "terminal"?

pkg/cmd/resource/resource.go
ob-stripe: Because the package's name is `resource`, golint thinks `resource.ResourceCmd` is a redundant name and the struct should be named `resource.Cmd`.

pkg/cmd/resource/namespace.go
ob-stripe: goimports is configured to use three groups:
1. built-ins
2. third-party packages
3. internal packages

i.e. all packages that start with `github.com/stripe/stripe-cli` are grouped separately at the end.

https://github.com/stripe/stripe-cli/blob/b489043a11b42ffae12a3a98dd01e122d89ce393/.golangci.yml#L6

pkg/cmd/resource/namespace.go
ob-stripe: Yes. We're already overriding the default template at the root level here: https://github.com/stripe/stripe-cli/blob/b489043a11b42ffae12a3a98dd01e122d89ce393/pkg/cmd/root.go#L53-L96

in order to use ANSI bold sequences for the section names. Here I'm reusing the same pattern and also changing some section names ("commands" to "resources").

There might be a better, less repetitive way to do this though.

pkg/cmd/resource/namespace_test.go
ob-stripe: 👍

pkg/cmd/resource/namespace.go
ob-stripe: Yes, exactly. Will update the comment 👍

pkg/cmd/resource/resource.go
brandur-stripe: Ah, I see.

pkg/cmd/resource/namespace.go
brandur-stripe: Ah, I see. Okay, just as long as we have a system.

ob-stripe:
I ended up using this PR as the integration branch for the entire feature. Individual commits have been reviewed in separate PRs, I've just squashed everything in a single commit and am going to merge into master.

PR #90, opened 2019-08-12
Generate resource commands

ob-stripe:
### Reviewers
r? @brandur-stripe @tomer-stripe @aarongreen-stripe 
cc @stripe/dev-platform

 ### Summary
This PR:
- removes the vfsgen-based spec embedding
- adds a custom generator script (`gen_resources.go`) that generates a Go source file (`resources.go`) for all the namespace/resource/operation commands

The reason I'm doing it this way instead of dynamically creating the commands from the embedded spec is that parsing the JSON spec is an expensive operation (~130 milliseconds on my laptop) due to the spec being so large, and we don't want to slow down the CLI too much.


pkg/cmd/gen_resources.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type OperationData struct {
+	Path     string
+	HTTPVerb string
+}
+
+func main() {
+	funcMap := template.FuncMap{
+		"ToCamel": strcase.ToCamel,
```
tomer-stripe: 🐪:D

pkg/cmd/gen_resources.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	tmpl := template.Must(template.New("resources.go.tpl").Funcs(funcMap).ParseFiles("resources.go.tpl"))
```
tomer-stripe: Could you explain a bit more (and maybe add a comment) about why we need this?

pkg/cmd/gen_resources.go
```diff
@@ -0,0 +1,128 @@
...
+type ResourceData struct {
+	Operations map[string]*OperationData
+}
+
+type OperationData struct {
+	Path     string
+	HTTPVerb string
+}
+
+func main() {
```
tomer-stripe: Generally, could you comment the steps here to explain what's happening?

ob-stripe:
I've added some comments to the `gen_resources.go` script, ptal @tomer-stripe

tomer-stripe:
lgtm, thanks for adding those clarification comments!

ob-stripe:
Thanks!

PR #98, opened 2019-08-13
Add support for passing request parameters as arguments

ob-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
Add support for passing request parameters as arguments instead of having to use the `-d` flag in resource commands.

Unfortunately the current state of the code makes it kind of difficult to write automated tests for this, I think we'd need to do some refactoring of the UI layer as well as the `requests` package (to inject the API key).

I did do some manual testing:
```sh
$ ./stripe charges update --help
Usage:
  stripe charges update <charge> [param1=value1] [param2=value2] ...

Flags:
  -v, --api-version string      Set the Stripe API version to use for your request
  -c, --confirm                 Automatically confirm the command being entered. WARNING: This will result in NOT being prompted for confirmation for certain commands
  -e, --expand stringArray      Response attributes to expand inline. Available on all API requests, see the documentation for specific objects that support expansion
  -h, --help                    help for update
  -i, --idempotency string      Sets the idempotency key for your request, preventing replaying the same requests within a 24 hour period
  -s, --show-headers            Show headers on responses to GET, POST, and DELETE requests
      --stripe-account string   Set a header identifying the connected account for which the request is being made

$./stripe charges update ch_123 "description=Foo bar baz" "metadata[key]=value"
{
  "id": "ch_123",
  "object": "charge",
  ...
  "description": "Foo bar baz",
  ...
  "metadata": {
    "key": "value"
  },
  ...
}
```

### DIFF
--- a/pkg/cmd/resource/resource.go
+++ b/pkg/cmd/resource/resource.go
@@ -36,6 +36,8 @@ func GetResourceCmdName(name string) string {
 	case "balance":
 		// `balance` is a singleton resource and is not pluralized
 		return "balance"
+	case "capability":
+		return "capabilities"
 	case "three_d_secure":
 		return "3d_secure"
 	default:
--- a/pkg/cmd/delete.go
+++ b/pkg/cmd/delete.go
@@ -42,7 +42,7 @@ To delete a charge:
 		RunE: gc.reqs.RunRequestsCmd,
 	}
 
-	gc.reqs.InitFlags()
+	gc.reqs.InitFlags(true)
 
 	return gc
 }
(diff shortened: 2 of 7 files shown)


pkg/cmd/resource/resource.go
```diff
@@ -36,6 +36,8 @@ func GetResourceCmdName(name string) string {
 	case "balance":
 		// `balance` is a singleton resource and is not pluralized
 		return "balance"
+	case "capability":
+		return "capabilities"
```
ob-stripe: Opportunistic fix to avoid creating a `capabilitys` command :)

tomer-stripe:
lgtm!

PR #99, opened 2019-08-14
Read from persisted color settings

tomer-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
This lets the CLI read persisted color settings configured by users.

* I set the default color to `""` so that we know if someone is using the command line flag we should override the pre-defined configs
* I moved reading the profile config in to be a bit higher so that we can set the log colors sooner
* `GetColor` reads off the profile (like most other settings) and checks if color is provided first at a top level, then if it's set in the config file. If it's unset in the config file, it returns `""` and goes with auto.

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -35,12 +35,12 @@ func (p *Profile) GetColor() (string, error) {
 
 	color = viper.GetString(p.GetConfigField("color"))
 	switch color {
-	case "", "auto":
-		return "auto", nil
-	case "on":
-		return "on", nil
-	case "off":
-		return "off", nil
+	case "", ColorAuto:
+		return ColorAuto, nil
+	case ColorOn:
+		return ColorOn, nil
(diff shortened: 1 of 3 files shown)


aarongreen-stripe:
Does `viper.GetString(...)` pull the command-line arg? Maybe I'm missing something obvious, but I don't see where we override the config file with the command-line argument.

pkg/config/profile.go
```diff
@@ -26,6 +27,25 @@ func (p *Profile) CreateProfile() error {
 	return nil
 }
 
+func (p *Profile) GetColor() (string, error) {
+	color := viper.GetString("color")
```
aarongreen-stripe: Just making sure, this reads properly from the command-line argument?

tomer-stripe: Yep! This correctly grabs the command-line argument and overrides the stored config.

PR #107, opened 2019-08-14
Fix color support for Windows 10

### LINKED ISSUE #93: Windows command line showing invalid characters
### Issue
_Describe what happened and what you were trying to do_

I ran the stripe executable from the command line and while the connections seem to establish correctly, invalid characters are output to the console window which is both "ugly", but also tricky (i.e. to copy the webhook signing secret).

image

### Expected Behavior
_Tell us what you expected to happen_

I expect the intention is to include these values in quotes, or to mark them up, however apparently the control characters aren't being recognized.

### Steps to reproduce
_What are the steps we can take to reproduce this and verify it's fixed?_

Windows 10 Home
Version	10.0.18362 Build 18362

I followed the very basic steps to download and execute the stripe.exe - nothing fancy. You can see the entire output of the commands I've run in the screenshot above.

Command line codepage is 437 with font Consolas if it matters (grabbed from command properties).

### Environment
_Select one of: macOS, Linux, Windows, Other_

Windows - see above for more details.

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
The Windows console does not support ANSI color codes, resulting in strange characters in the output (cf. #93). With Windows 10, the console _can_ support ANSI colors, but support must be explicitly enabled.

This PR does that by using some code adapted from this comment: https://github.com/sirupsen/logrus/issues/496#issue-210627761.

It also changes the [spinner charset](https://github.com/briandowns/spinner#available-character-sets) to use standard ASCII characters on Windows, since the default console font does not support the fancy characters.

Fixes #93 (to an extent).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,6 @@ require (
 	github.com/x-cray/logrus-prefixed-formatter v0.5.2
 	golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4
 	golang.org/x/net v0.0.0-20190724013045-ca1201d0de80 // indirect
-	golang.org/x/sys v0.0.0-20190726091711-fc99dfbffb4e // indirect
+	golang.org/x/sys v0.0.0-20190726091711-fc99dfbffb4e
 	golang.org/x/text v0.3.2 // indirect
 )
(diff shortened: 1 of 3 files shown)


ob-stripe:
Released as v0.4.4.

PR #100, opened 2019-08-14
Help improvements

ob-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
A bunch of improvements to help messages:
- wrap flag descriptions according to terminal width \o/
- make short messages shorter to ensure that command descriptions fit under 80 characters
- make short messages more consistent (don't use third person, don't add a trailing `.`)
- add required arguments to `delete`/`get`/`post` and `trigger`
- use the dedicated `Example` arg where applicable

PR #110, opened 2019-08-15
Add resources help topic

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
This PR adds a `resources` "help topic", i.e. a subcommand whose only purpose is to display a help message.

Output looks like this:

```
$ ./stripe
⚠️  The Stripe CLI is in beta! Share your feedback with `stripe feedback` ⚠️

[snip]

Additional help topics:
stripe resources                       list namespace and resource subcommands

Use "stripe [command] --help" for more information about a command.
```

```
$ ./stripe resources
⚠️  The Stripe CLI is in beta! Share your feedback with `stripe feedback` ⚠️

Available Namespaces:
  checkout
  issuing
  radar
  reporting
  terminal

Available Resources:
  3d_secure
  account_links
[snip]
  usage_records
  webhook_endpoints

Use "stripe [command] --help" for more information about a command.
```

Once I can get short descriptions for each resource in the spec, we'll display them there too.

This PR also uses the new wrapping functions from #100 in the help templates for the namespace, resource and operation subcommands.


ob-stripe:
I added a couple of tests to check that `stripe --help` and `stripe resources` render without errors and that the output contains substrings we expect to see from the templates. It's a bit janky but it'll do for now.

> I wonder if we should be wrapping to a width of `max(80, terminal_width)`?

If we do that I think we might as well just wrap to 80 characters, as it's fairly unlikely that the terminal width would be less than 80 characters, right?

ob-stripe:
(Going to merge this as-is for now because I want to get resource commands out the door, but let's talk more about optimal line length!)

brandur-stripe:
> If we do that I think we might as well just wrap to 80 characters, as it's fairly unlikely that the terminal width would be less than 80 characters, right?

Probably not that common I guess, but I've occasionally done it in the past :) If the code's already written to allow for maximum possible graceful degradation, I'd say: why not.

PR #115, opened 2019-08-16
Various lint fixes

ob-stripe:
### Reviewers
r? @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
Various lint fixes. Nothing terribly out of the ordinary, but let me know if you have any questions.

### DIFF
--- a/pkg/cmd/listen_test.go
+++ b/pkg/cmd/listen_test.go
@@ -3,20 +3,20 @@ package cmd
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/stripe/stripe-cli/pkg/requests"
 )
 
 func TestParseUrl(t *testing.T) {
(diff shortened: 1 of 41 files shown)


pkg/cmd/listen_test.go
```diff
@@ -3,7 +3,7 @@ package cmd
 import (
 	"testing"
 
-	assert "github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
```
brandur: Not sure what the general project convention is, but FYI in case you didn't know: there's a slight difference between assert and require in that the require variant will fail tests on the first failure. Generally I've found this to be a sane default personally, but admittedly I probably deviate from the Go community in that opinion.

ob-stripe: Oh, I was mostly fixing the inconsistency because some tests used `assert` and others aliased `assert` to `require`.

I've updated everything to use `require` directly.

ob-stripe:
Thanks!

PR #117, opened 2019-08-19
Configure .env for samples

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform @adreyfus-stripe @ctrudeau-stripe @auchenberg-stripe 

 ### Summary
This fixes a bug with .env and configures the files.

The way I'm doing this is:
* open the .env.example from the new direction
* parse the .env.example file
* make a request to authorize the CLI for webhooks to get the webhook signing secret
* override the api key and webhook signing secret
* write to a new .env file in the new directory

I haven't added logic yet to set the publishable key. I'm checking to see if we can return the publishable key through the API which I'd prefer doing for our users.

Also I didn't add tests since the dotenv lib doesn't support passing in a filesystem. I was thinking about opening a PR to handle that correctly but I don't want to block this PR.


pkg/samples/samples.go
```diff
@@ -233,6 +235,47 @@ func (s *Samples) Copy(target string) error {
 	return nil
 }
 
+func (s *Samples) ConfigureDotEnv(sampleLocation string) error {
```
adreyfus-stripe: We hardcode some of the .env paths in the servers so I think we'd have a similar problem to the STATIC_DIR path problem when users pick a single language. I'm updating the servers to hardcode the paths in a ENV_PATH variable, so if you look for that and remove one of the "../" that should probably work.

PR #118, opened 2019-08-19
Setup samples commands

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform @adreyfus-stripe @ctrudeau-stripe @auchenberg-stripe 

 ### Summary
This moves all the sample's code to to `stripe samples create ...` and `stripe samples list`.

I created a new branch because rebasing went poorly on the other branch.


pkg/cmd/samples.go
```diff
@@ -1,88 +1,27 @@
...
+			Use:   "samples",
+			Short: `Sample integrations built by Stripe`,
+			Long:  ``,
 		},
 	}
+
+	samplesCmd.Cmd.AddCommand(samples.NewCreateCmd(&Config))
+	samplesCmd.Cmd.AddCommand(samples.NewListCmd())
+
+	return samplesCmd.Cmd
```
ob-stripe: Any reason you're deviating from the usual pattern of returning the `whateverCmd` struct and then calling `rootCmd.AddCommand(whateverCmd.cmd)`?

pkg/cmd/samples/create.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// Perform the copy of the sample given the selected options
+	// from the selections above
+	err = sample.Copy(targetPath)
+	if err != nil {
+		return err
+	}
+
+	// TODO: setup .env
```
adreyfus-stripe: just flagging that another TODO is to setup STATIC_DIR path if they've specified a language

adreyfus-stripe:
yay! ty

pkg/cmd/samples.go
brandur-stripe: Noticed this as well! It looks a little deviant in the list in `root.go` above.

brandur-stripe:
Still looks good! What was the status of where we're going with testing again?

pkg/cmd/samples.go
tomer-stripe: I changed this because we don't actually use the returned cmd struct anywhere so it seemed unnecessary to export it, but at that point the linter is unhappy because we're returning an un-exported type. I'll change this back for now.

pkg/cmd/samples/create.go
tomer-stripe: Thanks :D! I'll take care of that in the other PR where I focus more on the .env

tomer-stripe:
@ob-stripe @brandur-stripe for testing, I wasn't entirely sure what to test here given we haven't setup the stdout capture yet and that seemed it'd be the most useful here. A lot of the other changes are more structural. The `runCreateCmd` has its components unit-tested in other PRs but testing the entire thing is tricky since this specifically has the local file system and the native git interfaces.

tomer-stripe:
@ob-stripe @brandur-stripe I changed back to the existing pattern! I'm going to merge this for now and follow up with testing later since there isn't much I can't do here for that.

PR #125, opened 2019-08-21
[wip] Setup fixtures command

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 

 ### Summary
_marked as WIP because I'm not super happy with the code atm, wanted to get initial feedback_

Still to do:
- [x] more test cases
- [ ] url variable parsing
- [ ] setup .env
- [ ] add metadata to flag when a fixture was created: `{"_created_by_fixture": "<DATE>"}`

This implements a relatively simple `fixtures` command that takes a json file of a specific structure and makes requests to the API. The JSON files looks like:
```
[
  {
    "name": "cust_bender",
    "http": {
      "path": "/v1/customers",
      "method": "post"
    },
    "data": {
      "email": "bender@planex.com",
      "address": {
        "city": "New New York"
      }
    }
  },
  {
    "name": "ch_bender",
    "http": {
      "path": "/v1/charges",
      "method": "post"
    },
    "data": {
      "customer": "#$cust_bender:id",
      "source": "tok_visa",
      "amount": "100",
      "currency": "usd"
    }
  }
]
```

Where:
* `name` specifies the name of the request, used to store the responses for future reference
* `http` defines the path and http method for the request
* `data` is all of the params for the request

For making the requests, I'm using the pre-exiting requests package we built.

The way this works is by recursively parsing the data object and building the appropriate params (as well as correctly nesting fields for things like `address[city]`). This uses reflect to determine if the data is a string, int, map, or array. While the data is being parsed, anytime a string value comes up I try to parse it for a query, which is indicated by a string started with `#$` (I tried to find something that was unlikely to trigger false positives, `$` seemed risky).

The `#$` queries are structured as: `#$<NAME>:<JSON PATH>`, where the json path is defined from the `gojsonq` library: https://github.com/thedevsaddam/gojsonq/wiki/Queries#findpath

I have some light unit testing as a sanity check and also verified manually that this works:
Screen Shot 2019-08-21 at 4 29 06 AM
Screen Shot 2019-08-21 at 4 28 50 AM
Screen Shot 2019-08-21 at 4 28 35 AM

Any feedback on how this works, how I could clean it up/simplify it? The recursive code is a bit gnarly and I think I don't need the top-most case of `parseInterface`.


brandur-stripe:
@tomer-stripe Would you mind putting together a/a couple sample command line invocation(s) for what the feature would look like in use? I don't think I'm 100% successfully wrapping my head around what I'd use it for just based on the description.

tomer-stripe:
@brandur-stripe sure! The command itself it's pretty "small". If you have a json file called `test_data.json` (that contains the json from the description) you'd run it with `stripe samples fixtures test_data.json`. That command parses the json file and then makes all the corresponding API requests in the JSON file; the idea being that we can setup some fixture data for users to use our samples with some data that's easy to populate.

The json file itself is an ordered list of API requests for the CLI to run on behalf of the users, with some extra logic so that we can have requests use data from previous requests to make associations.

PR #126, opened 2019-08-21
Add command to quickly open stripe pages

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform @auchenberg-stripe @mg-stripe 

 ### Summary
Add a command to quickly open stripe pages.

I hard-coded this for the time being because I the dashboard does not consistently name paths, which would've made the more dynamic solution have a few tricky edge-cases. I think the long-term best solution is to probably generate some json file from our backend that we could easily parse.

The other benefit of this is that it also works with tab-completion since we list out all the paths specifically.

I didn't add search because I couldn't find a way to pre-search on /docs or the apiref through the url. We could do that as a follow up if we have a way to do this.

### DIFF
--- a/pkg/cmd/open.go
+++ b/pkg/cmd/open.go
@@ -1,0 +1,82 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/stripe/stripe-cli/pkg/open"
+	"github.com/stripe/stripe-cli/pkg/validators"
+)
+
+var nameURLmap = map[string]string{
+	"dashboard":                          "https://dashboard.stripe.com/",
+	"dashboard/payments":                 "https://dashboard.stripe.com/payments",
+	"dashboard/disputes":                 "https://dashboard.stripe.com/disputes",
+	"dashboard/balance":                  "https://dashboard.stripe.com/balance/overview",
+	"dashboard/payouts":                  "https://dashboard.stripe.com/payouts",
+	"dashboard/topups":                   "https://dashboard.stripe.com/topups",
+	"dashboard/transactions":             "https://dashboard.stripe.com/balance",
+	"dashboard/customers":                "https://dashboard.stripe.com/customers",
+	"dashboard/atlas":                    "https://dashboard.stripe.com/atlas",
+	"dashboard/radar":                    "https://dashboard.stripe.com/radar",
+	"dashboard/radar/reviews":            "https://dashboard.stripe.com/radar/reviews",
+	"dashboard/radar/list":               "https://dashboard.stripe.com/radar/list",
+	"dashboard/radar/rules":              "https://dashboard.stripe.com/radar/rules",
+	"dashboard/billing":                  "https://dashboard.stripe.com/billing",
+	"dashboard/invoices":                 "https://dashboard.stripe.com/invoices",
+	"dashboard/subscriptions":            "https://dashboard.stripe.com/subscriptions",
+	"dashboard/subscriptions/products":   "https://dashboard.stripe.com/subscriptions/products",
+	"dashboard/tax-rates":                "https://dashboard.stripe.com/tax-rates",
+	"dashboard/coupons":                  "https://dashboard.stripe.com/coupons",
+	"dashboard/connect":                  "https://dashboard.stripe.com/connect/overview",
+	"dashboard/connect/accounts":         "https://dashboard.stripe.com/connect/accounts/overview",
+	"dashboard/connect/transfers":        "https://dashboard.stripe.com/connect/transfers",
+	"dashboard/connect/collected-fees":   "https://dashboard.stripe.com/connect/application_fees",
+	"dashboard/orders":                   "https://dashboard.stripe.com/orders",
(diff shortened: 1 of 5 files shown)


mg-stripe:
Love the idea for tab completion!

pkg/cmd/open.go
```diff
@@ -0,0 +1,89 @@
...
+	"dashboard/terminal/hardware_orders": "https://dashboard.stripe.com/terminal/hardware_orders",
+	"dashboard/developers":               "https://dashboard.stripe.com/developers",
+	"dashboard/apikeys":                  "https://dashboard.stripe.com/apikeys",
+	"dashboard/webhooks":                 "https://dashboard.stripe.com/webhooks",
+	"dashboard/events":                   "https://dashboard.stripe.com/events",
+	"dashboard/logs":                     "https://dashboard.stripe.com/logs",
+	"dashboard/settings":                 "https://dashboard.stripe.com/settings",
+	"api":                                "https://stripe.com/docs/api",
+	"apiref":                             "https://stripe.com/docs/api",
+	"docs":                               "https://stripe.com/docs",
```
brandur-stripe: I hate to be that one-issue-person all the time, but any chance we could alphabetize this list? The reasons I think it's a good idea are (1) if you're looking for a particular URL in the list (say you're wondering if you have to add a new one), it's easier to find when things are sorted, and (2) it gives a definitive and deterministic place to add new URLs instead of just tacking them in randomly or on the end.

auchenberg:
Wooha, fast @tomer-stripe!

The list looks like a great starting point! +1 on the help command. Do we have `stripe help` or `stripe --help` to list all available commands? Enabling ``-help` as a arg on any "category" could be interesting too. @brandur-stripe

auchenberg-stripe:
Wooha, fast @tomer-stripe!

The list looks like a great starting point! +1 on the help command. Do we have `stripe help` or `stripe --help` to list all available commands? Enabling ``-help` as a arg on any "category" could be interesting too. @brandur-stripe

pkg/cmd/open.go
tomer-stripe: I definitely should've thought to do that, good call out.

tomer-stripe:
@brandur-stripe @auchenberg-stripe I started adding this to `--help` but ended up not doing that because the list is so large that it looked quite weird. I was also worried that if we expanded this list it'd just get worse since people will have to scroll through a massive list of shortcuts to corresponding urls in their terminal, which hasn't always felt super great to me with large lists like this.

I'm open to ideas if you have them! I was thinking of dropping a link to our readme or something that would show all the shortcuts (browsers have felt better for me for large lists like this). Maybe we could also shortcut it and say "We support dashboard links like dashboard/<product>, docs, and the apiref"?

For keeping the list up to date -- agreed there. I mentioned this in the description too but I think what I'd rather do is reference some externally hosted json file that has this mapping and is easier for us to update without a full release/deploy. This seemed faster to do now, although maybe what I could do is put the json file in GitHub and commit it to the repo locally, then try to read the raw file or fall back to the local one. That way keeping the list updated is at least a matter of updating a json file. What do you think>

brandur-stripe:
> I'm open to ideas if you have them! I was thinking of dropping a link to our readme or something that would show all the shortcuts (browsers have felt better for me for large lists like this). Maybe we could also shortcut it and say "We support dashboard links like dashboard/, docs, and the apiref"?

What if we kept the list out of `--help`, but instead put it in another option called `--list` that _just_ spits out a list of names and nothing else. That way a user could do something like this to find things they're looking for:

``` sh
stripe open --list | grep connect
```

We'd also mention the existence of `--list` from `--help`.

I stole this idea from rbenv, which has a very similar problem in that there are a hundred different versions of Ruby that can potentially be installed so they added a `--list`:

```
$ rbenv install --list | grep '2\.6'
  2.2.6
  2.6.0-dev
  2.6.0-preview1
  2.6.0-preview2
  2.6.0-preview3
  2.6.0-rc1
  2.6.0-rc2
  2.6.0
  2.6.1
  2.6.2
  2.6.3
  jruby-9.2.6.0
  rbx-2.2.6
  rbx-2.6
```

I'd assert pretty strongly that there needs to be _some_ other built-in route for discoverability beyond just autocomplete. Linking to something in the README might work, but IMO it's way more ideal if you can figure out how to use a command line invocation from the command line.

> For keeping the list up to date -- agreed there. I mentioned this in the description too but I think what I'd rather do is reference some externally hosted json file that has this mapping and is easier for us to update without a full release/deploy. This seemed faster to do now, although maybe what I could do is put the json file in GitHub and commit it to the repo locally, then try to read the raw file or fall back to the local one. That way keeping the list updated is at least a matter of updating a json file. What do you think?

Yeah, it's probably okay to skip this for now, but I'd probably skip it for now. One trouble is that I don't think we even have an automated way server-side to generate such a file — you could programmatically generate a list of all dashboard links, but you'd probably want to manually curate to filter noise anyway.

PR #127, opened 2019-08-21
Embed the OpenAPI spec once again

ob-stripe:
### Reviewers
r? @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
I stopped embedding the spec in #87 because I only needed it at compile time to generate the resource commands. But I think for generating the full help messages with the list of parameters, it's a lot easier to do this while parsing the spec at runtime. It won't impact performance in the general case because the spec will only be parsed when calling `--help` on an operation command, not every time the CLI is invoked.


ob-stripe:
I agree, that's what I had in mind as a long-term goal: translate the JSON/YAML spec into a 1:1 Go data structure at compile time, and deal with any data transformations at runtime (the transformations are fairly inexpensive, it's really the JSON parsing that takes a long time).

brandur-stripe:
> I agree, that's what I had in mind as a long-term goal: translate the JSON/YAML spec into a 1:1 Go data structure at compile time, and deal with any data transformations at runtime (the transformations are fairly inexpensive, it's really the JSON parsing that takes a long time).

Nice! Looking forward to it then.

PR #138, opened 2019-08-29
Partition API keys by test mode/live mode

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

### Summary
With this, I'm explicitly writing live/test mode API keys. I also changed the `GetAPIKey` to take a boolean that determines live/test mode.

In the profile logic, if we see that we're writing new API keys, I delete the old format ones. In the mean time, I've added another alias to be backwards compatible.

I also renamed the `--api-key` flag to be `--test-mode-api-key` but I'm not sure I like this. @ob-stripe @brandur-stripe what would you think about having this remain as `--api-key`?

### DIFF
--- a/pkg/login/poll.go
+++ b/pkg/login/poll.go
@@ -15,16 +15,19 @@ import (
 const maxAttemptsDefault = 2 * 60
 const intervalDefault = 1 * time.Second
 
-type pollAPIKeyResponse struct {
-	Redeemed           bool   `json:"redeemed"`
-	AccountID          string `json:"account_id"`
-	AccountDisplayName string `json:"account_display_name"`
-	APIKey             string `json:"testmode_key_secret"`
-	PublishableKey     string `json:"testmode_key_publishable"`
+// PollAPIKeyResponse returns the data of the polling client login
+type PollAPIKeyResponse struct {
+	Redeemed               bool   `json:"redeemed"`
+	AccountID              string `json:"account_id"`
+	AccountDisplayName     string `json:"account_display_name"`
+	TestModeAPIKey         string `json:"testmode_key_secret"`
+	TestModePublishableKey string `json:"testmode_key_publishable"`
 }
 
 // PollForKey polls Stripe at the specified interval until either the API key is available or we've reached the max attempts.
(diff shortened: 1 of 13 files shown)


ob-stripe:
> I also renamed the `--api-key` flag to be `--test-mode-api-key` but I'm not sure I like this. @ob-stripe @brandur-stripe what would you think about having this remain as `--api-key`?

Yeah, I would prefer to keep the flag as `--api-key` and have it be mutually exclusive with the future `--livemode` flag. If you're passing `--api-key`, you're overriding whichever API key (test or live) would have been used from the profile.

brandur-stripe:
Yeah, I'd +1 OB on that — `--test-mode-api-key` is getting to be pretty long.

tomer-stripe:
@ob-stripe @brandur-stripe fixed!

pkg/login/poll.go
```diff
@@ -15,16 +15,19 @@ import (
...
+	Redeemed               bool   `json:"redeemed"`
+	AccountID              string `json:"account_id"`
+	AccountDisplayName     string `json:"account_display_name"`
+	TestModeAPIKey         string `json:"testmode_key_secret"`
+	TestModePublishableKey string `json:"testmode_key_publishable"`
 }
 
 // PollForKey polls Stripe at the specified interval until either the API key is available or we've reached the max attempts.
-func PollForKey(pollURL string, interval time.Duration, maxAttempts int) (string, string, *Account, error) {
+func PollForKey(pollURL string, interval time.Duration, maxAttempts int) (PollAPIKeyResponse, *Account, error) {
```
ob-stripe: Will defer to Brandur, but I think it's a bit more idiomatic to return a pointer (`* PollAPIKeyResponse`) and return `nil` instead of an "empty" struct when returning an error.

tomer-stripe: oh, yes, I always do that. Changing now!

madeleineth:
I think this broke a test of mine with the error "API key is too short, must be at least 12 characters long" so I'm leaving that string in the comments as Google-bait, along with the note that I had to change `secret_key` in my config file to `test_mode_api_key`.

PR #140, opened 2019-09-04
Implement fixtures functionality

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 

 ### Summary
_I copied this PR over from #125 since that PR got a bit weird with a bunch of rebasing on the parent_

This implements a relatively simple `fixtures` command that takes a json file of a specific structure and makes requests to the API. The JSON files looks like:
```
{
	"_meta": {
		"template_version": "0"
	},
	"fixtures": [
		{
			"name": "cust_bender",
			"path": "/v1/customers",
			"method": "post",
			"params": {
				"email": "bender@planex.com",
				"address": {
					"city": "New New York"
				}
			}
		},
		{
			"name": "char_bender",
			"path": "/v1/charges",
			"method": "post",
			"params": {
				"customer": "${cust_bender:id}",
				"source": "tok_visa",
				"amount": "100",
				"currency": "usd",
				"capture": false
			}
		},
		{
			"name": "capt_bender",
			"path": "/v1/charges/${char_bender:id}/capture",
			"method": "post"
		}
	]
}
```

Where:
* `_meta` contains some information about the fixture, like the template version for the CLI to know if the fixture is compatible
* `fixtures` which contains a lists of API requests to make
* `fixtures[].name` identifies the request
* `fixtures[].path` stores the path for the request to be made
* `fixtures[].params` is a map of key/value pairs to put into the request
* `env` is a map of local .env values to overwrite (optional)

For making the requests, I'm using the pre-exiting requests package we built.

The way this works is by recursively parsing the data object and building the appropriate params (as well as correctly nesting fields for things like `address[city]`). This uses reflect to determine if the data is a string, int, map, or array. While the data is being parsed, anytime a string value comes up I try to parse it for a query, which is indicated by `${...}`.

The `${...}` queries are structured as: `${<NAME>:<JSON PATH>}`, where the json path is defined from the `gojsonq` library: https://github.com/thedevsaddam/gojsonq/wiki/Queries#findpath

URLs can also contain data from past queries by adding the query into the path`/v1/charges/${...}` in between slashes.

### DIFF
--- a/pkg/cmd/samples/fixtures.go
+++ b/pkg/cmd/samples/fixtures.go
@@ -38,13 +38,12 @@ func (fc *FixturesCmd) runFixturesCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	fixture := s.Fixture{
-		Fs:      afero.NewOsFs(),
-		APIKey:  apiKey,
-		BaseURL: stripe.DefaultAPIBaseURL,
-	}
-
-	err = fixture.NewFixture(args[0])
+	fixture, err := s.NewFixture(
+		afero.NewOsFs(),
+		apiKey,
+		stripe.DefaultAPIBaseURL,
+		args[0],
+	)
 	if err != nil {
 		return err
 	}
--- a/pkg/cmd/samples/fixtures.go
+++ b/pkg/cmd/samples/fixtures.go
@@ -43,7 +43,21 @@ func (fc *FixturesCmd) runFixturesCmd(cmd *cobra.Command, args []string) error {
 		APIKey:  apiKey,
 		BaseURL: stripe.DefaultAPIBaseURL,
 	}
-	fixture.NewFixture(args[0])
+
+	err = fixture.NewFixture(args[0])
+	if err != nil {
+		return err
+	}
+
+	err = fixture.Execute()
+	if err != nil {
+		return err
+	}
+
+	err = fixture.UpdateEnv()
+	if err != nil {
+		return err
+	}
 
 	return nil
 }
--- a/pkg/cmd/samples/fixtures.go
+++ b/pkg/cmd/samples/fixtures.go
@@ -1,0 +1,49 @@
+package samples
+
+import (
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/config"
+	s "github.com/stripe/stripe-cli/pkg/samples"
+	"github.com/stripe/stripe-cli/pkg/stripe"
+)
+
+// FixturesCmd prints a list of all the available sample projects that users can
+// generate
+type FixturesCmd struct {
+	Cmd *cobra.Command
+	Cfg *config.Config
+}
+
+// NewFixturesCmd creates and returns a list command for samples
+func NewFixturesCmd(cfg *config.Config) *FixturesCmd {
+	fixturesCmd := &FixturesCmd{
+		Cfg: cfg,
+	}
+
+	fixturesCmd.Cmd = &cobra.Command{
+		Use:   "fixtures",
+		Short: "Run fixtures to populate your account with data",
+		Long:  `Run fixtures to populate your account with data`,
+		RunE:  fixturesCmd.runFixturesCmd,
+	}
+
+	return fixturesCmd
+}
+
+func (fc *FixturesCmd) runFixturesCmd(cmd *cobra.Command, args []string) error {
+	apiKey, err := fc.Cfg.Profile.GetAPIKey(false)
+	if err != nil {
+		return err
+	}
+
+	fixture := s.Fixture{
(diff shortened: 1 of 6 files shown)


ob-stripe:
Haven't looked at the code yet, but just looking at the structure of the JSON, I think it's odd that `http.params` is a special hash to form URLs, while `data` is outside of the `http` hash. I also don't love the `#$` sigil, seems a bit too esoteric/random to me.

What do you think of something like this?

```json
{
  "_meta": {
    "template_version": "0"
  },
  "fixtures": [
    {
      "name": "cust_bender",
      "path": "/v1/customers",
      "method": "post",
      "params": {
        "email": "bender@planex.com",
        "address": {
          "city": "New New York"
        }
      }
    },
    {
      "name": "char_bender",
      "path": "/v1/charges",
      "method": "post",
      "params": {
        "customer": "${cust_bender:id}",
        "source": "tok_visa",
        "amount": "100",
        "currency": "usd",
        "capture": false
      }
    },
    {
      "name": "capt_bender",
      "path": "/v1/charges/${char_bender:id}/capture",
      "method": "post"
    }
  ]
}
```

pkg/cmd/samples/fixtures.go
```diff
@@ -0,0 +1,49 @@
...
+func (fc *FixturesCmd) runFixturesCmd(cmd *cobra.Command, args []string) error {
+	apiKey, err := fc.Cfg.Profile.GetAPIKey(false)
+	if err != nil {
+		return err
+	}
+
+	fixture := s.Fixture{
+		Fs:      afero.NewOsFs(),
+		APIKey:  apiKey,
+		BaseURL: stripe.DefaultAPIBaseURL,
```
ob-stripe: Could you add an `--api-base` flag? Would be consistent with other commands, and it's useful to test with stripe-mock / devboxes etc.

tomer-stripe:
@ob-stripe I like your template more! It feels a bit more intuitive, especially with less hierarchy and nesting. Will shift to that structure.

pkg/cmd/samples/fixtures.go
```diff
@@ -0,0 +1,49 @@
...
+		Short: "Run fixtures to populate your account with data",
+		Long:  `Run fixtures to populate your account with data`,
+		RunE:  fixturesCmd.runFixturesCmd,
+	}
+
+	return fixturesCmd
+}
+
+func (fc *FixturesCmd) runFixturesCmd(cmd *cobra.Command, args []string) error {
+	apiKey, err := fc.Cfg.Profile.GetAPIKey(false)
```
adreyfus-stripe: Do we have metrics or tracking yet? This would be a good one to track

tomer-stripe: Not yet! We're still trying to figure out how to do this :(

tomer-stripe:
ptal @ob-stripe 

All feedback should be addressed!

pkg/samples/fixtures.go
```diff
@@ -0,0 +1,283 @@
...
+type Fixture struct {
+	Fs        afero.Fs
+	APIKey    string
+	BaseURL   string
+	responses map[string]*gojsonq.JSONQ
+	fixture   fixtureFile
+}
+
+// NewFixture creates and executes a fixtures steps for populating test data
+func (fxt *Fixture) NewFixture(file string) error {
```
ob-stripe: I think it's conventional for functions named `NewFoo` to initialize and return a new `Foo` struct. That's not what's happening here, so I think we should change it in one of two ways:
1. rename the method to `Load` or `LoadFromFile` or something
2. keep `NewFixture`, but make it a non-struct method that actually initialize and returns a new `Fixture`

I have a slight preference for the second way, but up to you.

pkg/samples/fixtures.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	return nil
+}
+
+func (fxt *Fixture) makeRequest(data fixture) ([]byte, error) {
+	var rp requests.RequestParameters
+	if data.Method == "post" {
+		now := time.Now().String()
+		metadata := fmt.Sprintf("metadata[_created_by_fixture]=%s", now)
```
ob-stripe: Some resources might not accept metadata, so this is slightly risky. We can keep this as a documented limitation for now, but long term we should check the OpenAPI spec to decide whether or not to include metadata.

pkg/samples/fixtures.go
tomer-stripe: Ahh yea, sorry about that! Will change that now.

pkg/samples/fixtures.go
tomer-stripe: I thought about that too, would the API fail here if we passed metadata to a resource that doesn't support it?

ob-stripe: Yes, it would return a 401 `invalid_request` error.

PR #143, opened 2019-09-05
Capitalize the language names in the selection prompt

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
Request from @romain-stripe to properly capitalize the languages in the select prompt. I hardcode mappings from the folder name to the display name and vice versa to map the selections the user makes.

### DIFF
--- a/pkg/samples/samples.go
+++ b/pkg/samples/samples.go
@@ -40,13 +40,7 @@ var languageDisplayNames = map[string]string{
 	"ruby":   "Ruby",
 }
 
-var displayNameLanguages = map[string]string{
-	"Java":   "java",
-	"Node":   "node",
-	"Python": "python",
-	"PHP":    "php",
-	"Ruby":   "ruby",
-}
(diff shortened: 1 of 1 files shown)


pkg/samples/samples.go
```diff
@@ -32,6 +32,22 @@ var fileEnvTemplate = map[string]func(string) string{
...
+	"ruby":   "Ruby",
+}
+
+var displayNameLanguages = map[string]string{
+	"Java":   "java",
+	"Node":   "node",
+	"Python": "python",
+	"PHP":    "php",
+	"Ruby":   "ruby",
+}
```
brandur-stripe: Hm, I wonder if we should have a routine to just reverse the other map on `init()` or something.

brandur-stripe: (Just because it seems a little undesirable that you could add something to one and forget to add it to the other.)

tomer-stripe: Fixed!

PR #149, opened 2019-09-08
Do not display or print filtered events

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
Event filtering only filters events when POSTing requests, but it should also prevent them from being displayed at all. This PR handles that case by setting up a map to check against when processing the webhook event

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -47,9 +47,6 @@ type Config struct {
 	// EndpointsMap is a mapping of local webhook endpoint urls to the events they consume
 	EndpointRoutes []EndpointRoute
 
-	// Events is the supported event types for the command
-	Events []string
-
 	APIBaseURL string
 
 	// WebSocketFeature is the feature specified for the websocket connection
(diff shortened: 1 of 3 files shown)


pkg/proxy/proxy.go
```diff
@@ -63,6 +66,8 @@ type Config struct {
 
 	// Force use of unencrypted ws:// protocol instead of wss://
 	NoWSS bool
+
+	supportedEvents map[string]bool
```
ob-stripe: I think this should be a property of `Proxy`, not `Config`. Maybe we should have a `NewProxy` function to initialize the `Proxy` struct and do the conversion from slice to map.

pkg/proxy/proxy.go
```diff
@@ -169,24 +178,26 @@ func (p *Proxy) processWebhookEvent(msg websocket.IncomingMessage) {
...
-		"Received %sevent: %s [type: %s]",
-		maybeConnect,
-		ansi.Linkify(evt.ID, evt.urlForEventID(), p.cfg.Log.Out),
-		ansi.Linkify(evt.Type, evt.urlForEventType(), p.cfg.Log.Out),
-	)
-
-	if p.cfg.PrintJSON {
-		fmt.Println(webhookEvent.EventPayload)
-	}
+	if p.cfg.supportedEvents[evt.Type] {
```
tomer-stripe: This is also broken, we need to support `*` as well.

tomer-stripe:
ptal @ob-stripe 

Resolved!

Issue #43, opened 2019-07-17
AVG AntiVirus flags the CLI as suspicious


tomer-stripe:
Reported by @homemadePopcorn in #37:

> stripe.exe was labelled as a suspicious file by my anti-virus software. I can add an exception to that file on my firewall but just thought I would write the issue down.

Is there something we can do to make Windows happier with that behavior? We didn't have issues with this when we were doing Windows testing locally, wonder if there's something specific it's flagging.

ob-stripe:
@homemadePopcorn Can you tell a little more about your environment? I.e. Windows version, antivirus name and version, or anything else that could help in reproducing the issue.

homemadePopcorn:
Sure, I'm using AVG Anti-virus software v19.6.3098.  Windows 10 pro v1903.

ob-stripe:
Thanks!

I just uploaded `stripe.exe` from the 0.2.2 release to virustotal.com, here are the results: https://www.virustotal.com/gui/file/b9bb89ba261974e9bfe2146f76fffe98d8e3681eb249e59835b132c9b5f6ec81/detection

AVG reports no issues there, but you might have different settings on your computer.

My guess is that AVG is being overly paranoid because this is a new executable it's never seen before. Hopefully once enough people start using the CLI it will get whitelisted. If this becomes a recurring issue we can also see about directly reaching out antivirus software providers.

ob-stripe:
Oh, I wonder if this is because we try to run `uname -a` to compute the extended user agent. We can probably exclude that entirely on Windows since the chances of `uname` existing on Windows are very slim.

tomer-stripe:
Going to close this since we think we might've addressed the underlying issue! We can re-open if it surfaces again.

PR #170, opened 2019-09-17
#43: use uname syscall instead of shelling out

joek-stripe:
### Reviewers
r? @ob-stripe 
cc @stripe/dev-platform

 ### Summary
Our CLI was being flagged by AVG anti-malware tool, it was suspected this was because we were shelling out to `uname` to retrieve info. 

In my experience determining why a particular AV would choose to flag a golang binary is opaque, still though using the uname syscall is preferable here over shelling out.

### DIFF
--- a/pkg/useragent/useragent.go
+++ b/pkg/useragent/useragent.go
@@ -1,10 +1,11 @@
 package useragent
 
 import (
-	"bytes"
 	"encoding/json"
-	"os/exec"
+	"fmt"
 	"runtime"
+
+	"golang.org/x/sys/unix"
 
 	"github.com/stripe/stripe-cli/pkg/version"
 )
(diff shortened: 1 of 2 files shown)


pkg/useragent/useragent.go
```diff
@@ -61,25 +61,21 @@ var encodedUserAgent string
...
+		err := unix.Uname(u)
+		if err != nil {
+			panic(err)
+		}
+
+		return fmt.Sprintf("%s %s %s %s %s", u.Sysname, u.Nodename, u.Release, u.Version, u.Machine)
 	}
 
-	return out.String()
+	return ""
```
tomer-stripe: Instead of returning nothing could we try to get the hostname for windows?

ob-stripe: `uname` is used to get information about the OS/kernel, not the hostname. Windows doesn't have a great/easy to use equivalent.

pkg/useragent/useragent_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"runtime"
+	"testing"
+)
+
+func TestUnameNotEmpty(t *testing.T) {
+	if runtime.GOOS != "windows" {
+		u := getUname()
+
+		if u == "" {
+			t.Error("Failed to pull unix name")
```
ob-stripe: We use the `require` library in our tests to make it easier to test various assertions. Can you change this to:
```go
require.NotEmpty(t, u);
```

joek-stripe: done

PR #172, opened 2019-09-18
Change listen shorthand json

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
Avoid conflict with the `-p` for `--project-name`


brandur-stripe:
You may or may not have thought of this already, but just a note that this CLI's likely to get big enough that you're going to run out of letters eventually :)

It might be a good idea to start defaulting to not assigning shorthands for most options, and only assigning them when you're pretty confident that an options is going to be one that _humans_ will want to use as part of common workflows.

Easy non-shorthand options are those, for example, that are most likely to be put into scripts calling into the Stripe CLI where no one has to type them.

PR #173, opened 2019-09-18
Add a dark color scheme for json colorize

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
Adds a dark color scheme for terminals with lighter background colors:

Screen Shot 2019-09-17 at 8 26 04 PM

A bit limited by what ansi colors are supported so it's not perfect, but it's a step up.

### DIFF
--- a/pkg/requests/base.go
+++ b/pkg/requests/base.go
@@ -45,6 +45,8 @@ type Base struct {
 	// SuppressOutput is used by `trigger` to hide output
 	SuppressOutput bool
 
+	DarkStyle bool
+
 	APIBaseURL string
 
 	Livemode bool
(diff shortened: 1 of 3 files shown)


pkg/requests/base.go
```diff
@@ -93,6 +95,7 @@ func (rb *Base) InitFlags(includeData bool) {
 	rb.Cmd.Flags().BoolVarP(&rb.showHeaders, "show-headers", "s", false, "Show headers on responses to GET, POST, and DELETE requests")
 	rb.Cmd.Flags().BoolVarP(&rb.autoConfirm, "confirm", "c", false, "Automatically confirm the command being entered. WARNING: This will result in NOT being prompted for confirmation for certain commands")
 	rb.Cmd.Flags().BoolVar(&rb.Livemode, "livemode", false, "Make a request against live mode instead of test mode")
+	rb.Cmd.Flags().BoolVar(&rb.DarkStyle, "dark-style", false, "Use a darker color scheme better suited for lighter terminals")
```
brandur-stripe: Don't feel too strongly on this, but what do you think about `--theme dark` or something in case you decide to add more themes in the future?

tomer-stripe: It feels a little weird to me to have a flag with just one text option right now so I'm thinking of keeping the bool. It should be easy for us to map this to a theme in the future if we need to!

PR #174, opened 2019-09-18
Refactored examples.go to use test tokens instead of test card data

joek-stripe:
### Reviewers
r? @ob-stripe 
cc @stripe/dev-platform

 ### Summary
`stripe trigger` command was resulting in invalid data in the dashboard. This was due to it attempting to tokenize raw card data (test cards).

Refactored examples.go to use [test tokens](https://stripe.com/docs/testing#cards) instead and added request body checks in examples_test.go so we will catch something like this in future

### DIFF
--- a/pkg/requests/examples.go
+++ b/pkg/requests/examples.go
@@ -8,6 +8,11 @@ import (
 	"github.com/stripe/stripe-cli/pkg/config"
 )
 
+const (
+	validToken    = "tok_visa"
+	declinedToken = "tok_chargeDeclined"
+)
+
 func parseResponse(response []byte) (map[string]interface{}, error) {
 	var result map[string]interface{}
 	err := json.Unmarshal(response, &result)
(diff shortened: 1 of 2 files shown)


pkg/requests/examples.go
```diff
@@ -86,25 +74,25 @@ func (ex *Examples) chargeCreated(card string, data []string) (map[string]interf
...
 	reqURL := fmt.Sprintf("/v1/charges/%s/capture", charge["id"])
-
 	_, err = ex.performStripeRequest(req, reqURL, params)
 	return err
 }
 
 // ChargeFailed fails to create a charge
 func (ex *Examples) ChargeFailed() error {
-	_, err := ex.chargeCreated("4000000000000002", []string{
+	_, err := ex.chargeCreated(dummyToken, []string{
```
ob-stripe: `tok_visa` (the value of `dummyToken`) is equivalent to the `4242424242424242` test number that results in a successful charge, so you can't use for the `ChargeFailed` function.

The test token that matches the `4000000000000002` test number (that results in a failed charge) is `tok_chargeDeclined` (cf. https://stripe.com/docs/testing#cards-responses).

joek-stripe: 🤦‍♂ fixing, sorry

joek-stripe: pushed fix to remote fork

PR #176, opened 2019-09-18
display help page on unknown commands

stephchen-stripe:
### Reviewers
r? @ob-stripe 
cc @stripe/dev-platform

 ### Summary
show help page when `stripe` receives an unknown command

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -55,9 +55,17 @@ func Execute() {
 	rootCmd.SetVersionTemplate(version.Template)
 	if err := rootCmd.Execute(); err != nil {
 		if strings.Contains(err.Error(), "unknown command") {
-			rootCmd.Help()
+			suggestions := rootCmd.SuggestionsFor(os.Args[1])
+			sugg_str := ""
+			if len(suggestions) > 0 {
+				sugg_str = fmt.Sprintf(" Did you mean \"%s\"?", suggestions[0])
+			}
+			fmt.Println(fmt.Sprintf("Unknown command \"%s\" for \"%s\".%s\n" +
+				"If not, see \"stripe --help\" for a list of available commands.",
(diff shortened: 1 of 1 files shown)


ob-stripe:
This works, but I'm worried the output is too verbose since the help is so long. Maybe we should do something similar to `git`:

```sh
$ git checkot
git: 'checkot' is not a git command. See 'git --help'.

The most similar command is
	checkout
```

@tomer-stripe wdyt?

stephchen-stripe:
@ob-stripe so we have the "closest match" correction enabled just as with git although the edit distance is currently set to 2 (so it doesn't catch "configure" => "config" as in the original ticket). we could just adjust the current error message to tell the user to go to "--help" (it currently doesn't) instead of displaying the whole help page

ob-stripe:
> we could just adjust the current error message to tell the user to go to "--help" (it currently doesn't) instead of displaying the whole help page

Yes, that's what I meant. It shouldn't be too hard to catch the "unknown command" error as you're currently doing and then reconstructing our own customized error message using the same functions as cobra itself: https://github.com/spf13/cobra/blob/f619abc1d7edb30544d1c956719d203ace73e9b1/args.go#L21

stephchen-stripe:
flow now looks like this:
```
st-stephchen1:stripe-cli stephchen$ stripe-dev configu
Unknown command "configu" for "stripe". Did you mean "config"?
If not, see "stripe --help" for a list of available commands.
```

stephchen-stripe:
flow now looks like this
```
st-stephchen1:stripe-cli stephchen$ stripe-dev configu
Unknown command "configu" for "stripe". Did you mean "config"?
If not, see "stripe --help" for a list of available commands.
exit status 1
st-stephchen1:stripe-cli stephchen$ stripe-dev configuasdf
Unknown command "configuasdf" for "stripe".
See "stripe --help" for a list of available commands.
exit status 1
```

PR #180, opened 2019-09-18
Check for ssh clients when logging in

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
This checks if the CLI is possibly running in an SSH client and if it is, prints the url instead of trying to open the browser (pulled from here: https://unix.stackexchange.com/questions/9605/how-can-i-detect-if-the-shell-is-controlled-from-ssh)

I tried to see if there was a good test to add here but since we're not catching stdout I wasn't sure of a decent way to validate this. Open to ideas!

### DIFF
--- a/pkg/login/client_login.go
+++ b/pkg/login/client_login.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 
+	"github.com/briandowns/spinner"
 	"github.com/stripe/stripe-cli/pkg/ansi"
 	"github.com/stripe/stripe-cli/pkg/config"
 	"github.com/stripe/stripe-cli/pkg/open"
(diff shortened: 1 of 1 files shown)


tomer-stripe:
> Is the general idea here that if SSHed, the open command (or Linux equivalent) will still work, but it'll just open it on the wrong system?
Yea, that's a possibility here! It's come up a few times that people have been ssh'd on a box and tried to login but got the "opening browser" prompt.

ob-stripe:
Yeah, we need to come up with a standard way of passing stdin/out/err handles to all CLI commands so we can replace them with buffers in test.

PR #181, opened 2019-09-19
Add feature to resend webhook based on event-id

joek-stripe:
### Reviewers
r? @ob-stripe
cc @stripe/dev-platform

 ### Summary
The CLI should be able to replay the exact webhook being sent to a user within a few minutes of receiving it.

e.g.
```
stripe trigger --event evt_123
```

DX-3594

### DIFF
--- a/pkg/cmd/trigger.go
+++ b/pkg/cmd/trigger.go
@@ -17,13 +17,14 @@ type triggerCmd struct {
 	cmd *cobra.Command
 
 	apiBaseURL string
+	eventID    string
 }
 
 func newTriggerCmd() *triggerCmd {
 	tc := &triggerCmd{}
 	tc.cmd = &cobra.Command{
 		Use:  "trigger <event>",
-		Args: validators.ExactArgs(1),
+		Args: validators.MaximumNArgs(1),
 		ValidArgs: []string{
 			"charge.captured",
 			"charge.failed",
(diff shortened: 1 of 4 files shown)


pkg/cmd/trigger.go
```diff
@@ -65,7 +66,9 @@ needed to create the triggered event.
   payment_intent.created
   payment_intent.payment_failed
   payment_intent.succeeded
-  payment_method.attached`,
+  payment_method.attached
+
+  --event <evt_id>`,
```
ob-stripe: We should improve the help message here. Something like this:
```

You can also resend a past event using the --event flag:
  stripe trigger --event evt_123
```

pkg/cmd/trigger.go
```diff
@@ -76,6 +79,7 @@ needed to create the triggered event.
 	// Hidden configuration flags, useful for dev/debugging
 	tc.cmd.Flags().StringVar(&tc.apiBaseURL, "api-base", stripe.DefaultAPIBaseURL, "Sets the API base URL")
 	tc.cmd.Flags().MarkHidden("api-base") // #nosec G104
+	tc.cmd.Flags().StringVar(&tc.eventID, "event", "", "event-id to resend")
```
ob-stripe:
```suggestion
	tc.cmd.Flags().StringVar(&tc.eventID, "event", "", "ID of the event to resend")
```

pkg/cmd/trigger.go
joek-stripe: I'll improve the help messages, but for the livemode I'll do that in a later PR.

PR #184, opened 2019-09-19
`get` command supports being given arbitrary ids

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
The `get` and `delete` commands support being given a specific list of ids without a path and the command will automatically prepend the correct path and work:

`stripe get cus_12345` => 200

This maintains the current url normalizing functionality as-is.


pkg/requests/ids.go
```diff
@@ -0,0 +1,51 @@
+package requests
+
+import "regexp"
+
+var idURLMap = map[string]string{
+	"txn":   "/v1/balance_transactions/",
+	"ch":    "/v1/charges/",
```
ob-stripe: Can you add an entry for `py` that also points to `/v1/charges`? (Some charge objects have a `py_` prefix because Reasons™)

pkg/requests/ids.go
```diff
@@ -0,0 +1,51 @@
...
+	"tml":   "/v1/terminal/locations/",
+	"tmr":   "/v1/terminal/readers/",
+	"or":    "/v1/orders/",
+	"orret": "/v1/order_returns/",
+	"sku":   "/v1/skus/",
+	"sqr":   "/v1/sigma/scheduled_query_runs/",
+	"we":    "/v1/webhook_endpoints/",
+}
+
+var idRegex = regexp.MustCompile("^([a-z]{2,5})_(test_|live_)?[a-zA-Z0-9]{3,}$")
```
ob-stripe: Do we actually use `test_|live_` in IDs?

pkg/requests/ids.go
tomer-stripe: 😬 yep!

pkg/requests/ids.go
tomer-stripe: I think so! That's how our api ref documents checkout sessions: https://stripe.com/docs/api/checkout/sessions/retrieve

ob-stripe: 👍

PR #186, opened 2019-09-19
Add instructions for what to do after generating completion files

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
Add output/instructions for what to do with completion files:
```
» stripe-dev completion
Generating bash completion file in: stripe-completion.bash
You'll need to either source this file or move it to bash's autocomplete folder.
To source, run: `source stripe-completion.bash`
To load automatically, move to zsh's autocomplete directory: `mv stripe-completion.bash /usr/local/share/zsh/site-functions`

### DIFF
--- a/pkg/cmd/completion.go
+++ b/pkg/cmd/completion.go
@@ -1,8 +1,12 @@
 package cmd
 
 import (
+	"fmt"
+	"runtime"
+
 	"github.com/spf13/cobra"
 
+	"github.com/stripe/stripe-cli/pkg/ansi"
 	"github.com/stripe/stripe-cli/pkg/validators"
 )
 
@@ -10,6 +14,17 @@ type completionCmd struct {
 	cmd *cobra.Command
 
 	shell string
+}
+
+var installDirectory = map[string]map[string]string{
+	"darwin": map[string]string{
+		"zsh":  "/usr/local/share/zsh/site-functions",
+		"bash": "/usr/local/etc/bash_completion.d",
+	},
+	"linux": map[string]string{
+		"zsh":  "/usr/share/zsh/functions/Completion/Linux",
+		"bash": "/usr/share/bash-completion/completions",
+	},
 }
 
 func newCompletionCmd() *completionCmd {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/cmd/completion.go [hunk 1]
```diff
@@ -20,10 +35,28 @@ func newCompletionCmd() *completionCmd {
 		Short: "Generate bash and zsh completion scripts",
 		Args:  validators.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			userOS := runtime.GOOS
+			if userOS == "windows" {
+				fmt.Println("Warning: autocompletion for Windows is currently not supported and may not work.")
+				// Give them instructions for linux because Windows _might_ be using the linux fs under the hood
```
ob-stripe: I think this is fine, but just to clarify: with WSL (Windows Subsystem for Linux), it's the Linux executable that is being run, and `runtime.GOOS` would (correctly) return `"linux"`, and the shell would actually be `bash` or `zsh` or some other Linux shell.

pkg/cmd/completion.go
```diff
@@ -12,6 +16,17 @@ type completionCmd struct {
...
 }
 
+var installDirectory = map[string]map[string]string{
+	"darwin": map[string]string{
+		"zsh":  "/usr/local/share/zsh/site-functions",
+		"bash": "/usr/local/etc/bash_completion.d",
+	},
+	"linux": map[string]string{
+		"zsh":  "/usr/share/zsh/functions/Completion/Linux",
+		"bash": "/usr/share/bash-completion/completions",
```
ob-stripe: I'm not sure we should be making assumptions about directories in the CLI. Apparently on Linux `/etc/bash_completion.d/` can also be used.

tomer-stripe:
@ob-stripe probably, yea. I was worried this would miss a ton of different configurations (and looks like it has. Unsure what the best thing to do here since describing them seems complicated and possibly error prone. I wonder if we can just say "completions are here, installations are different depending on operating system and shell please look them up"? That feels unsatisfying though :(

ob-stripe:
Maybe we can write detailed instructions as wiki articles, and just print a short message and a link to the wiki?

tomer-stripe:
Yea that might make more sense. Or at least a wiki page with resources. I'll close this for now

PR #192, opened 2019-09-20
Print options

mimi-stripe:
### Reviewers
r? @ob-stripe 
cc @stripe/dev-platform

 ### Summary
When running listen with options, prints the options to console so user can see what events are being forwarded, URL they are forwarding to, etc.

Wondering if we should hide the output behind a -V verbose flag?
The current output looks like this:

Screen Shot 2019-09-20 at 1 43 02 PM

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -138,6 +138,8 @@ func (p *Proxy) Run() error {
 		p.cfg.Log.Fatalf("Aborting")
 	}
 
+	p.printOptions()
+
 	// Block until context is done (i.e. Ctrl+C is pressed)
 	<-p.ctx.Done()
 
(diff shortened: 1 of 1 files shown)


ob-stripe:
@mimi-stripe Can you share a screenshot of what the output looks like with a bunch of options?

mimi-stripe:
> @mimi-stripe Can you share a screenshot of what the output looks like with a bunch of options?

Yes! Sorry I forgot to attach it!

ob-stripe:
I haven't forgotten about this, but we're a little bit on the fence about the format to use or even if we want to display this by default. cc @tomer-stripe @mg-stripe @romain-stripe for opinions.

mg-stripe:
A few notes:
  - We should always use sentence case (_Using Latest API Version _ -> _Using latest API version_).
  - Should we also include the actual version number, rather than just _latest_ or _default_?
  - I agree that this is probably too verbose as a default. For example, if someone hasn't specific Connect as an option it can be weird to show that configuration option every time (especially when it's simply named _Connect_: we may want a more precise name).

ob-stripe:
Closing for now since we're not sure this is a desirable feature. We'll revisit if we get requests for this.

PR #194, opened 2019-09-23
provide warning if user includes an invalid event in `listen --events`

stephchen-stripe:
### Reviewers
r? @ob-stripe 
cc @stripe/dev-platform

 ### Summary
generates list of valid events from the API spec. when a user tries to listen to an invalid event, prints a warning but otherwise does nothing. in response to https://jira.corp.stripe.com/browse/DX-4705


pkg/cmd/events_list.go
```diff
@@ -0,0 +1,147 @@
+// This file is generated; DO NOT EDIT.
+
+package cmd
+
+func addEventsToListenCmd(cmd *listenCmd) {
```
ob-stripe: I don't think you need a function here. You can simply declare a global map that holds the values:

```go
ValidEvents := map[string]bool {
  "*": true,
  "account.application.authorized": true,
  // etc.
}
```

then access it directly from `listenCmd`:

```golang
	for _, event := range lc.events {
		if _, found := ValidEvents[event]; !found {
			fmt.Println(fmt.Sprintf("Warning: You're attempting to listen for \"%s\", which isn't a valid event", event))
		}
	}
```

tomer-stripe: Agree with @ob-stripe on accessing directly, it feels unnecessary to define a function here for that.

PR #195, opened 2019-09-24
Send forward URL in webhook responses

ob-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform @ballantyne-stripe @carolinem-stripe 

 ### Summary
Include the forward URL in webhook responses, so we can use it to display it in the dashboard.


tomer-stripe:
lgtm!

PR #199, opened 2019-09-25
Display error properties when tailing logs

mimi-stripe:
[wip] 

### Reviewers
r? @ob-stripe 
cc @stripe/dev-platform

 ### Summary
Changes to tailer to allow 'logs tail' to accept error info from request log events, and print to user.

### DIFF
--- a/pkg/logtailing/tailer.go
+++ b/pkg/logtailing/tailer.go
@@ -68,17 +68,17 @@ type Tailer struct {
 
 // EventPayload is the mapping for fields in event payloads from request log tailing
 type EventPayload struct {
-	CreatedAt int    `json:"created_at"`
-	Livemode  bool   `json:"livemode"`
-	Method    string `json:"method"`
-	RequestID string `json:"request_id"`
-	Status    int    `json:"status"`
-	URL       string `json:"url"`
-	Error     string `json:"error"`
+	CreatedAt int             `json:"created_at"`
+	Livemode  bool            `json:"livemode"`
+	Method    string          `json:"method"`
+	RequestID string          `json:"request_id"`
+	Status    int             `json:"status"`
+	URL       string          `json:"url"`
+	Error     RequestLogError `json:"error"`
 }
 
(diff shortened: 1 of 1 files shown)


pkg/logtailing/tailer.go
```diff
@@ -74,6 +74,17 @@ type EventPayload struct {
 	RequestID string `json:"request_id"`
 	Status    int    `json:"status"`
 	URL       string `json:"url"`
+	Error     string `json:"error"`
```
ob-stripe: I think this should be typed as `RedactedError` instead of `string`.

pkg/logtailing/tailer.go
```diff
@@ -74,6 +74,17 @@ type EventPayload struct {
...
+}
+
+// RedactedError is the mapping for fields in error from an EventPayload
+type RedactedError struct {
```
ob-stripe: We can probably just call this `RequestLogError` or something. From the CLI's perspective it's not really important that the error is "redacted".

pkg/logtailing/tailer.go
```diff
@@ -190,6 +201,15 @@ func (tailer *Tailer) processRequestLogEvent(msg websocket.IncomingMessage) {
 	color := ansi.Color(os.Stdout)
 	outputStr := fmt.Sprintf("%s [%d] %s %s [%s]", color.Faint(localTime), coloredStatus, payload.Method, payload.URL, requestLink)
 	fmt.Println(outputStr)
+
+	var payloadError RedactedError
+	if err := json.Unmarshal([]byte(payload.Error), &payloadError); err != nil {
```
ob-stripe: Assuming the stripecli service sends the `error` field as a hash, not as a string, we don't need to unmarshal it ourselves. Just unmarshaling the top-level `EventPayload` will do the right thing.

mimi-stripe: Oh cool, I didn't know it would work recursively like that, nice!

PR #200, opened 2019-09-26
pass simple parameters to operation commands in as flags instead of URL-style query args

stephchen-stripe:
### Reviewers
r? @ob-stripe 
cc @stripe/dev-platform

 ### Summary
on operation commands such as `stripe customer create`, we currently pass in details such as customer shipping address or email as URL-style `email=a@a.com` query args, e.g. `stripe customer create email=a@a.com balance=5`. this changes the operation interface so that string and int parameters are taken as flags, e.g. `stripe customer create --email a@a.com --balance 5`. nonscalar parameters are still taken in the old format but must be under a `-d` flag, e.g. `stripe customer create --balance 5 -d "shipping[name]=name"`. only parameters in the request URL are still accepted as args.

### DIFF
--- a/pkg/cmd/resource/operation.go
+++ b/pkg/cmd/resource/operation.go
@@ -46,7 +46,7 @@ func (oc *OperationCmd) runOperationCmd(cmd *cobra.Command, args []string) error
 	for stringProp, stringVal := range oc.stringPropFlags {
 		// only include fields explicitly set by the user to avoid conflicts between e.g. account_balance, balance
 		if oc.Cmd.Flags().Changed(stringProp) {
-			flagParams = append(flagParams, fmt.Sprintf("%s=\"%s\"", stringProp, *stringVal))
+			flagParams = append(flagParams, fmt.Sprintf("%s=%s", stringProp, *stringVal))
 		}
 	}
 	for intProp, intVal := range oc.intPropFlags {
--- a/pkg/cmd/resource/operation.go
+++ b/pkg/cmd/resource/operation.go
@@ -40,9 +40,23 @@ func (oc *OperationCmd) runOperationCmd(cmd *cobra.Command, args []string) error
 		return err
 	}
 
-	path := formatURL(oc.Path, args[:len(oc.URLParams)])
+	path := formatURL(oc.Path, args[:len(oc.URLParams)]) 	// @@ todo validate # of params
 
-	oc.Parameters.AppendData(args[len(oc.URLParams):]) 	// @@ todo this
+	flagParams := make([]string, 0)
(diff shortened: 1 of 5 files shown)


pkg/cmd/resource/operation.go
```diff
@@ -29,6 +29,11 @@ type OperationCmd struct {
 	HTTPVerb  string
 	Path      string
 	URLParams []string
+
+	stringFlags  map[string]*string
+	intFlags     map[string]*int
```
stephchen-stripe: if it'd be better these could probably be merged into one map of strings

pkg/cmd/resource/operation_test.go
```diff
@@ -38,7 +38,7 @@ func TestRunOperationCmd(t *testing.T) {
 		require.Equal(t, http.MethodPost, r.Method)
 		require.Equal(t, "/v1/bars/bar_123", r.URL.Path)
 		require.Equal(t, "Bearer sk_test_1234", r.Header.Get("Authorization"))
-		require.Equal(t, "param1=value1&param2=value2", string(body))
+		require.True(t, string(body) == "param1=value1&param2=value2" || string(body) == "param2=value2&param1=value1")
```
stephchen-stripe: because ordering in maps isn't guaranteed

pkg/cmd/resource/operation.go
```diff
@@ -29,6 +29,11 @@ type OperationCmd struct {
 	HTTPVerb  string
 	Path      string
 	URLParams []string
+
+	stringFlags map[string]*string
+	intFlags    map[string]*int
```
ob-stripe: I think it would be better to just store every scalar parameter as a string, and let the API return an error message when an invalid type is used. The error message returned by cobra/pflag isn't super explicit or helpful:

```
$ go run cmd/stripe/main.go charges create --source=tok_visa --currency=usd -s --amount=abc
invalid argument "abc" for "--amount" flag: strconv.ParseInt: parsing "abc": invalid syntax
exit status 1
```

pkg/cmd/resource/operation_test.go
ob-stripe: Oof, good point. We should deserialize the form encoded string using [`ParseQuery`](https://golang.org/pkg/net/url/#ParseQuery) instead and test the individual values. Would make it easier to test when there are more than two keys in the query string.

ob-stripe:
ptal @stephchen-stripe

stephchen-stripe:
sounds good @ob-stripe -- i was looking to display them separately but didn't figure that out. thanks for mentoring!

PR #211, opened 2019-10-01
Group request parameters flags separately in help messages

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Group request parameters flags separately from "normal" flags in help messages.

Couple of screen shots:

image

image


tomer-stripe:
@ob-stripe do you think the request params should be `--amount=` instead of just `--amount`? The way it is right now makes it look like a bunch of boolean flags.

PR #226, opened 2019-10-09
Fix config file permissions to not be world-readable

ob-stripe:
### Reviewers
r? @brandur-stripe  @tomer-stripe @joek-stripe 
cc @stripe/dev-platform

 ### Summary
Use 0600 (`rw-------`) instead of the default 0644 (`rw-r--r--`) for the default configuration file. Since it stores API keys (and in the future may store _live_ API keys), it's important to keep it secure.

I used viper's `SetConfigPermissions` method, but that only applies to creating new files, so I also added a manual permissions change to fix existing files.

Not that this only applies to the default configuration file. If the user provides a custom file with `--config`, it's their responsibility to secure the file.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -76,6 +76,14 @@ func (c *Config) InitConfig() {
 		c.ProfilesFile = configFile
 		viper.SetConfigType("toml")
 		viper.SetConfigFile(configFile)
+		viper.SetConfigPermissions(os.FileMode(0600))
+
+		// Try to change permissions manually, because we used to create files
+		// with default permissions (0644)
+		err := os.Chmod(configFile, os.FileMode(0600))
+		if err != nil && !os.IsNotExist(err) {
+			log.Fatalf("%s", err)
+		}
 	}
 
 	// If a profiles file is found, read it in.


brandur-stripe:
LGTM.

I think you might have gotten hit by a new version in Golint or something. There's a bunch of failures in CI of this nature:

```
./bin/golangci-lint run ./...
pkg/ansi/ansi.go:133:2: return statements should not be cuddled if block has more than two lines (wsl)
	return s
	^
pkg/ansi/markdownterm.go:60:2: expressions should not be cuddled with blocks (wsl)
	out.WriteString("\n")
	^
pkg/ansi/markdownterm.go:77:2: if statements should only be cuddled with assignments (wsl)
	if !text() {
	^
pkg/ansi/markdownterm.go:81:2: expressions should not be cuddled with blocks (wsl)
	out.WriteString("\n")
	^
pkg/ansi/markdownterm_test.go:17:2: assignments should only be cuddled with other assignments (wsl)
	r := MarkdownTermRenderer(flags)
	^
```

brandur-stripe:
ptal @ob-stripe

brandur-stripe:
Also wasn't immediately obvious what the lint error meant, but it seems what it wants you to do is add more newlines between things:

``` diff
diff --git a/pkg/ansi/ansi.go b/pkg/ansi/ansi.go
index 482309f..375d345 100644
--- a/pkg/ansi/ansi.go
+++ b/pkg/ansi/ansi.go
@@ -130,6 +130,7 @@ func StartSpinner(msg string, w io.Writer) *spinner.Spinner {
        }

        s.Start()
+
        return s
 }
```

pkg/config/config.go
```diff
@@ -75,6 +75,12 @@ func (c *Config) InitConfig() {
 		c.ProfilesFile = configFile
 		viper.SetConfigType("toml")
 		viper.SetConfigFile(configFile)
+		viper.SetConfigPermissions(os.FileMode(0600))
+
+		// Try to change permissions manually, because we used to create files
+		// with default permissions (0644). Ignore errors (e.g. if the file
+		// doesn't exist yet).
+		_ = os.Chmod(configFile, os.FileMode(0600))
```
joek-stripe: Is this consistent on non-unixes (aka Windows)?

joek-stripe: Also instead of swallowing the error, can we swallow the "file-non-existant" subset of this https://golang.org/src/os/error.go?s=1314:1377#L30 just in-case there is some other permissions related error that the user may need help with

ob-stripe: For Windows:
> On Windows, only the 0200 bit (owner writable) of mode is used; it controls whether the file's read-only attribute is set or cleared. The other bits are currently unused. For compatibility with Go 1.12 and earlier, use a non-zero mode. Use mode 0400 for a read-only file and 0600 for a readable+writable file.

ob-stripe:
ptal @joek-stripe

PR #227, opened 2019-10-09
Newlines, newlines everywhere

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
The latest version of golangci-lint includes a new linter (https://github.com/bombsimon/wsl) that forces you to add empty lines. Because we enable all linters by default, this causes new CI builds to fail.

I'm a little ambivalent about it, but I think I like it overall. If you'd rather, we can just disable the linter though.

### DIFF
--- a/pkg/requests/examples.go
+++ b/pkg/requests/examples.go
@@ -19,10 +19,12 @@ const (
 
 func parseResponse(response []byte) (map[string]interface{}, error) {
 	var result map[string]interface{}
+
 	err := json.Unmarshal(response, &result)
 	if err != nil {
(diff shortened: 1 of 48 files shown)


pkg/requests/examples.go
```diff
@@ -479,10 +505,10 @@ func (ex *Examples) InvoicePaymentFailed() error {
 		return err
 	}
 
-	fmt.Println("hello")
```
ob-stripe: bye

ob-stripe:
Thanks Brandur!

PR #236, opened 2019-10-14
Telemetry: command path

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Add very basic telemetry: send the command path (e.g. `charges create`, `listen`, etc.) in API requests, in a new `Stripe-CLI-Telemetry` header.

The header contains a JSON string, so we can easily add more information in there.

Users can opt out of telemetry by setting `STRIPE_CLI_TELEMETRY_OPTOUT` to `1` or `true`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,6 +366,10 @@ All of the available shortcuts are listed in the [wiki page](https://github.com/
 
 For dashboard pages, you can also add the `--livemode` flag to open the page directly in live mode.
 
+## Telemetry
+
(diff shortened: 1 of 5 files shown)


ob-stripe:
@tomer-stripe Can you take another look? I've changed the way it works to only send the command name, but no arguments or flags.

ptal @tomer-stripe

tomer-stripe:
lgtm!

PR #237, opened 2019-10-16
Add events resend command

ob-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform @rattrayalex-stripe 

 ### Summary
- Removes the `--event` flag in the `trigger` command (`trigger` is now exclusively used to trigger new test events, not replay existing ones)
- Add a new `events resend` subcommand (`events` already has `retrieve` and `list` subcommands which are regular API operation commands). The command expects one argument (the event ID) and accepts two optional flags: `--webhook-endpoint` to replay the event on a specific endpoint and `--account` to replay an event from a connected account. If `--webhook-endpoint` is omitted, then the CLI will send `for_stripecli=true` in the request to instruct the API to replay the event to the CLI endpoints.

Because the new command inherits all the standard API operation command infrastructure, it supports the `--livemode` flag.

### DIFF
--- a/pkg/cmd/resource/events.go
+++ b/pkg/cmd/resource/events.go
@@ -1,0 +1,31 @@
+package resource
+
+import (
+	"errors"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/config"
+)
+
+// AddEventsSubCmds adds custom subcommands to the `events` command created
+// automatically as a resource command.
+func AddEventsSubCmds(rootCmd *cobra.Command, cfg *config.Config) error {
+	found := false
+
+	for _, cmd := range rootCmd.Commands() {
+		if cmd.Use == "events" {
+			found = true
+
+			NewEventsResendCmd(cmd, cfg)
+
+			break
+		}
+	}
+
+	if !found {
+		return errors.New("Could not find events command")
+	}
+
+	return nil
+}
--- a/pkg/cmd/resource/events_resend.go
+++ b/pkg/cmd/resource/events_resend.go
@@ -1,0 +1,40 @@
+package resource
+
+import (
+	"net/http"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/config"
+)
+
+// EventsResendCmd represents the event resend API operation command. This
+// command is manually defined because it has a custom behavior.
+type EventsResendCmd struct {
+	opCmd *OperationCmd
+}
+
+func (erc *EventsResendCmd) runEventsResendCmd(cmd *cobra.Command, args []string) error {
+	// If the `webhook-endpoint` flag was not passed, then add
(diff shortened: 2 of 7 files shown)


rattrayalex-stripe:
nit: I don't think this should pretend to be an api endpoint (which it's not, at least to users) but instead be a new top-level command, `stripe resend evt_123`. Otherwise this paves the way for there to be other "resource commands" which aren't resource commands, and could leave users guessing as to whether a given resource command is a real api endpoint

ob-stripe:
> nit: I don't think this should pretend to be an api endpoint (which it's not, at least to users) but instead be a new top-level command, stripe resend evt_123. Otherwise this paves the way for there to be other "resource commands" which aren't resource commands, and could leave users guessing as to whether a given resource command is a real api endpoint

Fair point. I agree that this could be a potential source of confusion. @tomer-stripe Do you also think this should simply be `stripe resend`? It would make the code a bit cleaner too (no need to mess with the template for codegen'd commands).

tomer-stripe:
We will need to differentiate API commands from non-API commands for behaviors like this. We're planning to add commands to `terminal` that will not be API commands because we won't have a better namespace for it so this won't be the only case. It doesn't seem too far-fetched that this would be a common pattern long term, for example we could also add non-API features for things like Sigma or Radar.

I think I prefer `stripe events resend` or `stripe webhook resend` because it's more explicit about what's happening (plus then we can also move `trigger` into that as well and localize all non-listen webhooks functionality in on place).

rattrayalex-stripe:
Interesting, so it'd look like this?
```
Webhook commands:
  listen                        Listen for webhook events
  webhook trigger                       Trigger test webhook events to fire
  webhook resend                       Resend a previous webhook to any CLI listeners
```

ob-stripe:
We could also move `listen` into `webhook listen`, but keep the top-level `listen` command as an alias since it's an important feature.

tomer-stripe:
Yea, I was thinking it would look something like that. Agree with @ob-stripe's idea around having both for listen (I wonder if we should do a `stripe tail` alias with `logs tail` too but that's tangential)

ob-stripe:
Alright, after discussing it internally we're going with `events resend`.

@tomer-stripe Made a small change to make the code a little bit cleaner, so we don't have to mess with the generated template to add the new subcommand.

ptal @tomer-stripe

rattrayalex-stripe:
Curious why that one? How will we document it?

pkg/cmd/resource/events.go
```diff
@@ -0,0 +1,31 @@
...
+		if cmd.Use == "events" {
+			found = true
+
+			NewEventsResendCmd(cmd, cfg)
+
+			break
+		}
+	}
+
+	if !found {
```
tomer-stripe: Is this mostly for people who develop on the cli? Seems like it won't happen during runtime.

ob-stripe: Yep, exactly.

PR #243, opened 2019-10-17
Better samples configuration

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 

 ### Summary
This cleans up the sample command to use a `.cli.json` configuration file that'll be added to all samples. The config file looks like:

```
{
  "name": "web-elements-card-payment",
  "configureDotEnv": true,
  "postInstall": {
    "message": "Hello world!"
  },
  "integrations": [
    {
      "name": "using-webhooks",
      "clients": ["html"],
      "servers": ["java", "node", "php-slim", "php", "python", "ruby"]
    },
    {
      "name": "without-webhooks",
      "clients": ["html"],
      "servers": ["java", "node", "php-slim", "php", "python", "ruby"]
    }
  ]
}
```


pkg/samples/samples.go
```diff
@@ -401,7 +277,7 @@ func (s *Samples) ConfigureDotEnv(sampleLocation string) error {
 		return err
 	}
 
-	dotenv["STRIPE_PUBLIC_KEY"] = publishableKey
+	dotenv["STRIPE_PUBLISHABLE_KEY"] = publishableKey
```
adreyfus-stripe: Yaaaaaay

pkg/samples/samples.go
```diff
@@ -19,55 +18,56 @@ import (
...
+	}
 
-func javaPath(path string) string {
-	return fmt.Sprintf(`String ENV_PATH = "%s/";`, path)
+	return []string{}
 }
 
-func jsPath(path string) string {
-	return fmt.Sprintf(`const envPath = resolve(__dirname, "%s/.env");`, path)
+type integration struct {
```
brandur-stripe: Would you mind if we renamed this to `sampleConfigIntegration` so that it's really easy to see that the two structs are related? (And both used for configuration.)

pkg/samples/samples.go
```diff
@@ -19,55 +18,56 @@ import (
...
+	Name    string   `json:"name"`
+	Clients []string `json:"clients"`
+	Servers []string `json:"servers"`
```
brandur-stripe: Would you mind putting a basic doc on `Clients` and `Servers` describing what each is?

pkg/samples/samples.go
tomer-stripe: Done!

PR #244, opened 2019-10-18
Refactor trigger commands to use fixtures

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
Removes a lot of the code for triggers and uses the fixtures logic under the hood.

Questions for you two:
1. what do you think we should do with the existing test suite?
2. what should we do with the webhooks and retry requests that live in that file?


ob-stripe:
❤️ So much cleaner.

> what do you think we should do with the existing test suite?

Provided the code for the fixtures logic is itself well tested, I don't think we need test the fixtures files.

> what should we do with the webhooks and retry requests that live in that file?

I've already removed the retry code in #237. The only remaining function is for listing webhook endpoints, and we can probably move it elsewhere too (probably in the `stripe` package).

tomer-stripe:
@ob-stripe cool! The underlying logic for fixtures has some tests so I'll assume that's fine for now. I'll remove the rest of the examples code then :D

Freezystem:
Hi @tomer-stripe, I don't know if it's feasible but it would be great if we also be able to trigger an event for a specific resource ID, like so:

`stripe trigger payment_intent.succeeded --id pi_xxxxxxxxxxx`

In this case, as the PaymentIntent already exists before triggering the event, `payment_intent.created` should not be fired. 

It would be so much easier to integrate triggers in CI e2e tests as we fire the exact event we want on an already existing resource that needs to have a specific configuration (amount, currency, metadata, etc...).

Let me know if I'm not clear or if I can help on this one.

tomer-stripe:
@Freezystem I like the idea but I'm not sure if I'll be able to get to that as part of this review. Would you be able to use the normal `payment_intents` command for that? It should be `stripe payment_intents confirm <id>`

Freezystem:
> I'm not sure if I'll be able to get to that as part of this review
Yeah sure this must be part of another PR

> Would you be able to use the normal payment_intents command for that?
Yes it's great to confirm a PaymentIntent but I'd also like to make a PaymentIntent fail. 
Maybe it misses the `stripe payment_intent payment_failed <id>` command.

tomer-stripe:
ptal @ob-stripe @brandur-stripe

tomer-stripe:
@ob-stripe I'm not entirely sure what's going on but running `make build` with `//build +dev` causes the `fs.go` file to not get picked up (it complains that the var `FS` doesn't exist). If I remove that it works fine but then there's the weird double declaration issue.

So... I'm not entirely sure what's going on with gen right now. It worked in generating the commands again but it's a really hacky solution to remove it everytime.

PR #256, opened 2019-10-29
Manually check for env variables and disable the automatic ones

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
We're thinking that viper's automatic env wasn't playing well with how we setup user profiles. Since the two most likely cases of env variables are `device_name` and `api_key` this adds a quick check for those and uses that if found.

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/config/profile.go
```diff
@@ -68,6 +73,10 @@ func (p *Profile) GetDeviceName() (string, error) {
 
 // GetAPIKey will return the existing key for the given profile
 func (p *Profile) GetAPIKey(livemode bool) (string, error) {
+	if os.Getenv("STRIPE_API_KEY") != "" {
+		return os.Getenv("STRIPE_API_KEY"), nil
```
ob-stripe: We might want to validate the API key like we do for `p.APIKey` just below.

tomer-stripe: Good call, will fix real quick!

PR #257, opened 2019-10-30
Do not generate commands for deprecated API methods

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Do not generate commands for deprecated API methods.

I'm not thrilled with the implementation (the template file has too much logic) but this works for now. At some point I want to change the entire generation process to simply transform the OpenAPI spec file into a giant Go structure and simply use the struct at runtime for our needs, instead of directly generating the final Go code. This would be slightly less efficient but the vast majority of the time spent processing the OpenAPI is just parsing the JSON file into memory anyway.


ob-stripe:
Released as v0.8.2.

PR #271, opened 2019-11-04
Add some end-to-end smoke testing

tomer-stripe:
### Reviewers
r? @ob-stripe @brandur-stripe 
cc @stripe/dev-platform

 ### Summary
I wanted to build a framework where we could gradually add some end-to-end smoke testing for the CLI. I'm starting with `trigger` since that's one of the more tedious ones to test manually.

You would run this with `go run -tags dev test.go`

What do you two think of this direction?

### DIFF
--- a/test.go
+++ b/test.go
@@ -1,0 +1,82 @@
+// build +dev
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/stripe/stripe-cli/pkg/fixtures"
+)
+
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Seems like it'd be worth a try! I'd suggest in addition:

1. Move it to a subfolder and give the file a slightly more descriptive name.
2. Make sure it runs in CI.
3. Make sure it gives us a good error message back in case something fails.

ob-stripe:
I like the idea of an end-to-end test! If you can address Brandur's feedback, I think this would be a great addition.

PR #285, opened 2019-11-05
Add example to run Docker container

phylor:
### Reviewers
r? @
cc @stripe/dev-platform

 ### Summary
Docker containers cannot use `stripe-cli login` unless you use volumes to store the login data. Instead use the `--api-key` option to authenticate. I guess an additional example would help out to clarify that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ docker run --rm -it stripe/stripe-cli version
 stripe version x.y.z (beta)
 ```
 
+Please refer to the [wiki](../../wiki/Using-with-Docker) for detailed instructions on how to use the Docker container.
+
 ### Without package managers
 
 Instructions are also available for installing and using the CLI [without a package manager](https://github.com/stripe/stripe-cli/wiki/Installing-and-updating#without-a-package-manager).


ob-stripe:
Hi @phylor, thanks for the contribution!

Two remarks:

1. The name of the executable is simply `stripe`, not `stripe-cli`

2. I don't feel super strongly about it, but I think it'd preferable to pass the API key via an environment variable rather than the `--api-key` flag:
```sh
$ docker run --env STRIPE_API_KEY=sk_test_... --rm -it stripe/stripe-cli listen --forward-to=localhost:5000/hooks

**Comment by auchenberg** [comment]
We are covering Docker support in https://github.com/stripe/stripe-cli/wiki/installation#docker. We should update the instructions there in order to keep the README lean. I think we should mention Docker, so people know where to read more.

**Comment by phylor** [comment]
@ob-stripe @auchenberg Thanks for the quick response! Your suggestions make sense! If you add it to the wiki, would it make sense to refer to the wiki from the Docker section in the README? I wouldn't have checked there because it's called "installation".

**Comment by PhilETaylor** [comment]
Should also add to docker docs to mount .config/stripe folder to share the secrets with the container... without which you need to provide keys on every command . :(

**Comment by auchenberg** [comment]
@phylor @ob-stripe 

How about that we update the `README` under the Docker section to link to the wiki for Docker usage? I created https://github.com/stripe/stripe-cli/wiki/Using-with-Docker

Something like `Read more here on how to use the Stripe CLI with Docker`?

**Comment by PhilETaylor** [comment]
I had success with sharing the .config/stripe folder as such:

`docker run -v /Users/phil/.config:/root/.config --rm -it stripe/stripe-cli version`

which I then `alias stripe="docker run -v /Users/phil/.config:/root/.config --rm -it stripe/stripe-cli version"` in my `.zshrc` and/or `.bash_profile` files

that way you dont ever "store" credentials in a volume or container

**Comment by phylor** [comment]
@auchenberg Nice, thank you! I changed this PR to add the wiki reference to the README (copied the wording you already had about the reference to the linux wiki).

Maybe you could also add @PhilETaylor comments to the wiki? That would certainly help as well!

**Comment by Freezystem** [comment]
Maybe you can add something about the use of the `stripe-cli` within a docker-compose.
To get the stripe-cli to be able to broadcast hooks to other containers inside the same docker-compose you have to use this specific address:
```sh
listen --api-key rk_test_***** --device-name "local-$USER" --forward-to host.docker.internal:8080/hooks
```
This example also show how to register the hook as the user name in the Stripe Dashboard. This allow several machine to be able to run the listen command without being kicked.

PhilETaylor:
Maybe even check out https://github.com/stripe/stripe-cli/issues/304 and add notes around that topic (topic: some commands are useless in a container) and notes about opening ports if needed :)

ob-stripe:
Thanks!

PR #314, opened 2019-11-15
Don't use raw PANs in triggers

ob-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
Don't use raw PANs in triggers. Instead use the synthetic payment method tokens.


triggers/payment_method.attached.json
```diff
@@ -9,23 +9,12 @@
       "method": "post"
     },
     {
-      "name": "payment_method",
-      "path": "/v1/payment_methods",
+      "name": "payment_method_attach",
+      "path": "/v1/payment_methods/pm_card_visa/attach",
```
tomer-stripe: Ohh cool, didn't know you could do that!

PR #358, opened 2020-01-28
serve cmd

tomer-stripe:
### Reviewers
r? @ob-stripe @adreyfus-stripe 
cc @stripe/dev-platform @suz-stripe @brandur-stripe 

 ### Summary
This adds a simple command to serve static files from the CLI. We'd like this for samples since a number of them have an `index.html` file and we've been suggested folks use the Python simple server, which may require installing the entire Python ecosystem.

The command is `stripe serve` (or `stripe srv`) which takes an optional argument that is a path to a directory. If not path is given, it serve the local folder: `'.'`.

The command has some basic logging available to it when requests are made:
```
➜ stripe-dev serve
Starting stripe server at address http://localhost:4242
::1 - - [28/Jan/2020:10:33:36 -0800] "GET / HTTP/1.1" 304 0
::1 - - [28/Jan/2020:10:33:41 -0800] "GET /.editorconfig HTTP/1.1" 200 233
^Csignal: interrupt
```

If an `index.html` file exists in the folder, the CLI will automatically render that. If not, it'll display a list of files in that specific directory.

The command itself will continue running (and block) until the user kills it with ctrl+c.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/google/go-cmp v0.3.1 // indirect
 	github.com/google/go-github/v28 v28.1.1
 	github.com/gordonklaus/ineffassign v0.0.0-20190601041439-ed7b1b5ee0f8 // indirect
+	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/websocket v1.4.1
 	github.com/iancoleman/strcase v0.0.0-20190422225806-e506e3ef7365
 	github.com/joho/godotenv v1.3.0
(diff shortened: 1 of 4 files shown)


suz-stripe:
I think this is a nice convenience and helps keep our documentation more concise with fewer non-Stripe related rabbit holes :100:

tomer-stripe:
@ob-stripe that's a good idea. I wonder if we could leverage the cli setup files we made for samples to give users additional instructions 🤔

@adreyfus-stripe what do you think?

adreyfus-stripe:
I like this as a first pass! Ultimately we could extend to add more serve config via .cli.json

PR #366, opened 2020-02-07
Remove a defer statement inside of a loop

gsquire:
### Reviewers
cc @stripe/dev-platform

 ### Summary
I caught this with my [linter](https://github.com/gsquire/dll) while reviewing the code for learning purposes. I know retries may not happen frequently but if they did, they were leaking the previous connection's response body.


tomer-stripe:
Sorry for the delay, thanks for catching this!

PR #371, opened 2020-02-11
add terminal quickstart feature (beta)

suz-stripe:
### Reviewers
r? @tomer-stripe @ob-stripe 
cc @stripe/dev-platform

 ### Summary

I built a “quickstart” experience for Stripe Terminal into the Stripe CLI, to allow developers to run a test payment on their new Verifone P400 Terminal reader in < 5 minutes.

The command to run (in test mode only) is `stripe terminal quickstart --api-key sk_test_123`

To see a GIF of what the flow generally looks like when using the command, expand the section below:

<details>
<summary>Expand for GIF</summary>
terminal animation showing the quickstart interactive prompt
</details>

**Feedback**

All feedback is very welcome!

Would love nitpicky feedback about idiomatic Golang stuff because I'm so new to the language, especially around code duplication and type choices. I made some escape hatches in a few utility functions with interface types but would love guidance around this if you think it's going to be a concern.

Also - I'd love for volunteers to pull this down and build + play with it if you have a Verifone P400 reader 🙏

⚠️ this PR depends on some allow-listing work, so please don't merge just yet until it's all :ok:

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -131,7 +131,7 @@ func (p *Profile) GetPublishableKey() string {
 	return ""
 }
 
-// GetTerminalPosDeviceID returns the device id from the config for Terminal quickstart to use
+// GetTerminalPOSDeviceID returns the device id from the config for Terminal quickstart to use
 func (p *Profile) GetTerminalPOSDeviceID() string {
 	if err := viper.ReadInConfig(); err == nil {
 		return viper.GetString(p.GetConfigField("terminal_pos_device_id"))
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -21,6 +21,7 @@ type Profile struct {
 	LiveModePublishableKey string
 	TestModeAPIKey         string
 	TestModePublishableKey string
+	TerminalPosDeviceID    string
 }
 
 // CreateProfile creates a profile when logging in
@@ -125,6 +126,15 @@ func (p *Profile) GetPublishableKey() string {
 		}
 
 		return viper.GetString(p.GetConfigField("test_mode_publishable_key"))
+	}
+
+	return ""
+}
+
+// GetTerminalPosDeviceID returns the device id from the config for Terminal quickstart to use
+func (p *Profile) GetTerminalPosDeviceID() string {
+	if err := viper.ReadInConfig(); err == nil {
+		return viper.GetString(p.GetConfigField("terminal_pos_device_id"))
 	}
 
 	return ""
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -21,7 +21,7 @@ type Profile struct {
 	LiveModePublishableKey string
 	TestModeAPIKey         string
 	TestModePublishableKey string
-	TerminalPosDeviceID    string
+	TerminalPOSDeviceID    string
 }
 
 // CreateProfile creates a profile when logging in
@@ -132,7 +132,7 @@ func (p *Profile) GetPublishableKey() string {
 }
 
 // GetTerminalPosDeviceID returns the device id from the config for Terminal quickstart to use
-func (p *Profile) GetTerminalPosDeviceID() string {
+func (p *Profile) GetTerminalPOSDeviceID() string {
 	if err := viper.ReadInConfig(); err == nil {
 		return viper.GetString(p.GetConfigField("terminal_pos_device_id"))
 	}
--- a/pkg/terminal/p400/rabbit_service.go
+++ b/pkg/terminal/p400/rabbit_service.go
@@ -54,18 +54,23 @@ func encodeRPCContent(rpcContent interface{}) (string, error) {
 func CallRabbitService(tsCtx TerminalSessionContext, method string, methodContent interface{}, methodResponse interface{}, parentTraceID string) error {
 	encodedMethodContent, err := encodeRPCContent(methodContent)
 
-	var result RabbitServiceResponse
-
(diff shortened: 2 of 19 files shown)


pkg/config/profile.go
```diff
@@ -21,6 +21,7 @@ type Profile struct {
 	LiveModePublishableKey string
 	TestModeAPIKey         string
 	TestModePublishableKey string
+	TerminalPosDeviceID    string
```
tomer-stripe: nit: should we do `TerminalPOSDeviceID` (capitalize POS)?

pkg/terminal/p400/rabbit_service.go
```diff
@@ -0,0 +1,114 @@
...
+	formattedIP := strings.Join(strings.Split(tsCtx.IPAddress, "."), "-")
+
+	rabbitServiceURL := fmt.Sprintf(readerURL, formattedIP)
+	res, err := http.Post(rabbitServiceURL, "application/json", &payload)
+
+	if err != nil {
+		return err
+	}
+
+	json.NewDecoder(res.Body).Decode(&result)
```
tomer-stripe: possibly silly question, is there a difference between this and `json.Unmarshal(res.Body, &result)`?

pkg/terminal/p400/reader_errors.go
```diff
@@ -0,0 +1,38 @@
+package p400
+
+import (
+	"errors"
+)
+
+var (
```
tomer-stripe: Nice!

pkg/terminal/p400/reader_methods.go
```diff
@@ -0,0 +1,391 @@
...
+	PaymentIntentID    string
+	Amount             int
+	Currency           string
+}
+
+// DeviceInfo belongs to the Rabbit Service RPC call payload shape
+type DeviceInfo struct {
+	DeviceClass   string        `json:"device_class"`
+	DeviceUUID    string        `json:"device_uuid"`
+	HostOsVersion string        `json:"host_os_version"`
```
tomer-stripe: `HostOSVersion`? I don't have too strong feelings here though

pkg/terminal/quickstart_p400.go
```diff
@@ -0,0 +1,132 @@
...
+			return fmt.Errorf(err.Error())
+
+		case p400.ErrActivateReaderFailed.Error():
+			return fmt.Errorf(err.Error())
+
+		case p400.ErrConnectionTokenFailed.Error():
+			return fmt.Errorf(err.Error())
+
+		case p400.ErrNewRPCSessionFailed.Error():
+			return fmt.Errorf(err.Error())
```
tomer-stripe: Not a huuuuge change but you could use `fallthrough`: https://github.com/golang/go/wiki/Switch#fall-through

pkg/config/profile.go
suz-stripe: yes it totally should be 👍

pkg/terminal/p400/reader_methods.go
suz-stripe: ya I agree

pkg/terminal/quickstart_p400.go
suz-stripe: neat, will look at this thanks!

pkg/terminal/p400/rabbit_service.go
suz-stripe: Great question. The response body is a stream rather than a string / string byte array so it was more convenient to pass it to a decoder instead. Is it more idiomatic to use `Unmarshal` after converting the stream?

suz-stripe: I read more into this and it seems that it's preferable just to pass it to a `Decoder`

pkg/terminal/p400/rabbit_service.go
```diff
@@ -0,0 +1,136 @@
...
+		if len(t.DNSIPs) == 0 {
+			return ErrDNSFailed
+		}
+
+		return t.Err
+	}
+
+	json.NewDecoder(res.Body).Decode(&result)
+
+	res.Body.Close()
```
tomer-stripe, line 94: Go recommends putting a `defer` in front of `Close`s to make sure things close: https://gobyexample.com/defer

pkg/terminal/p400/reader_methods.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+// ReaderQueryPaymentContent represents the shape of the serialized protobuf sent to Rabbit Service for querying a payment for its collection state
+type ReaderQueryPaymentContent struct {
+	TransactionContext TransactionContext `json:"transaction_context"`
+}
+
+// ReaderQueryPaymentResponse is the decoded RPC response from calling the queryPaymentMethod reader method
+type ReaderQueryPaymentResponse struct {
+	PaymentMethod interface{} `json:"payment_method"`
```
tomer-stripe, line 126: For these `PaymentMethod`s, is there data you're pulling out of later? Wondering why they're `interface`s instead of [partially-defined] structs.

suz-stripe: we're not pulling any information out as the payment method will always be empty for a test payment. I wasn't able to get Go to encode/decode it to an actual empty object instead of `nil` unless I did it this way. A subsequent call to Stripe needs that to be an empty object in the request or it'll reject the payload.

suz-stripe:
ptal @tomer-stripe

PR #374, opened 2020-02-13
Add "oxxo-payment" Stripe Sample to CLI 

fay-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary

### DIFF
--- a/pkg/samples/list.go
+++ b/pkg/samples/list.go
@@ -156,7 +156,7 @@ var List = map[string]*SampleData{
 	"multiple-plan-subscriptions":      multiPlanSubscriptions,
 	"set-up-subscriptions":             singleSubscription,
 	"web-elements-sepa-debit-payment":  sepaPayment,
-	"oxxo-payment": 					oxxoPayment,
+	"oxxo-payment":                     oxxoPayment,
 	"web-elements-ideal-payment":       idealPayment,
 	"web-elements-fpx-payment":         fpxPayment,
 	"charging-a-saved-card":            chargingSavedCard,
--- a/pkg/samples/list.go
+++ b/pkg/samples/list.go
@@ -102,6 +102,12 @@ var singleSubscription = &SampleData{
 	URL:         "https://github.com/stripe-samples/set-up-subscriptions",
 }
 
+var oxxoPayment = &SampleData{
+	Name:        "oxxo-payment",
+	Description: "Build a payment form to collect OXXO and card payments",
+	URL:         "https://github.com/stripe-samples/oxxo-payment",
+}
+
 var sepaPayment = &SampleData{
 	Name:        "web-elements-sepa-debit-payment",
 	Description: "Accept SEPA Debit payments",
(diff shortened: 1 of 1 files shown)


tomer-stripe:
@fay-stripe overall looks good but you need a `.cli.json` file in the sample repo so that the CLI can parse it: https://github.com/stripe-samples/accept-a-card-payment/blob/master/.cli.json

fay-stripe:
@tomer-stripe Ok will add to oxxo-payment repo! Just OOC, I'm not seeing that file in https://github.com/stripe-samples/web-elements-ideal-payment is there a reason why that one doesn't need a .cli.json?

Also, I can just define it as having one "main" integration, similar to this: https://github.com/stripe-samples/round-up-and-donate/blob/master/.cli.json since I don't think we need "using-webhooks" and without, right?

fay-stripe:
Also, I've been trying to figure out how to gofmt that list.go file (pardon, it's my first time working with go) and I've ran `gofmt -s` in the stripe-cli/ directory but it looks like it's just hanging, and then I've ran `gofmt -s pkg/samples/list.go` but it doesn't look like it's actually saving the file, just printing to stdout.... and when I run `make lint` it still shows me the lint error

pkg/samples/list.go
```diff
@@ -156,7 +156,7 @@ var List = map[string]*SampleData{
 	"multiple-plan-subscriptions":      multiPlanSubscriptions,
 	"set-up-subscriptions":             singleSubscription,
 	"web-elements-sepa-debit-payment":  sepaPayment,
-	"oxxo-payment": 					oxxoPayment,
+	"oxxo-payment":                     oxxoPayment,
```
fay-stripe: I honestly don't see what the diff between these two are, but I guess it now passes `make lint` 🤷‍♀ . I ended up doing gofmt and just copy/pasting the output into VSCode and saving it that way

fay-stripe:
Added the .cli.json file here: https://github.com/stripe-samples/oxxo-payment/pull/3

pkg/samples/list.go
tomer-stripe: go likes it when everything is vertically lined up

tomer-stripe:
@fay-stripe web-elements-ideal-payments won't work with the CLI (if it's in there we should bug... someone).

`make lint` is the best way to run the linter, looks like it worked (I think?). 

And yea, if you have one `main` integration then the CLI won't prompt the user to make any integration-related choices.

fay-stripe:
> web-elements-ideal-payments won't work with the CLI (if it's in there we should bug... someone).

Oh it's just, I noticed here: https://github.com/stripe/stripe-cli/pull/374/files#diff-0d3b591470484cb5b58463d5ddd0c127R160 and the instructions in the README: https://github.com/stripe-samples/web-elements-ideal-payment

adreyfus-stripe:
Ah yeah I think we forgot to include a .cli.json. I can add to the ideal sample

adreyfus-stripe:
Added! https://github.com/stripe-samples/web-elements-ideal-payment/commit/7a8381ee7edbcca278bf18a5a183d1f5809de108

Issue #379, opened 2020-02-17
"the file contains a virus" on Windows 10


lukepuplett:
### Issue
Hello - I downloaded and unzipped `stripe_1.3.4_windows_x86_64.zip` and ran the `stripe.exe` binary inside but received:

> Operation did not complete successfully because the file contains a virus or potentially unwanted software.

![image](https://user-images.githubusercontent.com/5802524/74670648-861f3c80-51a1-11ea-95c5-fad8430625c7.png)

### Expected Behavior
I think I expected an installer to run.

### Steps to reproduce
On Windows 10 v2004 (19041.84), download and unzipp `stripe_1.3.4_windows_x86_64.zip` and run the `stripe.exe` binary, accept any warnings about running untrusted software.

### Traceback
None.

### Environment
Windows 10 v2004 (19041.84)

lukepuplett:
Just as a side note, I also tried running the Stripe CLI from a WSL2 console on my machine but although it works, it won't talk to the web server running on the Windows side (testing webhooks).

So if anyone knows how to solve that, its an alternative for us --poor sods-- developing on Windows.

Julienh:
Hello
Same problem, here is the Trojan detected : Trojan:Win32/Emali.A!cl

lukepuplett:
Just to come back on using Stripe CLI under WSL2 talking to a .NET Core website on the Windows side, you'll need to:

 - Ensure Kestrel is listening on the physical IP
 - Setup a hosts entry for an alias for your physical IP
 - Setup a self-signed certificate for that hostname
 - Add a Windows Firewall rule
 - Then use `stripe --skip-verify ...`

ob-stripe:
Hi, thanks for the report! It does look like recent versions of the CLI are reported as containing the "Trojan:Win32/Emali.A!cl" virus by Microsoft Defender. This is a false positive, and is safe to ignore. We will reach out to Microsoft to clear up the issue.

ob-stripe:
Closing this issue, but please feel free to reply if you're still getting false positives with the most recent versions of the CLI and we'll reopen.

R4wizard:
I've received this false positive today from Windows Defender.

VladMstv:
Same here. After I've executed "scoop install stripe", windows defender says that it's detected Trojan:Win32/Occamy.C41

skoggimord:
[...]

PR #382, opened 2020-02-19
Upload Windows archives to VirusTotal

ob-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
Automatically upload new releases to VirusTotal. I'm hoping this will let us catch issues like #379 earlier.

### DIFF
--- a/scripts/publish-deb-to-bintray.sh
+++ b/scripts/publish-deb-to-bintray.sh
@@ -9,6 +9,11 @@ PACKAGE="stripe"
 DISTRIBUTIONS="stable"
 COMPONENTS="main"
 
+if [ -z "$GITHUB_TOKEN" ]; then
+  echo "GITHUB_TOKEN is not set"
+  exit 1
+fi
+
 if [ -z "$BINTRAY_USER" ]; then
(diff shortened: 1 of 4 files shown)


scripts/publish-deb-to-bintray.sh
```diff
@@ -20,7 +25,7 @@ if [ -z "$BINTRAY_API_KEY" ]; then
 fi
 
 setVersion () {
-  VERSION=$(curl --silent "https://api.github.com/repos/stripe/stripe-cli/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/');
+  VERSION=$(curl --silent "https://api.github.com/repos/stripe/stripe-cli/releases/latest" -u $GITHUB_TOKEN: | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/');
```
ob-stripe: I've updated the .deb and .rpm release scripts to authenticate requests to GitHub. Since we use unauthenticated requests to check for new versions from the CLI itself, we can easily run into rate limits, which would cause the scripts to fail.

Issue #295, opened 2019-11-06
`trigger` with customizable data


ob-stripe:
### Description of the desired feature

The `trigger` command can be used to create sample events of various types. Many users would like to be customize the events.

For example, one could trigger a `charge.captured` event for an existing charge object:
```sh
$ stripe trigger charge.captured --charge ch_123
```

Or one could trigger a `customer.subscription.created` for a specific customer, with custom metadata going into the subscription object:
```sh
$ stripe trigger customer.subscription.created --customer cus_123 -d metadata[key]=value
```

At the moment, none of the above is possible.

### Current implementation of the `trigger` command

Under the hood, the `trigger` command works by sending a series of API requests. E.g. to trigger a `charge.captured` event, the CLI will send two requests to Stripe:
1. A charge creation request with `captured=false`, to create a new uncaptured charge object
2. A charge capture request for the newly created object

These series of API requests are defined using as JSON files ("fixtures" files) that are embedded in the executable, e.g. https://github.com/stripe/stripe-cli/blob/master/triggers/charge.captured.json.

### Difficulties in providing a generic interface to customize data

Due to the large number of possible event types and scenarios, it's not feasible to customize the behavior of the CLI per event type -- that would be near impossible to maintain in the long term. We want to keep the interface of the `trigger` command generic. However, because each "trigger" is defined as an arbitrary sequence of API requests, it is difficult to provide a generic CLI interface to customize the data that goes into each request.

For instance, for the case of capturing an existing charge object, the `trigger` command would have to omit the first request (the charge creation request) when the `--charge` flag is provided.

For the case of providing custom metadata when creating a subscription, it's even more complicated: under the hood, the `trigger` command needs to create a new customer and a new plan before it can create the new subscription. If custom metadata is provided on the command line, it's not clear which resource (the plan, the customer or the subscription) should receive the custom metadata.

### Creating your own fixtures files

It's not a perfect solution, but you can create your own fixtures files (i.e. sequences of API requests) with your own custom data and have the CLI play them with the `stripe fixtures` command.

E.g. in order to capture an existing charge object, you could create a `my_fixture.json` file with the following contents:

```json
{
  "_meta": {
    "template_version": 0
  },
  "fixtures": [
    },
    {
      "name": "capture",
      "path": "/v1/charges/ch_123/capture",
      "method": "post"
    }
  ]
}
```

then play it with `stripe fixtures my_fixture.json`. This would send a single request to capture the `ch_123` charge.

[...]

PR #387, opened 2020-02-23
Support env vars in fixtures

thorsten-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
There is a use case where I might want to set up a customer object with a specific email address or phone number and I want to pull that data from the `.env` file:

```
{
  "_meta": {
    "template_version": 0
  },
  "fixtures": [
    {
      "name": "cust_bender",
      "path": "/v1/customers",
      "method": "post",
      "params": {
        "source": "tok_amex",
        "email": "bender@planex.com",
        "phone": "${.env:PHONE}",
        "address": {
          "city": "New New York"
        }
      }
    }
  ]
}
```

I'm not a go expert, and specifically am not sure what `os.Getwd()` 
https://github.com/stripe/stripe-cli/compare/thor/support-env-vars-in-fixtures?expand=1#diff-133d387253cca3c9f0d5be8b5c845987R280
will refer to once compiled. Will that be the directory I'm running `stripe` in?

### DIFF
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -284,6 +284,9 @@ func (fxt *Fixture) parseQuery(value string) string {
 				fmt.Println(fmt.Sprintf("Env var %s not set. Looking for .env file.", key))
 				// Try to load from .env file
 				dir, err := os.Getwd()
+				if err != nil {
+					dir = ""
+				}
 				err = godotenv.Load(path.Join(dir, ".env"))
 				if err != nil {
 					fmt.Println(fmt.Sprintf("Error: %v", err))
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -277,15 +277,20 @@ func (fxt *Fixture) parseQuery(value string) string {
 
 		// Catch and insert .env values
 		if name == ".env" {
-			dir, err := os.Getwd() // which dir is this once compiled?
-			err = godotenv.Load(path.Join(dir, ".env"))
-			if err != nil {
-				fmt.Println(fmt.Sprintf("Error loading .env: %v", err))
-				return value
-			}
-
 			key := nameAndQuery[2]
+			// Check if env variable is present
 			envValue := os.Getenv(key)
+			if envValue == "" {
+				fmt.Println(fmt.Sprintf("Env var %s not set. Looking for .env file.", key))
+				// Try to load from .env file
+				dir, err := os.Getwd()
+				err = godotenv.Load(path.Join(dir, ".env"))
+				if err != nil {
+					fmt.Println(fmt.Sprintf("Error: %v", err))
+					return value
+				}
+				envValue = os.Getenv(key)
+			}
 			if envValue == "" {
 				fmt.Println(fmt.Sprintf("No value for env var: %s", key))
 				return value
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -269,11 +269,17 @@ func (fxt *Fixture) parseArray(params []interface{}, parent string, index int) [
 
 func (fxt *Fixture) parseQuery(value string) string {
 	// Queries to fill data will start with #$ and contain a : -- search for both
-	// to make sure that we're trying to parse a query
-	r := regexp.MustCompile(`\${(.+):(.+)}`)
+	// to make sure that we're trying to parse a query.
+	// Additionally look for an optional default value: ${name:json_path|default_value}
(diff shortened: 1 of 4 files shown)


pkg/fixtures/fixtures.go
```diff
@@ -275,6 +275,24 @@ func (fxt *Fixture) parseQuery(value string) string {
 		nameAndQuery := r.FindStringSubmatch(value)
 		name := nameAndQuery[1]
 
+		// Catch and insert .env values
+		if name == ".env" {
+			dir, err := os.Getwd() // which dir is this once compiled?
```
tomer-stripe: It'll be the dir you execute from.

thorsten-stripe:
> Generally this looks fine. My one big question is what happens if you don't have an env var. It seems like we need some way to gracefully handle defaults or empty values.

Thanks @tomer-stripe I believe it does error gracefully, as in it outputs the error message but continues with the requests.
image
 
In the case of error it simply returns the string value:
image

thorsten-stripe:
@tomer-stripe any idea what's going wrong in `make ci`?

tomer-stripe:
Hmmm, I'm wondering if every trigger should have some default value so that the user doesn't have to think about it. I think what would be great is if the fixture could act as the default and we could somehow override individual fields in there (check out #295). @ob-stripe suggests using flags but I wonder if env-vars would be easier since we wouldn't need to dynamically support arbitrary flags.

As for the CI, it looks like you added `stripe-go` as a dependency in this PR which we don't use. Try removing that?

thorsten-stripe:
We could support the nullish coalescing operator:
```
"phone": "${.env:PHONE} ?? +1234567890",
```

Or do you think that gets too verbose?

tomer-stripe:
Let me think about this a bit more...

@ob-stripe do you have any ideas here?

ob-stripe:
Hm, I'd prefer not to make the string interpolation too complex, but we could support default values in case the left-hand side is not set (regardless of whether it's a "regular" variable or an environment variable).

Maybe something like this?

```
"phone": "${.env:PHONE|+1234567890}",
```

thorsten-stripe:
@ob-stripe works for me :) I've updated the regex to support default value after pipe. I've also changed from `Find` to `FindR`(https://github.com/thedevsaddam/gojsonq/wiki/Queries#findrpath) to gracefully error when the json_path is incorrect. This means we can now do:

```
"phone": "${.env:PHONE|+1234567890}",
"customer": "${test_customer:id|cus_1234}"
```

pkg/fixtures/fixtures.go
```diff
@@ -269,18 +269,52 @@ func (fxt *Fixture) parseArray(params []interface{}, parent string, index int) [
...
-	r := regexp.MustCompile(`\${(.+):(.+)}`)
+	// to make sure that we're trying to parse a query.
+	// Additionally look for an optional default value: ${name:json_path|default_value}
+	r := regexp.MustCompile(`\${([^\|]+):([^\|]+)\|?(.+)?}`)
 	if r.Match([]byte(value)) {
 		nameAndQuery := r.FindStringSubmatch(value)
 		name := nameAndQuery[1]
 
+		// Check if there is a default value specified
+		if nameAndQuery[3] != "" {
```
tomer-stripe: Do you need to check the length here? (want to make sure we don't get any array out of bounds errors)

pkg/fixtures/fixtures.go
```diff
@@ -269,18 +269,52 @@ func (fxt *Fixture) parseArray(params []interface{}, parent string, index int) [
...
+			value = nameAndQuery[3]
+		}
+
+		// Catch and insert .env values
+		if name == ".env" {
```
tomer-stripe: Are there any unit tests you could add for this? It'd be helpful to verify that the parsing works and we don't accidentally introduce regressions. Check out https://github.com/stripe/stripe-cli/blob/master/pkg/fixtures/fixtures_test.go

pkg/fixtures/fixtures.go
thorsten-stripe: @tomer-stripe AFAIK the regex always returns three groups (https://play.golang.org/p/pf_oViIZBah), so we should be good unless we're concerned that this behaviour might change in Go?

thorsten-stripe:
@tomer-stripe I've added test for all cases and they run fine except for `TestParseWithEnvFile` where I'm struggling to write the `.env` file as part of the test. Any idea how to achieve that?

tomer-stripe:
Looks good, ty!

PR #388, opened 2020-02-24
Add Developer Office Hours Starter Sample

cjavilla-stripe:
### Reviewers
r? @tomer-stripe 
cc @adreyfus-stripe 

 ### Summary
Adding link to stripe-sample for developer office hours. We're adding a mostly empty Stripe Sample that viewers can install with `stripe samples create developer-office-hours` which will serve as a standard starting point with boilerplate already populated so the user only needs to write the code to grasp the concept covered in the episode.

### DIFF
--- a/pkg/samples/list.go
+++ b/pkg/samples/list.go
@@ -138,6 +138,12 @@ var connectOnboardingStandard = &SampleData{
 	URL:         "https://github.com/stripe-samples/connect-onboarding-for-standard",
 }
 
+var developerOfficeHours = &SampleData{
+	Name:        "developer-office-hours",
+	Description: "Starter templates for following along with Developer Office Hours",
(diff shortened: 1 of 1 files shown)


tomer-stripe:
Oh wait @cjavilla-stripe do you want the CLI to have a prompt to select the integration? If you change the name to `main` instead of `starter` (in .cli.json) it'll automatically select that.

cjavilla-stripe:
@tomas-stripe I do want to be able to select different demos, for now, it's just the starter but in the future we'll make demo specific samples. I tested locally and with starter it actually worked out of the box as the default without needing to select

cjavilla-stripe:
Sorry I cc'd the wrong person :) @tomer-stripe

PR #393, opened 2020-03-09
[WIP] Get samples from remote url

### LINKED ISSUE #391: Pull Stripe Samples list from GitHub API
### Problem
Today we hardcode a list of samples: https://github.com/stripe/stripe-cli/blob/master/pkg/samples/list.go. That has the disadvantage that we need to release a new version of the CLI for the sample to be available to users. 

### Feature
Instead of hardcoding a list within the CLI, utilise the GitHub API to pull a list of repos from the `stripe-samples` org.

thorsten-stripe:
Closes #391 
Closes #392 

 ### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
@tomer-stripe @adreyfus-stripe POC for hosting a JSON list of the Stripe samples on GitHub.io and pull the list of samples from there instead of hardcoding them in the Stripe CLI.

In this POC I'm fetching from https://github.com/thorsten-stripe/stripe-cli-gh-pages/blob/gh-pages/samples.json but longer-term I'd propose moving that part to https://github.com/stripe/stripe.github.io and maintain the list of samples there. That way we don't have to release a new version of the CLI every time we add a sample.

Wdyt?

### DIFF
--- a/pkg/cmd/samples/create.go
+++ b/pkg/cmd/samples/create.go
@@ -14,7 +14,6 @@ import (
 	gitpkg "github.com/stripe/stripe-cli/pkg/git"
 	"github.com/stripe/stripe-cli/pkg/samples"
 	"github.com/stripe/stripe-cli/pkg/validators"
-	"github.com/stripe/stripe-cli/pkg/version"
 
 	"gopkg.in/src-d/go-git.v4"
 )
@@ -53,7 +52,6 @@ local configuration to let you get started faster.`,
 }
 
 func (cc *CreateCmd) runCreateCmd(cmd *cobra.Command, args []string) error {
-	version.CheckLatestVersion()
 	samples.InitSampleList()
 
 	if len(args) == 0 {
--- a/pkg/cmd/samples/create.go
+++ b/pkg/cmd/samples/create.go
@@ -54,6 +54,7 @@ local configuration to let you get started faster.`,
 
 func (cc *CreateCmd) runCreateCmd(cmd *cobra.Command, args []string) error {
 	version.CheckLatestVersion()
+	samples.InitSampleList()
 
 	if len(args) == 0 {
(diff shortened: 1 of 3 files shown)


ctrudeau-stripe:
This is a great idea to make it easier to update the samples list. 

Another method could be to pull from a raw github JSON file. Not sure if there is a preferred way.

thorsten-stripe:
Thanks @ctrudeau-stripe. My assumption was that GitHub optimises the GitHub.io hosted content and therefore prefers fetching that, but that assumption could be wrong and a JSON file within the CLI master would be better. We should investigate 👍

tomer-stripe:
I like the idea but I'm a bit hesitant about pulling a JSON file from the static site. My main concern is whether or not that's technically within the allowed uses. I'm being a bit cautious here but if an issue were to come up (like github decides that anything on github.io should be browser-only), we'd have to release a new version with a fix which could be frustrating for users.

@thorsten-stripe did you look into cloning a repo? We have mechanisms to do that so it shouldn't be too bad to do.

ob-stripe:
I share Tomer's concerns, using GitHub Pages for this seems like a misuse. I would prefer if the samples definition file were hosted on a Stripe server.

auchenberg:
We already seem to be hosting some static files for webhooks IPs on `stripe.com`. Can we piggyback on this? https://stripe.com/docs/ips#downloading-ip-address-lists

thorsten-stripe:
Moving this to #399

@tomer-stripe had the great idea to use the existing clone, cache, and pull mechanisms to get the samples list from a remote repository: https://github.com/stripe-samples/samples-list

This might also allow us to have a GitHub action that automatically updates the directory repo anytime a new repository is added to the stripe-sample org \o/

PR #399, opened 2020-03-13
Get samples list from remote repository

thorsten-stripe:
Closes #393 

 ### Reviewers
r? @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
* Use the existing clone, cache, and pull mechanisms to get the samples list from a remote repository: https://github.com/stripe-samples/samples-list

@tomer-stripe could you review? Not a Golang expert and I'm not sure it's a robust implementation. Also `samples_test.go` is failing :(

### DIFF
--- a/pkg/cmd/samples/create.go
+++ b/pkg/cmd/samples/create.go
@@ -35,10 +35,9 @@ func NewCreateCmd(config *config.Config) *CreateCmd {
 		forceRefresh: false,
 	}
 	createCmd.Cmd = &cobra.Command{
-		Use:       "create <sample> [destination]",
-		ValidArgs: samples.Names(),
-		Args:      validators.MaximumNArgs(2),
-		Short:     "Setup and bootstrap a Stripe Sample",
+		Use:   "create <sample> [destination]",
+		Args:  validators.MaximumNArgs(2),
+		Short: "Setup and bootstrap a Stripe Sample",
 		Long: `The create command will locally clone a sample, let you select which integration,
 client, and server you want to run. It then automatically bootstraps the
 local configuration to let you get started faster.`,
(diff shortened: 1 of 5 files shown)


pkg/cmd/samples/create.go
```diff
@@ -35,10 +35,9 @@ func NewCreateCmd(config *config.Config) *CreateCmd {
 		forceRefresh: false,
 	}
 	createCmd.Cmd = &cobra.Command{
-		Use:       "create <sample> [destination]",
-		ValidArgs: samples.Names(),
```
tomer-stripe: Just occurred to me: one downside of this approach is that the names will no longer autocomplete. Not a blocker but calling it out.

pkg/samples/list.go
```diff
@@ -24,172 +35,94 @@ func (sd *SampleData) GitRepo() string {
...
+		return err
+	}
 
-var oxxoPayment = &SampleData{
-	Name:        "oxxo-payment",
-	Description: "Build a payment form to collect OXXO and card payments",
-	URL:         "https://github.com/stripe-samples/oxxo-payment",
-}
+	if _, err := s.Fs.Stat(listPath); os.IsNotExist(err) {
+		err = s.Git.Clone(listPath, "https://github.com/stripe-samples/samples-list.git")
```
tomer-stripe: Could you set the url as a `const`?

pkg/samples/list.go
```diff
@@ -24,172 +35,94 @@ func (sd *SampleData) GitRepo() string {
...
 
-var connectOnboardingStandard = &SampleData{
-	Name:        "connect-onboarding-for-standard",
-	Description: "Learn how to onboard users for Stripe Connect with Onboarding for Standard",
-	URL:         "https://github.com/stripe-samples/connect-onboarding-for-standard",
-}
+// GetSamples returns a list that contains a mapping of Stripe Samples that
+// we want to be available in the CLI to some of their metadata.
+// TODO: what do we want to name these for it to be easier for users to select?
+// TODO: should we group them by products for easier exploring?
```
tomer-stripe: @adreyfus-stripe do you have any thoughts here?

adreyfus-stripe: being able to filter by product could be neat!

pkg/cmd/samples/create.go
thorsten-stripe: Is this autocomplete via tab key? That doesn't seem to work today either.

pkg/samples/list.go
thorsten-stripe: I'll leave that for a separate sample. Discussing with @ctrudeau-stripe how to best approach this.

thorsten-stripe:
Thanks for the review! I've moved the URL to a const and fixed the test. 

PTAL :)

pkg/cmd/samples/create.go
tomer-stripe: Yea, I think the homebrew installation stuff is busted :(. Probably not a big deal though, this doesn't seem like something people are going to tab complete.

PR #411, opened 2020-03-28
Add subscription schedules to idURLMap

Just-Sieb:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Earlier today I tried fetching a subscription schedule object with `stripe get sub_sched_XXX` and found that it was not doing the automated detection of the proper url path. I added it to the map, but also had to modify the regex. From what I have gathered of the Stripe id pattern, I think the new regex should also still work well. I had to extend the length of the first capture group and allow underscores in it. It passed my smoke test of working with subscription schedules object ids not as well as some customer and subscription ones as well.


tomer-stripe:
Thanks!

We're having some trouble with Travis right now so I'm going to hold on merging this but this looks good and I don't think you need any changes right now.

tomer-stripe:
@Just-Sieb we got Travis working again, would you be able to push an empty commit to get a build started?

Just-Sieb:
@tomer-stripe I did and noticed one of the tests is failing. I am trying to figure out the best way to update the regex to get it working. I have a regex that works on test sites however it does not look like the Go regex parser allows [conditional regex statements](https://www.regular-expressions.info/conditional.html) since that causes it to fail to compile. I will see if I can come up with something that works.

Issue #408, opened 2020-03-26
Add logout command


hdoupe:
### Problem
I'm trying to clear the credentials for my local stripe cli install, but there doesn't appear to be a logout command.

### Feature
A stripe logout command seems like it would be very helpful. Many other cli's offer this option:

```
$ heroku logout
Logging out... done
```
```
$ stripe logout
```

### Examples
My current use case is that I set up the stripe cli account and it automatically linked my pc up with a client's test webhooks. I would like to remove my device from the webhooks page.

richardm-stripe:
Released in https://github.com/stripe/stripe-cli/releases/tag/v1.4.0, in #430

hdoupe:
Thanks for implementing this @richardm-stripe!

richardm-stripe:
Ha, credit goes to @ianjabour-stripe! I'm just closing issues.

PR #430, opened 2020-04-24
Add `stripe logout` command

ianjabour-stripe:
### Reviewers
r? @tomer-stripe @ob-stripe 
cc @stripe/dev-platform

 ### Summary
In response to #408 , this PR adds a command that lets users clear the credentials for one or all of their projects within their configuration.

- `$ stripe logout`: Will clear the credentials for the default project.
- `$ stripe logout --project-name <project>`: Will clear the credentials for the specified `<project>`
logout

- `$ stripe logout --all`: Will clear the credentials for all projects listed in user's config. 
logout_all


Chucky122:
Reviewers

r? @
cc @

Summary

In response to #408 , this PR adds a command that lets users clear the credentials for one or all of their projects within their configuration.

$ stripe logout: Will clear the credentials for the default project.
$ stripe logout --project-name <project>: Will clear the credentials for the specified <project>
$ stripe logout --all: Will clear the credentials for all projects listed in user's config.
You can view, comment on, or merge this pull request online at:

  https://github.com/stripe/stripe-cli/pull/430

Commit Summary

Add basic implementation of logout command.
Add logic to print message if user is already logged out when running logout command
Add logic to include the project name when -p, --project-name flag used.
Add --all flag to logout command and clean up some shared logic.
File Changes

A pkg/cmd/logout.go (37)
M pkg/cmd/root.go (1)
M pkg/config/config.go (38)
M pkg/config/profile.go (15)
A pkg/logout/logout.go (35)
A pkg/logout/logout_all.go (21)
Patch Links:

https://github.com/stripe/stripe-cli/pull/430.patch
https://github.com/stripe/stripe-cli/pull/430.diff
—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub, or unsubscribe.






Envoyé de mon iPhone Charles

ianjabour-stripe:
I'm struggling to test `ClearCredentials` and `ClearAllCredentials`, basically been manually testing in the CLI.

Any idea of how I can best test these two functions? I was trying to use follow the patterns used in `pkg/config/profile_test.go` with `TestWriteProfile` / `TestWriteProfilesMerge` but when it comes to deleting keys and merging in the new config it doesn't seem to stick. 😕

pkg/config/config.go
```diff
@@ -192,6 +200,37 @@ func (c *Config) PrintConfig() error {
 	return nil
 }
 
+// ClearAllCredentials clears all API key fields from every Profile.
+func (c *Config) ClearAllCredentials() error {
+	runtimeViper := viper.GetViper()
+	var err error
+
+	for field, value := range runtimeViper.AllSettings() {
+		// TODO: ianjabour - ideally find a better way to identify projects in config
```
tomer-stripe, line 209: Should we refactor some of the config code? @ob-stripe and I have felt like we've been needing too but haven't really yet...

There's a very basic function to check for this here: https://github.com/stripe/stripe-cli/blob/master/pkg/config/profile.go#L133-L136

pkg/logout/logout_all.go
```diff
@@ -0,0 +1,21 @@
+package logout
```
tomer-stripe, line 1: I don't think this needs to be its own file. You could put this in the `logout.go` file and `All` will be exported as normal.

pkg/config/config.go
```diff
@@ -29,6 +29,14 @@ const ColorOff = "off"
 // ColorAuto represents the auto-state for colors
 const ColorAuto = "auto"
 
+// CredentialFieldNames represents all fields tied to credentials for a profile
+var CredentialFieldNames = [4]string{
```
tomer-stripe, line 33: Do we want to delete more than just the credential fields? Presumably you could set certain configs per profile.

ianjabour-stripe: Oh ok, what other fields should we be removing?

pkg/logout/logout_all.go
ianjabour-stripe: Ok sounds good I'll move it into `logout.go`

pkg/config/config.go
ianjabour-stripe: So the issue with using the function you linked above is that it assumes I have the profile on hand. In this case I need to find all profiles in the config. Being that we've been only working with one profile at a time up till now, I wasn't able to find anything that could identify a key in the config as a profile.

pkg/config/config.go
ianjabour-stripe: Or did you mean we should make it possible to specify a list of additional fields to remove via a flag?

tomer-stripe: Well, more so that there could also be non-credential fields that we should remove on logout (like `color`). Is there a way to do this where it removes any field attached to a profile?

ianjabour-stripe: I could definitely truncate the entire profile, but is that what we want to do?

I feel like it might lead to unwanted behavior if we are just blindly deleting everything from someone's profile, especially their default profile.

Otherwise if there are other fields, like `color`, that should be removed along with the credentials we can keep a whitelist of them.

ianjabour-stripe:
Screenshots in the summary above have been updated to show new behavior.

tomer-stripe:
@ianjabour-stripe this looks good! I have one small question: should `--all` just delete the file? I'm wondering if there's a benefit to keeping the mostly-empty file around.

ob-stripe:
> I'm wondering if there's a benefit to keeping the mostly-empty file around.

I think there are a couple of global flags for things like color output. Those shouldn't be affected by `logout`.

ianjabour-stripe:
@ob-stripe just re-tested behavior for:

- `stripe logout`: deletes keys from the default profile
0_logout_default

- `stripe logout --project-name mars`: deletes keys from the mars profile
0_logout_project

- `stripe logout --all`: deletes keys from all profiles
0_logout_all

PR #434, opened 2020-04-28
Add fix around regex pattern used for matching fixture queries

ianjabour-stripe:
### Reviewers
r? @thorsten-stripe @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
The root issue of env variables not working in the fixture `path` was that the
`parsePath` method was not using the same regex pattern that was used in
`parseParam`. This divergence caused the env variables to be supported as
params but not in a path.

After adding a test and creating a common util function to extract the
fixture query matches it was found that even the regex pattern that
seemed to work for both cases needed to be tweaked for the case where
multiple fixture queries are used in a row as with:
```
/v1/charges/${char_bender:id}/capture/${cust_bender:id}
```

The screenshots below show the matching issues with the previous patterns and the new pattern matching. 

### Old patterns
Screen Shot 2020-04-27 at 6 15 13 PM

Screen Shot 2020-04-27 at 6 16 00 PM

### New pattern
Screen Shot 2020-04-27 at 6 21 52 PM

### DIFF
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -42,6 +42,12 @@ type fixture struct {
 	Params interface{} `json:"params"`
 }
 
+type fixtureQuery struct {
+	Name         string
+	Query        string
+	DefaultValue string
+}
+
 // Fixture contains a mapping of an individual fixtures responses for querying
 type Fixture struct {
 	Fs            afero.Fs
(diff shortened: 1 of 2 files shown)


tomer-stripe:
Overall looks good! Will give Thor a chance to look at it tonight but we can merge and release a fix tomorrow morning.

thorsten-stripe:
This looks great, thanks a mill @ianjabour-stripe 🎉 One qq, does the `/` in the regex not need to be escaped? 

`\${([^\|}]+):([^\|}]+)\|?([^\/\n]+)?}`

ianjabour-stripe:
Thanks @thorsten-stripe ! I used https://www.regexpal.com/ to test.

As for the `/`, I believe it doesn't need to be escaped when it's in a character / negated set because the only characters with a special significance inside the brackets are: `\`, `]`, `^`, `-`

If you'd prefer it though I can definitely still escape it. I don't think there's any harm in doing that.

thorsten-stripe:
I'm by no means a regex expert, that's why I wanted to double check with you. I'm fine with either 👍

ianjabour-stripe:
Ok sounds good, thanks Thor 👍

PR #439, opened 2020-04-29
Use mailchain/goreleaser-xcgo Docker image

ob-stripe:
### Reviewers
r? @brandur-stripe @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
This PR enables the use of cgo for all binaries, which should fix the DNS issue on macOS that some users have been observing, caused by a known upstream issue in Go itself (https://github.com/golang/go/issues/12524).

Instead of running GoReleaser directly in Travis, it's now run in a Docker image that is set up specifically to be able to use cgo when compiling binaries for macOS and Windows: https://github.com/mailchain/goreleaser-xcgo.

The Docker image is fairly simple: it's based on the official `dockercore/golang-cross` image, and simply adds GoReleaser and Docker to the container (we need Docker inside the Docker container itself, because GoReleaser needs to be able to push the Docker image it creates to the central repository).

I've tested this as best I could, by running the image locally:
```
# This works
$ docker run --privileged -v $PWD:/Users/ob/go/src/github.com/stripe/stripe-cli -v /var/run/docker.sock:/var/run/docker.sock -w /Users/ob/go/src/github.com/stripe/stripe-cli mailchain/goreleaser-xcgo --snapshot --rm-dist --skip-publish

# Log into the Docker container
$ docker run --privileged -v $PWD:/Users/ob/go/src/github.com/stripe/stripe-cli -v /var/run/docker.sock:/var/run/docker.sock -w /Users/ob/go/src/github.com/stripe/stripe-cli -it --entrypoint /bin/bash mailchain/goreleaser-xcgo

# Inside the Docker container
$ goreleaser --snapshot --skip-publish

# On the host machine, get the macOS binary
$ docker cp bd0aec3fdf03:/Users/ob/go/src/github.com/stripe/stripe-cli/dist/stripe-darwin-amd64_darwin_amd64/stripe .

# Run the binary
$ GODEBUG=netdns=cgo+1 ./stripe listen
go package net: using cgo DNS resolver

# In contrast, brew-installed binary
$ GODEBUG=netdns=cgo+1 stripe listen
go package net: built with netgo build tag; using Go's DNS resolver
```

PR #440, opened 2020-04-30
Fix Docker image

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Fix Docker image.

My change from yesterday to enable cgo (#439) unfortunately broke the Docker image for the CLI, because the Alpine image does not have the required libraries that the cgo-enabled `stripe` binary expects.

I've fixed this by disabling cgo for Linux, and keeping it enabled for macOS and Windows, and verified that the Docker image works again (by building locally). The main reason we wanted cgo was to fix the DNS issue on macOS anyway.

If it turns out that using cgo for Linux is desirable for some reason, there are a few ways we can enable it:
- compile 2 different Linux binaries, one with cgo enabled and one with cgo disabled. Use the first one for distributing and the second one for the Docker image. (This will result in a much more complex `.goreleaser.yml file.)
- compile the Linux binary with cgo enabled, and update the Docker image to support it. (But this will result in a larger Docker image.)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,7 +24,7 @@ builds:
       - -s -w -X github.com/stripe/stripe-cli/pkg/version.Version={{.Version}}
     binary: stripe
     env:
-      - CGO_ENABLED=1
+      - CGO_ENABLED=0
     main: ./cmd/stripe/main.go
     goos:
       - linux
@@ -67,6 +67,8 @@ changelog:
     - '^test:'
 nfpms:
   -
+    builds:
+    - stripe-linux-amd64
     vendor: Stripe
     homepage:  https://stripe.com
     maintainer: Stripe <support@stripe.com>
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -24,7 +24,7 @@ builds:
       - -s -w -X github.com/stripe/stripe-cli/pkg/version.Version={{.Version}}
     binary: stripe
     env:
-      - CGO_ENABLED=1
+      - CGO_ENABLED=0
```
ob-stripe, line 27: Strictly speaking this was all that was needed to fix the issue. The rest is just opportunistic fixes/updates.

brandur-stripe:
Seems fine! You might want to take at the failing build in the matrix — I don't think it's related to your change, but it seems like it's possible to get a data race while running the test suite.

PR #443, opened 2020-04-30
Generate flags for polymorphic scalar types

### LINKED ISSUE #352: stripe plans create not working (product missing)
### Issue
When I try to create a plan using cli, I am unable to provide the "product" parameter. It is required in the API, however, the cli doesn't seem to recognize it. This in effect breaks the whole functionality as product is a required parameter and the call fails without it.

Using stripe_1.2.9_windows_x86_64

ob-stripe:
### Reviewers
cc @stripe/dev-platform

 ### Summary
Generate flags for polymorphic scalar types in resource commands, i.e. polymorphic types where at least one of the types is scalar.

For instance, the `product` parameter in plan creation requests can be a hash or a string. This PR will let users use the string form with `stripe plans create --product=prod_123`, rather than being forced to use the `-d` escape hatch `stripe plans create -d product=prod_123`.

Fixes #352.

### DIFF
--- a/pkg/gen/gen_resources_cmds.go
+++ b/pkg/gen/gen_resources_cmds.go
@@ -154,12 +154,13 @@ func getTemplateData() (*TemplateData, error) {
 
 					if media, ok := requestContent["application/x-www-form-urlencoded"]; ok {
 						for propName, schema := range media.Schema.Properties {
-							// only deal with scalar types for now
-							if !scalarTypes[schema.Type] {
+							scalarType := getScalarType(schema)
+
+							if scalarType == nil {
 								continue
 							}
 
-							properties[propName] = schema.Type
+							properties[propName] = *scalarType
 						}
 					}
 				} else {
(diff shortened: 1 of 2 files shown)


ob-stripe:
Most of the new flags are for date filters, which can either be a timestamp or a hash with `gt`, `gte`, `lt` or `lte` keys: https://stripe.com/docs/api/charges/list#list_charges-created

pkg/gen/gen_resources_cmds.go
```diff
@@ -199,3 +200,32 @@ func parseSchemaName(name string) (string, string) {
...
+		if schema.Type == "string" {
+			if len(schema.Enum) == 1 && schema.Enum[0] == "" {
+				return nil
+			}
+		}
+		return &schema.Type
+	}
+
+	return nil
+}
```
brandur-stripe, line 231: Are there any existing bodies of tests that you could latch into to test the behavior? Maybe not too big of a deal since hopefully you can eyeball the generated output, but might help against a regression.

(Will obviously let you guys determine your favored testing methodology for the CLI, but this is the comment I would have left for most projects ;)

ob-stripe:
I tried to add tests for this but ran into weird build errors, I think due to the fact that the `gen_` scripts define their own `main` packages.

I'll look into refactoring the code later, but going to merge this now so it goes into today's release.

PR #444, opened 2020-05-02
Bundle DigiCert root CA certs in the CLI

ob-stripe:
r? @tomer-stripe @joek-stripe 
cc @stripe/dev-platform 

Bundle DigiCert root CA certs in the CLI, and use these for connections to Stripe's servers instead of the system's cert store.

```sh
$ ./stripe login
Your pairing code is: humble-tough-affirm-kudos
Press Enter to open the browser (^C to quit)^C

$ ./stripe login --dashboard-base https://httpbin.org
Post "https://httpbin.org/stripecli/auth": x509: certificate signed by unknown authority
```


pkg/stripe/certsfs.go
```diff
@@ -0,0 +1,8 @@
+//+build dev
+
+package stripe
+
+import "net/http"
+
+// FS exports the filesystem
+var CertsFS http.FileSystem = http.Dir("../../data/certs")
```
joek-stripe, line 8: Does this require that the certificate files are present locally to the CLI?

I was more expecting that we'd either embed the files via something like https://github.com/markbates/pkger or have them as embedded constant strings. Ideally the CLI is portable as a single golang binary, right?

ob-stripe: The cert files are embedded using https://github.com/shurcooL/vfsgen. The file below (`certsfs_vfsdata.go`) is the generated source file containing the actual data.

joek-stripe: Sorry, I missed that. Otherwise looks good

PR #445, opened 2020-05-04
Update dependencies (May 2020)

tomer-stripe:
### Reviewers
r? @ob-stripe 
cc @stripe/dev-platform

 ### Summary
Had to make one change to the spinner to use `time.Duration` instead of an int, everything else seems to work!


tomer-stripe:
ptal @ob-stripe all set!

PR #452, opened 2020-05-21
Prevent livemode secret keys from being used on the CLI.

bwang-stripe:
### Reviewers
r? @richardm-stripe  
cc @stripe/dev-platform @ob-stripe @tomer-stripe 

 ### Summary
For security reasons, prevent livemode secret keys (`sk_live_***`) from being used on the CLI. Users should instead use restricted keys (`rk_live_***`), which can be created via `stripe login`.


pkg/validators/validate_test.go
```diff
@@ -24,19 +24,24 @@ func TestLegacyAPIKeys(t *testing.T) {
 
 func TestPublishableAPIKey(t *testing.T) {
 	err := APIKey("pk_test_12345")
-	require.EqualError(t, err, "the CLI only supports using a secret or restricted key")
+	require.EqualError(t, err, "for security reasons, the Stripe CLI does not permit the use of secret keys in live mode. To generate restricted keys for use in live mode, use the `stripe login` command")
```
richardm-stripe, line 27: this error message should be more like "...only supports using a restricted key"

bwang-stripe: Fixed in the latest commit by restructuring the error message logic.

Now, users should see less disruption in their expected error messages, since we only display the new error message when a livemode secret key is passed. In all other cases, the default error message stays the same:`"the CLI only supports using a secret or restricted key"`.

bwang-stripe:
Pairing with Richard, and realized that this PR as written would block sk_live*** keys for all CLI commands - instead of just `listen` and `logs tail`. We also feel that the server-side changes and error messages will be sufficient on their own to block sk_live*** keys, so we're okay with closing this PR.

richardm-stripe:
Yes, it occurs to me having this check only in the backend has some advantages re: we can reverse it more easily, or narrow/broaden the scope

PR #456, opened 2020-06-01
Fix Racy Test

richardm-stripe:
### Reviewers
r? @tomer-stripe 
cc @ianjabour-stripe (paired with me)

 ### Summary

This test has been causing us much pain. Apparently the websocket client eventhandler can trigger multiple events. This changes the test to block only until the first event comes through and then just go with that, instead of assuming that only one event will ever happen.

### DIFF
--- a/pkg/websocket/client_test.go
+++ b/pkg/websocket/client_test.go
@@ -2,7 +2,6 @@ package websocket
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/websocket/client_test.go
```diff
@@ -67,19 +66,13 @@ func TestClientWebhookEventHandler(t *testing.T) {
...
-	}()
-
 	select {
-	case <-done:
+	case rcvMsg = <-rcvMsgChan:
 	case <-time.After(500 * time.Millisecond):
 		require.FailNow(t, "Timed out waiting for response from test server")
 	}
 
+	fmt.Println("assertions")
```
richardm-stripe, line 75:
```suggestion
```

pkg/websocket/client_test.go
```diff
@@ -2,6 +2,7 @@ package websocket
 
 import (
 	"encoding/json"
+	"fmt"
```
richardm-stripe, line 5:
```suggestion
```

richardm-stripe:
ptal @tomer-stripe CI is passing now

tomer-stripe:
Wooo!

PR #467, opened 2020-06-16
Fix int/string conversion in error message

SamWhited:
In Go 1.14 this would do the wrong thing silently, but this will cause
tests to fail on Go 1.15.

 ### Reviewers
r? @
cc @stripe/dev-platform

 ### Summary

With the disclaimer that I didn't look at what this code is actually doing, I suspect we don't want a string of one rune that matches the version number. Instead, we likely want the integer version number as a string that represents that integer. In Go 1.14 this would do the wrong thing silently, but in Go 1.15 this will cause a test failure.

### DIFF
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -95,7 +95,7 @@ func NewFixture(fs afero.Fs, apiKey, stripeAccount, baseURL, file string) (*Fixt
 	}
 
 	if fxt.fixture.Meta.Version > SupportedVersions {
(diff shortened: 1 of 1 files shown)


SamWhited:
Oops, nevermind, I did not know about the CLA. I don't sign CLAs. Of course, this is a trivial bug fix that's likely not copyrightable anyways so I hope you'll take this as a bug report and fix it. FWIW I release this tiny bit of code to the public domain where possible, etc.

tomer-stripe:
Closing, resolved this in #502!

Issue #80, opened 2019-08-07
Listen stops silently after computer goes to sleep


renchap:
### Issue
I am using `stripe listen --forward-to` (amazing feature 🙇) on my mac. It is running in the foreground in one terminal tab.
When I leave my computer for some time (goes to sleep) and I come back, the command is still running but no new webhooks are received and forwarded

### Expected Behavior
I would expect the CLI to either quit when not connected, or (better) reconnect automatically

### Steps to reproduce
Run `stripe listen --forward-to`, but the computer to sleep for a while, come back, trigger a webhook and see it delivered

### Environment
macOS, installed using brew

tomer-stripe:
Thanks @renchap! We'll look into this.

rsl:
is there a way with the stripe rubygem to detect if `stripe listen` is really listening? over the past few months of developing for a stripe feature, i've had several times `stripe listen` seems to be running but after trying a checkout, i get caught in a loop because we're never getting webhooks then lo and behold https://dashboard.stripe.com/test/webhooks reveals stripe isn't seeing my computer for some new random reason. restarting cli fixes 100% of the time but it's really painful to hit this and have to reset my database records, delete the customer record on stripe, etc, etc.

Anaphase:
Holy crap! After days of beating my head against a wall, it turns out this was my issue as well. `stripe listen` just stops working after my computer is awaken from sleep, even though the Stripe dashboard says my CLI is connected.

richardm-stripe:
A fix for this went out in 1.4.3 (#468).

jeromeetienne:
I still have the issue ```stripe --listen``` no more work after the computer went to sleep. and im running 1.5.5 on macosx. Any hint ?

im using ```stripe listen --forward-to example.orgr/my/path```

--- 

```
$ stripe -v
stripe version 1.5.5
```

PR #468, opened 2020-06-17
Refresh expired websocket sessions

richardm-stripe:
### Reviewers
r? @ob-stripe 
cc @tomer-stripe 

 ### Summary

- `stripe listen` and `stripe logs tail` now detect when a CLI session is expired, and attempt to create a new one and open a websocket connection with it up to three times, this should fix
https://github.com/stripe/stripe-cli/issues/80

- Change use of `context.Context` to be more idiomatic. From [the docs](https://golang.org/pkg/context/)
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

- Factored `withSIGTERMCancel` out of `proxy.Run` and `tailer.Run`. 

### Testing

- Hardcoded devbox stripecli to return the expired error, verified the command terminates after 3 attempts. Verified that the command can successfully reconnect when I un-hardcode the expiry halfway through.
- Against production, I disconnect wi-fi for 15 minutes and verify the command successfully reconnects. You can see the logs here: https://gist.github.com/richardm-stripe/356cf5fb7e6c802efd2adcc764adc157

### DIFF
--- a/pkg/logtailing/tailer.go
+++ b/pkg/logtailing/tailer.go
@@ -105,59 +105,84 @@ func New(cfg *Config) *Tailer {
 	}
 }
 
+func withSIGTERMCancel(ctx context.Context, onCancel func()) context.Context {
+	// Create a context that will be canceled when Ctrl+C is pressed
+	ctx, cancel := context.WithCancel(ctx)
+
+	interruptCh := make(chan os.Signal, 1)
+	signal.Notify(interruptCh, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		<-interruptCh
+		onCancel()
+		cancel()
+	}()
+	return ctx
+}
+
+const maxConnectAttempts = 3
+
 // Run sets the websocket connection
 func (t *Tailer) Run(ctx context.Context) error {
 	s := ansi.StartSpinner("Getting ready...", t.cfg.Log.Out)
 
-	// Create a context that will be canceled when Ctrl+C is pressed
-	ctx, cancel := context.WithCancel(ctx)
-	signal.Notify(t.interruptCh, os.Interrupt, syscall.SIGTERM)
-
-	go func() {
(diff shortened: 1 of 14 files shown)


pkg/logtailing/tailer.go
```diff
@@ -107,60 +105,84 @@ func New(cfg *Config) *Tailer {
...
+		go t.webSocketClient.Run(ctx)
+		nAttempts++
+
+		select {
+		case <-ctx.Done():
+			ansi.StopSpinner(s, "", t.cfg.Log.Out)
+			t.cfg.Log.Fatalf("Aborting")
+		case <-t.webSocketClient.NotifyExpired:
+			if nAttempts < maxConnectAttempts {
+				s = ansi.StartSpinner("Session expired, reconnecting...", t.cfg.Log.Out)
```
ob-stripe, line 180: Do we need to stop the old spinner first? Or does starting a new spinner automatically stops the existing one?

richardm-stripe: Good call. I had not been testing with color on. It was a jittery mess. I fixed it in the latest.

pkg/logtailing/tailer.go
```diff
@@ -107,60 +105,84 @@ func New(cfg *Config) *Tailer {
...
 
 	go func() {
+		<-interruptCh
+		onCancel()
+		cancel()
+	}()
+	return ctx
+}
+
+const maxConnectAttempts = 3
```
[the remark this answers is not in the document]
richardm-stripe: I doubt it? If refreshing doesn't work on the first attempt, it's very unlikely to work on subsequent attempts. So if bumping this value makes any material difference it means something else is wrong.

richardm-stripe:
ptal @ob-stripe 

Added a test, fixed the spinner

Issue #370, opened 2020-02-11
Play nice when output is piped to a pager such as `less`


fred-stripe:
### Problem

Sometimes the output from `stripe` is long, and I want to page through it with `less` rather than grab the mouse. I also like color highlighting, so I have an alias that always passes `--color=on` and pipes the output to `less -R` (this is how I use [Silver Searcher](https://github.com/ggreer/the_silver_searcher), etc).

However, it seems that stripe-cli still outputs the animation even when the output is not a tty, which results in a bunch of extra fluff at the top:

![Screen Shot 2020-02-11 at 11 45 30 AM](https://user-images.githubusercontent.com/23065173/74259720-59ad8100-4cc6-11ea-9f4e-af7ce7b3b69b.png)

### Feature

Minimal solution: detect if the output's a tty and disable animation when writing to a pipe.

Fancy solution: like some other Unix utilities, stripe-cli could detect PAGER and use that if the output is longer than the terminal's number of rows (or some threshold like 40 rows).

This is tricky, though, because you can't know a priori whether or not a pager supports color control sequences like less -R (-R is the "Raw" option to less), since you can't use the usual mechanisms to detect TTY features. The approach I've seen other tools take is to disable color when writing to a pipe unless explicitly asked for on the command line with --color=yes (`ls` and `ag` ~work like this for example).

Thanks as always for a great tool—this seems a bit self-congratulatory coming from another Stripe, but this tool genuinely makes my day of creating, running, and testing Stripe integrations easier—and for that I am grateful. Cheers!

polybuildr:
Hi @fred-stripe, I've taken a stab at this in #479 if you're interested! :)

bwang-stripe:
#479, which disables animation for pipes, is merged in v1.4.6, so I'm marking this as closed. Thanks @polybuildr!

PR #479, opened 2020-07-09
Don't use animated spinner for non-terminal Writers

polybuildr:
### Reviewers
r? @tomer-stripe 

 ### Summary
This change disables animated spinners for non-terminal Writers even when colors are forced with `--color=on`. Relevant use-case in #370.

### DIFF
--- a/pkg/ansi/ansi.go
+++ b/pkg/ansi/ansi.go
@@ -120,10 +120,10 @@ func getCharset() charset {
 
 const duration = time.Duration(100) * time.Millisecond
 
-// StartNewSpinner starts a new spinner with the given message. If the writer doesn't
-// support colors, it simply prints the message.
+// StartNewSpinner starts a new spinner with the given message. If the writer is not
+// a terminal or doesn't support colors, it simply prints the message.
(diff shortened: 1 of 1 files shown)


pkg/ansi/ansi.go
```diff
@@ -120,10 +120,10 @@ func getCharset() charset {
 
 const duration = time.Duration(100) * time.Millisecond
 
-// StartNewSpinner starts a new spinner with the given message. If the writer doesn't
-// support colors, it simply prints the message.
+// StartNewSpinner starts a new spinner with the given message. If the writer is not
+// a terminal or doesn't support colors, it simply prints the message.
 func StartNewSpinner(msg string, w io.Writer) *spinner.Spinner {
-	if !shouldUseColors(w) {
+	if !isTerminal(w) {
```
fred-stripe, line 126: Shouldn't this be the same as the corresponding check in StopSpinner()? 
```go
if !shouldUseColors(w) || !isTerminal(w) {
```

polybuildr: Whoops, of course! Fixed now.

polybuildr:
Hello! Just checking to see if this change looks fine to you @tomer-stripe :)

PR #486, opened 2020-07-15
Introduce a beta `playback` command

bwang-stripe:
### Reviewers
r? @richardm-stripe @tomer-stripe 
cc @stripe/dev-platform

 ### Summary
`stripe playback` is a prototype feature for the Stripe CLI. It is still under development. This project is inspired by the [VCR](https://github.com/vcr/vcr) approach to testing. This aims to make VCR-like testing for Stripe integrations easier and more accessible across our SDK languages, and provide a foundation for new Stripe-specific testing approaches in the future. 

The `stripe playback` command starts  Stripe API proxy server on your local machine. By pointing your API calls to this server when running tests, users can record all of their interactions with the Stripe API. These recordings are saved to a serialized format, and can be replayed on all subsequent runs of those same test. Playing back recordings yields faster, more consistent test suites, that don't need to talk to the `api.stripe.com`. This can be useful for CI environments.

Since `playback` runs as a separate HTTP proxy server on your machine, rather than a native library like VCR, it's harder to do configuration from inside tests. As part of our work, we're looking into ways to make this easier, such as providing wrapper packages for each SDK language.

More info on the current functionality is in `pkg/playback/README.md`.


pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+"replay": All received requests are terminated at the playback server, and responses are played back[1] from a cassette file. A existing cassette most be loaded.
+
+"auto": The server determines whether to run in "record" or "replay" mode on a per-cassette basis. If the cassette exists, operates in "replay" mode. If not, operates in "record" mode.
+
+Currently, stripe playback only supports serving over HTTP.
+
+[1]: requests are currently replayed sequentially in the same order they were recorded.
+` + endpointsDocString,
+		Example: `stripe playback
+  stripe playback --replaymode
```
richardm-stripe, line 77:
```suggestion
  stripe playback --mode replay
```

pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+The playback command starts a local proxy server that intercepts outgoing requests to the Stripe API.
+
+There are three modes of operation:
+
+"record": Any requests received are forwarded to the api.stripe.com, and the response is returned. All interactions
+are written to the loaded 'cassette' file for later playback in replay mode.
+
+"replay": All received requests are terminated at the playback server, and responses are played back[1] from a cassette file. A existing cassette most be loaded.
+
+"auto": The server determines whether to run in "record" or "replay" mode on a per-cassette basis. If the cassette exists, operates in "replay" mode. If not, operates in "record" mode.
```
richardm-stripe, line 70:
```suggestion
"auto" (default): The server determines whether to run in "record" or "replay" mode on a per-cassette basis. If the cassette exists, operates in "replay" mode. If not, operates in "record" mode.
```

richardm-stripe: Oh, I guess can we make auto the default?

pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+	if pc.mode != playback.Auto && pc.mode != playback.Record && pc.mode != playback.Replay {
+		return fmt.Errorf(
+			"\"%v\" is not a valid mode. It must be either \"%v\", \"%v\", or \"%v\"",
+			pc.mode, playback.Auto, playback.Record, playback.Replay)
+	}
+
+	// Check that cassette root directory is valid
+	var absoluteCassetteDir string
+	var err error
+	if filepath.IsAbs(pc.cassetteDir) {
```
richardm-stripe, line 111: Do we need this? `filePath.Abs` docs sounds like it noops if the path is already absolute? https://golang.org/pkg/path/filepath/#Abs

pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+	if !handle.Mode().IsDir() {
+		return fmt.Errorf("The provided `--cassette-root-dir` option is not a valid directory: %v", absoluteCassetteDir)
+	}
+
+	// TODO: disable HTTPS for now. it needs more work to be able to run in a released version of the stripe-cli
+	// eg: how should we package cert.pem and key.pem? (see stripe-mock's implementation)
+	pc.serveHTTPS = false
+
+	// --- Start up the playback HTTP server
+	// TODO: `playback` should handle setup (and teardown) of `stripe listen` as wcorell
```
richardm-stripe, line 137:
```suggestion
	// TODO: `playback` should handle setup (and teardown) of `stripe listen` as well
```

pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+			return fmt.Errorf("the directory \"%v\" does not exist. Please create it, then re-run the command", absoluteCassetteDir)
+		}
+		return fmt.Errorf("Unexpected error when checking --cassette-root-dir: %w", err)
+	}
+
+	if !handle.Mode().IsDir() {
+		return fmt.Errorf("The provided `--cassette-root-dir` option is not a valid directory: %v", absoluteCassetteDir)
+	}
+
+	// TODO: disable HTTPS for now. it needs more work to be able to run in a released version of the stripe-cli
```
richardm-stripe, line 132: Yeah let's remove the https related code from this branch entirely

pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	tr := &http.Transport{
+		// TODO: (is this okay?) create a custom client to skip verifying our self-signed HTTPS certificate
+		// TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{Transport: tr}
+
+	// --- Use the flag values to configure the server (over HTTP)
+	resp, err := client.Get(fullAddressString + "/pb/mode/" + pc.mode)
```
richardm-stripe, line 191: Instead of hitting the configuration endpoints after the server starts up, I think it would be more natural to pass in this configuration initially when we call `NewRecordReplayServer`. That way we don't have to worry about `err != nil` or `defer resp.Body.Close`)

pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		return errors.New("Non 200 status code received during startup: " + resp.Status)
+	}
+
+	resp, err = client.Get(fullAddressString + "/pb/cassette/load?filepath=" + pc.filepath)
```
richardm-stripe, line 201: (ditto, re: pass it in when initializing, not via http)

pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+		fmt.Printf("Listening via HTTPS on %v\n", addressString)
+	} else {
+		fmt.Printf("Listening via HTTP on %v\n", addressString)
+	}
+
+	fmt.Printf("Forwarding webhooks to %v\n", webhookAddress)
+	fmt.Println("-----------------------------")
+	fmt.Println()
+
+	select {}
```
richardm-stripe, line 244: ?

richardm-stripe: Does this just block indefinitely?

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+type serializable interface {
+	toBytes() (bytes []byte, err error)
+	fromBytes(input *io.Reader) (val interface{}, err error)
+}
+
+// Contains binary data representing a generic request and response saved in a cassette.
+type interactionType int
+
+const (
+	outgoingInteraction interactionType = iota // eg: Stripe API requests
```
richardm-stripe, line 35: <3

pkg/cmd/playback.go
```diff
@@ -0,0 +1,245 @@
...
+	webhookAddress string
+	serveHTTPS     bool
+}
+
+func newPlaybackCmd() *playbackCmd {
+	pc := &playbackCmd{}
+
+	pc.cmd = &cobra.Command{
+		// This flag doesn't actually hide the command because we use a template to display the --help text
+		// To unhide this command, see getUsageTemplate() in templates.go, where we explicitly hide the `playback` cmd.
```
richardm-stripe, line 54:
```suggestion
		// This flag doesn't actually hide the command because we use a template to display the --help text
		// To unhide this command, see getUsageTemplate() in templates.go, where we explicitly hide the `playback` cmd.
```
If I understand correctly, doesn't the change you made to templates.go *cause* the .Hidden flag to hide the command? i.e. if I added Hidden to `stripe listen`, it would be hidden without me having to make another change in templates.go?

If that's the case, we should remove the comment.

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// _, err = recorder.fileHandle.Write(reqBytes)
+	// recorder.fileHandle.Write([]byte("\n"))
+
+	// if err != nil {
+	// 	return err
+	// }
+
+	// _, err = recorder.fileHandle.Write(respBytes)
+	// recorder.fileHandle.Write([]byte("\n"))
```
richardm-stripe, line 93:
```suggestion
```

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return err
+}
+
+func (recorder *interactionRecorder) close() error {
+	// Write everything to a YAML File
+
+	// Put everything in a wrapping CassetteYaml struct that can be marshaled
```
richardm-stripe, line 101:
```suggestion
```
self-documenting enough

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+	// 	return err
+	// }
+
+	// _, err = recorder.fileHandle.Write(respBytes)
+	// recorder.fileHandle.Write([]byte("\n"))
+
+	return err
+}
+
+func (recorder *interactionRecorder) close() error {
```
richardm-stripe, line 98: Maybe "saveAndClose" though?

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	_, err = recorder.writer.Write(yamlBytes)
+	return err
+}
+
+// A interactionReplayer contains a set of recorded interactions and exposes
+// functionality to play them back.
+type interactionReplayer struct {
+	historyIndex int
```
richardm-stripe, line 117: maybe `current`, `cursor` or `cur`

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// requestComparator compares 2 structs in the context of comparing a given request struct against
+// a request recorded in the request/
+// It then determines 1) whether that are equivalent 2) whether we should short-circuit
+// our search (return this one immediately, or keep scanning the cassette)
+type requestComparator func(req1 interface{}, req2 interface{}) (accept bool, shortCircuitNow bool)
+
+// A struct that knows how to serialize and de-serialize itself
+type serializable interface {
```
richardm-stripe, line 26: This feels a bit odd to me, come to think of it -- the interface I would want for this would be the ability to just do
`thing, err := fromBytes(...)` and get `thing` (as an empty interface that I have to cast). But with this interface you have to
```go
deserializer := thing{}
thing, err := deserializer.fromBytes(...)
```
and so you're constructing this extra deserializer that isn't actually used.

What I think would be better is to change interactionReplayer so that instead of
```go
type interactionReplayer struct {
  ...
  reqType  serializable
  respType serializable
}
```

You have
```go
type interactionReplayer struct {
  ...
  serialize     func (interface{}) (bytes []byte, err error)
  deserialize func (*io.Reader) (interface{}, err error)
}
```
I *think* this is equivalent

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+func (replayer *interactionReplayer) write(req serializable) (resp *interface{}, err error) {
+	if len(replayer.cassette.Interactions) == 0 {
+		return nil, errors.New("nothing left in cassette to replay")
+	}
+
+	var lastAccepted interface{}
+	acceptedIdx := -1
+
+	// TODO: this can be optimized to do the deserialization from bytes
+	// once per interaction, instead of every time
```
richardm-stripe, line 152:
```suggestion
	// TODO: this can be optimized to do the deserialization from bytes
	// once per interaction, instead of every time
```

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+	for idx, val := range replayer.cassette.Interactions {
+		// Deserialize the recorded request in this interaction
```
richardm-stripe, line 154:
```suggestion
```

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+		var reader io.Reader = bytes.NewReader(val.Request)
```
richardm-stripe, line 155:
```suggestion
		// Deserialize the recorded request in this interaction
```

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+		requestStruct, err := replayer.reqType.fromBytes(&reader)
+		if err != nil {
+			return nil, fmt.Errorf("Error when deserializing cassette: %w", err)
+		}
+
+		// Compare it with the provided request
+		accept, shortCircuit := replayer.comparator(requestStruct, req)
```
richardm-stripe, line 162:
```suggestion
		accept, shortCircuit := replayer.comparator(requestStruct, req)
```

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+
+		// If it matches, then deserialize the matching recorded response
+		if accept {
```
richardm-stripe, line 165:
```suggestion
		if accept {
```

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+			acceptedIdx = idx
+
+			if shortCircuit {
+				break
+			}
+		}
+	}
+	if acceptedIdx != -1 {
+		// remove the matched event
+		replayer.cassette.Interactions = append(replayer.cassette.Interactions[:acceptedIdx], replayer.cassette.Interactions[acceptedIdx+1:]...)
```
richardm-stripe, line 183:
```suggestion
		replayer.cassette.Interactions = append(replayer.cassette.Interactions[:acceptedIdx], replayer.cassette.Interactions[acceptedIdx+1:]...)
```

pkg/playback/core_playback_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	s2 := basicEvent{Name: "Event 2", ID: 46}
+	r2 := basicEvent{Name: "Response 2", ID: 46}
+	fmt.Printf("%+v | %+v\n", s2, r2)
+	recorder.write(outgoingInteraction, s2, r2)
+
+	err = recorder.close()
+
+	if err != nil {
+		panic(err)
```
richardm-stripe, line 151: use `t.FailNow` or `t.Fatal` instead of panic

pkg/cmd/playback.go
bwang-stripe: Ah true, good catch.

pkg/playback/core_playback_test.go
```diff
@@ -0,0 +1,241 @@
...
+	fmt.Println("Recording...")
+
+	s1 := basicEvent{Name: "Event 1", ID: 23}
+	r1 := basicEvent{Name: "Response 1", ID: 23}
+	fmt.Printf("%+v | %+v\n", s1, r1)
+	recorder.write(outgoingInteraction, s1, r1)
+
+	s2 := basicEvent{Name: "Event 2", ID: 46}
+	r2 := basicEvent{Name: "Response 2", ID: 46}
+	fmt.Printf("%+v | %+v\n", s2, r2)
```
richardm-stripe, line 145: Let's remove the logging statements from the test, also probably the comments

pkg/cmd/playback.go
bwang-stripe: Yep!

pkg/playback/core_playback.go
```diff
@@ -0,0 +1,210 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+
+	"gopkg.in/yaml.v2"
+)
+
+func check(err error) {
```
richardm-stripe, line 13: This should be removed since you only use it in the tests. Your test version should probably be something like
```go
func check(t *testing.T, err error) {
  t.Fatal(err)
}
```

pkg/playback/serializer.go
```diff
@@ -0,0 +1,72 @@
+package playback
```
richardm-stripe, line 1: filename, maybe http_serializer?

pkg/cmd/playback.go
bwang-stripe: You're right, will fix!

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,660 @@
...
+func handleErrorInHandler(w http.ResponseWriter, err error, statusCode int) {
+	if err == nil {
+		return
+	}
+
+	// This line will log a relatively benign 'superfluous response.WriteHeader' line if
+	// w.WriteHeader was already called on this http.ResponseWriter. Ideally, response.WriteHeader
+	// should not be called before this function is called, since the HTTP response code can only be
+	// set once.
+	// TODO: refactor usage of this function so that we do not get this 'superfluous' log message
```
richardm-stripe, line 26: Let's do this now

pkg/playback/core_playback.go
bwang-stripe: Yeah, that's a bit clearer - done.

pkg/playback/core_playback.go
bwang-stripe: went with cursor!

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,660 @@
...
+	rr.cassetteLoaded = true
+}
+
+// InitializeServer sets up and returns a http.Server that acts as a playback proxy
+func (rr *RecordReplayServer) InitializeServer(address string) *http.Server {
+	customMux := http.NewServeMux()
+	server := &http.Server{Addr: address, Handler: customMux}
+
+	// --- Webhook endpoint
+	customMux.HandleFunc("/pb/webhooks", rr.webhookHandler)
```
richardm-stripe, line 552: Probably `/playback` > `/pb`

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,660 @@
...
+func (httpReplayer *replayServer) initializeServer(address string) *http.Server {
+	customMux := http.NewServeMux()
+	customMux.HandleFunc("/", httpReplayer.handler)
+
+	server := &http.Server{Addr: address, Handler: customMux}
+
+	return server
+}
+
+// These constants define the modes the playback server can be in
```
richardm-stripe, line 385: I'd move everything below this line to a separate file

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,660 @@
...
+// InitializeServer sets up and returns a http.Server that acts as a playback proxy
+func (rr *RecordReplayServer) InitializeServer(address string) *http.Server {
+	customMux := http.NewServeMux()
+	server := &http.Server{Addr: address, Handler: customMux}
+
+	// --- Webhook endpoint
+	customMux.HandleFunc("/pb/webhooks", rr.webhookHandler)
+
+	// --- Server control handlers
+	customMux.HandleFunc("/pb/mode/", func(w http.ResponseWriter, r *http.Request) {
```
richardm-stripe, line 555: I would factor the "model" logic out from the "controller" logic in these handlers.

i.e. you'd have
```
func (rr *RecordReplayServer) SwitchMode (modeString string) (bool, string) {...}
func (rr *RecordReplayServer) SetRoot (absoluteCassetteDir string) (bool, string) {...}
func (rr *RecordReplayServer) LoadCassette (filepath string) (bool, string) {...}
```
(where bool is "success" and string is "message"), also for throroughness you should probably be passing the error message through in the response body as well as to stdout -- might help users get the debug output they need in their test output

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,660 @@
...
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	defer resp.Body.Close() // we need to close the body
+
+	if err != nil {
+		handleErrorInHandler(w, err, 500)
+		return
+	}
+	io.Copy(w, bytes.NewBuffer(bodyBytes))
+	// we need to reset the reader to be able to read the body again
+	resp.Body = ioutil.NopCloser(bytes.NewBuffer(bodyBytes))
```
richardm-stripe, line 127: This feels strange -- we shouldn't have to read from resp.Body more than once.
Would it be possible to make "newSerializableHTTPResponse" accept an io.Reader instead of an http.Response? I don't think it's strictly necessary that what we "serialize" is a literal http.Response... it just has to be something that we can construct an http.Response from.

(If this doesn't make sense, I probably don't know what I'm talking about)

richardm-stripe: I guess what I am suggesting is, don't serialize/deserialize the http request/response themselves -- serialize a 
```
struct {
  headers: map[string]string
  body: []byte
  statusCode: int
}
```
and then add a little bit more wrapping code around this to convert to the literal http.Request / http.Response when you need to...

richardm-stripe: (this still might not make sense)

pkg/playback/http_playback_test.go
```diff
@@ -0,0 +1,493 @@
...
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/joho/godotenv"
+	"github.com/stretchr/testify/assert"
+	"github.com/stripe/stripe-go"
```
richardm-stripe, line 18: So there's two disadvantages to this test:

1. It introduces a dependency on stripe-go, which has the potential to cause trouble -- the stripe-go interface might change and break this, we have to worry about keeping it updated, etc. Also stripe-go is a pretty big package.
2. Those `.bin` files are kind of big and introduce some indirection in understanding the test cases.

I think it was worthwhile to write this test the way you have, as proof that vcr works with actual stripe and a useful feedback loop, but I think we should rewrite the test in so that the fixtures are small enough to be defined inline, and are not specific to the stripe API. That is, keep `TestGetFromSimpleWebsite`, but don't hit `https://gobyexample.com` -- always hit the fixture server -- but have the fixture server's response be defined as a string inline (just like, "foo" or something) instead of loading it from a file -- remove the other test cases.

pkg/playback/core_playback.go
bwang-stripe: 😎

pkg/playback/core_playback.go
bwang-stripe: Totally agreed, its a clunky interface. I like the structure that you suggested. 

I ran with that and refactored the serialization in this [commit](https://github.com/stripe/stripe-cli/pull/486/commits/c7b020cc3a89d53d90ae39107b4e93a83a2de98a).

pkg/playback/http_playback.go
bwang-stripe: How about this reorganization?

- `struct ReplayServer` -> `http_replayer.go`
- `struct RecordServer` -> `http_recorder.go`
- rename `RecordReplayServer` -> `PlaybackServer`
- `struct PlaybackServer` -> (stays in) `http_playback.go`

And also, rename `core_playback.go` -> `interactions_core.go` (to distance it from the naming of `http_playback.go`)

pkg/playback/http_playback.go
bwang-stripe: Made this refactor in the latest commit, though slightly different from what you outlined (instead of returning bool & string, I just return error). 

The consequence of this tweak is that the "model" can't specify a success message, so it falls to the controller to specify the success message when it handles the (err == nil) case. This feels like an okay division of responsibility to me.

pkg/cmd/playback.go
```diff
@@ -0,0 +1,196 @@
...
+
+	resp, err = client.Get(fullAddressString + "/playback/cassette/load?filepath=" + pc.filepath)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		return errors.New("Non 200 status code received during startup: " + resp.Status)
+	}
```
bwang-stripe, line 164: (I think we lost the comment you put here earlier Richard, but it still unresolved so paraphrasing again): 

We should change this so that we can pass these values to the playback.Server on initialization too, and not have to do HTTP calls when starting up the server for the first time.

bwang-stripe: Also, I think there are race conditions where the server (which starts up in the goroutine above) doesn't start listening before we send these startup HTTP requests - so a even better reason to get rid of these...

bwang-stripe: Fixed in commit 63e526b5aebc8a7a7fc2122a50090d172df49aa9.

pkg/playback/http_playback.go
bwang-stripe: Addressed this in 0bb511c5b8c910a9d01be9002f79e3364150897c.

pkg/playback/http_playback.go
bwang-stripe: This makes sense! And I implemented this fix in: 9d9fdc6f020ceb8f36df74c454aee83b82dc9c82

pkg/playback/http_playback_test.go
bwang-stripe: Tests are updated as of 5ef6701718627002aabba37e4848137df95523ea

pkg/cmd/playback.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/stripe/stripe-cli/pkg/playback"
+	"github.com/stripe/stripe-cli/pkg/validators"
+)
+
+const defaultPort = 13111
+const defaultWebhookPort = 13112
+const endpointsDocString = `
+--- Controlling the server ---
+You can configure the running server instance via HTTP GET endpoints (prefixed with "/playback/").
```
richardm-stripe, line 18: Can we make these POST?

GETs typically are idempotent and non-effectful -- and these are definitely effectful operations, so as a user of this api I would expect to be instructed to use POST.

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,413 @@
...
+		return nil, err
+	}
+	return res, nil
+}
+
+// Writes to the HTTP response using the given HTTP status code and error text as the response body. Simultaneously, prints the error text to stdout.
+// Note: since http.ResponseWriter streams the response, you will not be able to change the HTTP status code after calling this function.
+func writeErrorToHTTPResponse(w http.ResponseWriter, err error, statusCode int) {
+	w.WriteHeader(statusCode)
+	fmt.Fprintf(w, "%v\n", err)
```
richardm-stripe, line 40: What do you think about writing the error as JSON to the application server, instead of plain text? All the other responses we are forwarding to the server (from Stripe) are JSON, so probably we want a vcr-related error to trigger the same error-handling flow as Stripe errors. If we just send plain text probably the user will get a JSON parse error and it will be harder for them to see the text of the error message.

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,413 @@
...
+
+	// --- Server control handlers
+	customMux.HandleFunc("/playback/mode/", func(w http.ResponseWriter, r *http.Request) {
+		// get mode
+		modeString := strings.TrimPrefix(r.URL.Path, "/playback/mode/")
+
+		wasCassetteLoaded := rr.cassetteLoaded
+		err := rr.switchMode(modeString)
+		if wasCassetteLoaded {
+			fmt.Println("/playback/mode: unloaded the cassette. Please load a new cassette before recording/replaying any new interactions.")
```
richardm-stripe, line 114: (Non-blocking suggestion) WDYT about doing the logging consistent with how we do it in e.g. `stripe listen`? That is, we log with a log.Logger object, rather than writing explicitly to stdout with `fmt.Print*`. This would let us more easily introduce a "debug" log level if necessary, and I think it also makes it easier to add support for a `--json` flag

https://github.com/stripe/stripe-cli/blob/master/pkg/proxy/proxy.go#L68

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,413 @@
...
+		if os.IsNotExist(err) {
+			shouldCreateNewFile = true
+			rr.isRecordingInAutoMode = true
+		} else {
+			shouldCreateNewFile = false
+			rr.isRecordingInAutoMode = false
+		}
+	}
+
+	if shouldCreateNewFile {
```
richardm-stripe, line 348: (Non-blocking suggestion) you can remove a little bit of indirection here:
extract the branches of this if statement into `openCassetteFile`/`openOrCreateCassetteFile` functions. Then instead of setting "shouldCreateNewFile", just call the functions directly.

Then it will be easier to visualize the difference between the cases of your `switch` all in one place.

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,413 @@
...
+			return
+		}
+
+		var statusMsg string
+
+		if isRecording {
+			statusMsg = fmt.Sprintf("Recording to %v", relativeFilepath)
+		} else {
+			statusMsg = fmt.Sprintf("Replaying from %v", relativeFilepath)
+		}
```
richardm-stripe, line 212: (non-blocking) this bit could be its own method

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+func (rr *Server) handler(w http.ResponseWriter, r *http.Request) {
+	if !rr.cassetteLoaded {
+		w.WriteHeader(400)
+		fmt.Fprint(w, "No cassette is loaded.\n")
+		return
+	}
+
+	switch rr.mode {
```
richardm-stripe, line 250: (non-blocking suggestion): 

You could DRY this up a little:
```go
if (rr.isRecording()) {
  ...
} else {
  ...
}
```
if you implemented
```go
func (rr *Server) isRecording() {
  switch rr.mode {
  case Record:
    return true
  case Replay:
    return false
  case Auto:
    return rr.isRecordingInAutoMode
  }
  default: //throw
}
```

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,413 @@
...
+	rr.cassetteLoaded = true
+	return nil
+}
+
+func (rr *Server) ejectCassette() error {
+	if !rr.cassetteLoaded {
+		return fmt.Errorf("tried to eject when no cassette is loaded")
+	}
+
+	switch rr.mode {
```
richardm-stripe, line 385: (this switch also can be deduped if you implement `rr.isRecording()` )

pkg/playback/http_playback.go
```diff
@@ -0,0 +1,413 @@
+package playback
```
richardm-stripe, line 1: just wanted to say I like how you reorganized this file :+1:

pkg/playback/http_playback_test.go
```diff
@@ -0,0 +1,275 @@
...
+	ts := startMockServer(fixtureResponses)
+	defer ts.Close()
+	remoteURL = ts.URL
+
+	// -- Setup Playback server
+	addressString := "localhost:13111"
+	cassetteFilepath := "test_record_replay_single_run.yaml"
+
+	// for now, write cassettes to this directory. Ideally we have a test output folder
+	cassetteDirectory, err := filepath.Abs("")
```
richardm-stripe, line 200: We shouldn't do this.
This would be better: https://golang.org/pkg/io/ioutil/#TempDir

Or build in the ability to inject a mock filesystem (i.e. define a `FileOpener` interface and put it on the server struct and delegate to that) -- but I think tempdir gives higher confidence since this is an integration test more than a unit test.

pkg/playback/http_playback_test.go
```diff
@@ -0,0 +1,275 @@
...
+	// Cleanup file
+	_, err = os.Stat(cassetteFilepath)
+	if err == nil {
+		os.Remove(cassetteFilepath)
+	}
+
+	// --- END REPLAY MODE
+}
+
+// // TODO: Test auto mode on the full server
```
richardm-stripe, line 272: we can track the TODOs in a Jira maybe, let's remove the comments.

pkg/playback/http_recorder.go
```diff
@@ -0,0 +1,143 @@
...
+	recorder, err := newInteractionRecorder(writer, httpRequestToBytes, httpResponseToBytes)
+	if err != nil {
+		return err
+	}
+	httpRecorder.recorder = recorder
+
+	return nil
+}
+
+func (httpRecorder *recordServer) webhookHandler(w http.ResponseWriter, r *http.Request) {
```
richardm-stripe, line 36: can you add docstrings for `webhookHandler` and `handler`? Just a little reminder of what the functions are for/how they are called could go a long way

pkg/playback/http_replayer.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return err
+	}
+
+	httpReplayer.replayer = replayer
+
+	return nil
+}
+
+func (httpReplayer *replayServer) handler(w http.ResponseWriter, r *http.Request) {
```
richardm-stripe, line 42: here, too -- a docstring for `handler` could go a long way

pkg/playback/http_replayer.go
```diff
@@ -0,0 +1,198 @@
...
+		return &httpResponse{}, err
+	}
+
+	wrappedResponse := (*uncastResponse).(httpResponse)
+
+	return &wrappedResponse, err
+}
+
+// Reads any contiguous set of webhook recordings from the start of the cassette
+// The caller must call response.Body().Close() on each of the returned http.Response's, otherwise there will be a resource leak
```
richardm-stripe, line 143: This comment is out of date, right? Since we are using the "wrapped" httpResponse, instead of http.Response directly?

pkg/playback/http_replayer.go
```diff
@@ -0,0 +1,198 @@
...
+		// TODO: test webhooks
+		webhookRequests = append(webhookRequests, webhookHTTPRequest.(*httpRequest))
+
+		var respReader io.Reader = bytes.NewReader(rawWebhookBytes.Response)
+		webhookHTTPResponse, err := httpResponsefromBytes(&respReader)
+		if err != nil {
+			return nil, nil, fmt.Errorf("Error when deserializing cassette to replay webhooks: %w", err)
+		}
+
+		// Caller of this function is expected to call .Body().Close() on each http.Response
```
richardm-stripe, line 184: This comment too.

pkg/playback/http_serializer.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+type httpRequest struct {
+	Method  string
+	Body    []byte
+	Headers http.Header
+	URL     url.URL
+}
+
+func newHTTPRequest(req *http.Request) (wrappedRequest httpRequest, err error) {
```
richardm-stripe, line 76: A docstring explaining the "why" of this file would be good, too.

richardm-stripe:
ptal @bwang-stripe

This is looking great and I think this is almost there! I have two "blocking" comments that you should definitely address before you merge.

- Let's use POST instead of GET (I think this is just a docs change)
- Let's make the test interact only with files in a tempdir and not the cwd.

The rest of my comments, if you agree with them I think it also makes sense to tackle them now, or in an immediate follow-up PR, since it would be good to get the initial commit in soon.

pkg/playback/http_playback.go
richardm-stripe: (This is a pretty big refactor, maybe best to do in a follow-up PR)

pkg/cmd/playback.go
```diff
@@ -0,0 +1,169 @@
...
+			pc.mode, playback.Auto, playback.Record, playback.Replay)
+	}
+
+	// Check that cassette root directory is valid
+	absoluteCassetteDir, err := filepath.Abs(pc.cassetteDir)
+	if err != nil {
+		return fmt.Errorf("Error with --cassette-root-dir: %w", err)
+	}
+
+	handle, err := os.Stat(absoluteCassetteDir)
```
suz-stripe, line 108: there are a couple of references to `handle` and the word on its own makes it a little ambiguous. What is it handling? Perhaps `filepathHandler`

pkg/playback/http_recorder.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		writeErrorToHTTPResponse(w, fmt.Errorf("Error when recording HTTP response to cassette: %w", err), 500)
+		return
+	}
+	// Now we can write to the response:
+	// The header *must* be written first, since writing the body with implicitly and irreversibly set
+	// the status code to 200 if not already set.
+	copyHTTPHeader(w.Header(), wrappedResp.Headers) // header map must be written before calling w.WriteHeader
+	w.WriteHeader(wrappedResp.StatusCode)
+	io.Copy(w, bytes.NewBuffer(wrappedResp.Body))
```
suz-stripe, line 124: are there any opportunities for the cli to panic if any of these lines fail here?

pkg/playback/http_replayer.go
suz-stripe: +1

pkg/playback/http_replayer.go
```diff
@@ -0,0 +1,198 @@
...
+		// we log the error and keep going.
+		// TODO: is the above the appropriate behavior?
+		httpReplayer.replayLock.Wait()       // only send webhooks after the previous request/response is handled
+		httpReplayer.replayLock.Add(1)       // acquire lock so we can send webhooks
+		defer httpReplayer.replayLock.Done() // release lock when done sending webhooks
+
+		for i, webhookReq := range webhookRequests {
+			resp, err := forwardRequest(webhookReq, httpReplayer.webhookURL)
+			if err != nil {
+				fmt.Printf("ERROR when forwarding webhook requests: %v", err)
```
suz-stripe, line 101: Could we perhaps add a unique identifier in the following error messages that call out which request failed to forward? Such as the Stripe API URL called and the method.

pkg/playback/interactions.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		accept, shortCircuit := replayer.comparator(requestStruct, req)
+
+		if accept {
+			var reader io.Reader = bytes.NewReader(val.Response)
+			responseStruct, err := replayer.respDeserializer(&reader)
+
+			if err != nil {
+				return nil, errors.New("error when deserializing cassette")
```
suz-stripe, line 144: this error could be a bit better if it matched line 134

pkg/cmd/playback.go
bwang-stripe: Good point - fixed in 5321bc7f7d7e1992444b86f205e475a8eced928c!

pkg/playback/http_playback_test.go
bwang-stripe: Fixed in d253aa567c91a011f30b1a8024ed988a4dae034e.

pkg/playback/http_replayer.go
bwang-stripe: Thanks for catching - fixed in 2f60063d2bd1e627c2288aadfe50345bdf191a1e.

pkg/playback/http_replayer.go
bwang-stripe: Fixed in 2f60063d2bd1e627c2288aadfe50345bdf191a1e

pkg/playback/http_recorder.go
bwang-stripe: Done! 85caf865a9d8b828a9635da735339795656f2173

pkg/playback/http_replayer.go
bwang-stripe: Done! 85caf865a9d8b828a9635da735339795656f2173

pkg/playback/http_serializer.go
bwang-stripe: Done! 85caf865a9d8b828a9635da735339795656f2173

pkg/playback/http_playback.go
bwang-stripe: Ah good idea, that explains some weird coloring I was seeing. I updated it to return JSON in the same format as the normal Stripe API in 53bfbfdeca119d1b9e93ba6b0e0f5f7e0f4c6ae9.


Here's the before and after when using the Stripe CLI to send a request to the server that returns a `playback` error:

Before (with strange coloring):
Screen Shot 2020-07-23 at 6 01 12 PM

After (with proper coloring):
Screen Shot 2020-07-23 at 6 00 11 PM

pkg/cmd/playback.go
```diff
@@ -0,0 +1,169 @@
...
+	if !handle.Mode().IsDir() {
+		return fmt.Errorf("The provided `--cassette-root-dir` option is not a valid directory: %v", absoluteCassetteDir)
+	}
+
+	// --- Start up the playback HTTP server
+	// TODO: `playback` should handle setup (and teardown) of `stripe listen` as well
+	addressString := pc.address
+	remoteURL := pc.apiBaseURL
+
+	webhookAddress := "http://" + pc.webhookAddress
```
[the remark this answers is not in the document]
bwang-stripe: Ah good point - I updated it so that we expect the user to specify the `http://` part of the URL, and added a check that errors if its omitted. 9e4fa35b2efd20db36e41c9a796d4c7adff40a46

pkg/cmd/playback.go
bwang-stripe: Done!  b652efbf6dbc6bb22db8febd6a00a8ca22e980ad

pkg/playback/http_playback.go
bwang-stripe: Another related change made it make sense to this in this PR - switched over to logrus.Logger (which `stripe listen` uses) in feacb33396bd5fb09a3feef1749ea5d61eba03c3.

pkg/playback/interactions.go
bwang-stripe: done!

pkg/playback/http_recorder.go
bwang-stripe: Oh good catch - I added logging and now have the code check the error returned by `io.Copy` and log.Fatal if non-`nil` . feacb33396bd5fb09a3feef1749ea5d61eba03c3

pkg/playback/http_replayer.go
bwang-stripe: Ah good idea - done in 3eacd92780930d55afa79d0c12f2ee5e7e0aec41.

pkg/cmd/playback.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return fmt.Errorf("Error with --cassette-root-dir: %w", err)
+	}
+
+	handle, err := os.Stat(absoluteCassetteDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("the directory \"%v\" does not exist. Please create it, then re-run the command", absoluteCassetteDir)
+		}
+		return fmt.Errorf("Unexpected error when checking --cassette-root-dir: %w", err)
```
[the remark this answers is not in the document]
bwang-stripe: Done in 27b4eff286a1ea016b3d9c24b3dfea2e308bcaeb

pkg/playback/http_playback.go
bwang-stripe: Done! a03d879fc64e5319744cc5974bb4b4e5c40e7ddc

pkg/playback/http_playback.go
bwang-stripe: DRY'ed up in 961d13ecaf1f0c8830e80fe070101dbe0ce126ab

pkg/playback/http_playback.go
bwang-stripe: Done in 961d13ecaf1f0c8830e80fe070101dbe0ce126ab

pkg/playback/http_playback.go
bwang-stripe: Done in 961d13ecaf1f0c8830e80fe070101dbe0ce126ab

pkg/playback/http_recorder.go
```diff
@@ -0,0 +1,143 @@
...
+
+	wrappedReq, err := newHTTPRequest(r)
+	if err != nil {
+		writeErrorToHTTPResponse(w, fmt.Errorf("Unexpected error processing incoming webhook request: %w", err), 500)
+		return
+	}
+
+	// --- Pass request to local webhook endpoint
+	resp, err := forwardRequest(&wrappedReq, httpRecorder.webhookURL)
+	// TODO: this response is going back to Stripe, what is the correct error handling logic here?
```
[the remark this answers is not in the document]
richardm-stripe: To my understanding, the webhooks are ACKed by our infra before the infra forwards it to `stripe listen` (or stripe playback). The benefit of forwarding the response back is to power a view on the dashboard that shows CLI responses:
image

So, IMO the correct way to handle forwarding errors is to log a warning. That's kind of the best we can do.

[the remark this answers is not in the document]
richardm-stripe: (thanks @bwang-stripe for the screenshot)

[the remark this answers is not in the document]
richardm-stripe: So action item: remove the TODO comment (or replace with a comment explaining the implications of a forwarding failure)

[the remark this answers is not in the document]
bwang-stripe: Makes sense to me, thanks for the discussion all! I did as Richard suggests in 9b14b94057965c8e04d5a35752660766742edc57.

pkg/playback/http_playback_test.go
bwang-stripe: Done - I kept the surviving TODOs but linked them to their Jira tickets so that's its easier for future people to identify where the changes should be made. See b698ea234bb8be7e5d9d2e99d3ffca8356541223.

pkg/cmd/playback.go
```diff
@@ -0,0 +1,183 @@
...
+	// // Hidden configuration flags, useful for dev/debugging
+	pc.cmd.Flags().StringVar(&pc.apiBaseURL, "api-base", "https://api.stripe.com", "The API base URL")
+	pc.cmd.Flags().MarkHidden("api-base") // #nosec G104
+
+	return pc
+}
+
+func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
+	fmt.Println()
+	fmt.Println("Seting up playback server...")
```
richardm-stripe, line 95:
```suggestion
	fmt.Println("Setting up playback server...")
```

pkg/playback/http_playback.go
```diff
@@ -431,28 +424,40 @@ func (rr *Server) ejectCassette() error {
...
+func (rr *Server) isRecording() (isRecording bool, err error) {
+	switch rr.mode {
+	case Record:
+		return true, nil
+	case Replay:
+		return false, nil
+	case Auto:
+		return rr.isRecordingInAutoMode, nil
+	default: //throw
+		return false, fmt.Errorf("Unexpected mode \"%v\"in playback server - this likely indicates a bug in the implementation. Please try restarting the server", rr.mode)
```
richardm-stripe, line 456: It would be better to crash here, I think. Then you wouldn't need to pipe around the extra `err`, and it doesn't really help the UX to continue.

bwang-stripe: Good point! 990af9d10cc2e01d539f4df3cfa58d054609a61b

PR #502, opened 2020-09-16
Upgrade to go 1.15.x and upgrade dependencies

tomer-stripe:
### Reviewers
r? @ob-stripe @richardm-stripe 
cc @stripe/developer-products 

 ### Summary
Upgrade go version and dependencies. goreleaser deprecated `github` in favor of `tap` (https://goreleaser.com/deprecations#brewsgithub) so I'm updating that before it breaks anything.

PR #510, opened 2020-09-30
Undo goreleaser config change for homebrew

tomer-stripe:
### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products

 ### Summary
We're using [`goreleaser-xcgo`](https://github.com/mailchain/goreleaser-xcgo) to run goreleaser in docker on Travis (due to previous issues. That docker image is running version [`0.127.0`](https://github.com/mailchain/goreleaser-xcgo/blob/master/Dockerfile#L5), which doesn't include the new syntax that I updated to in #502. I don't want to try to update the version on the docker image so I'm undoing this one syntax change for now.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -81,7 +81,7 @@ nfpms:
     - rpm
 brews:
   -
-    tap:
+    github:
       owner: stripe
       name: homebrew-stripe-cli
     commit_author:


pepin-stripe (approved):
I don't know what that means or how it works, but LGTM?

PR #525, opened 2020-10-27
Run listen command when starting stripe playback

pepin-stripe:
### Reviewers
r? @richardm-stripe 
cc @stripe/developer-products @tomer-stripe 

 ### Summary
`stripe playback` needs to use `stripe listen` while in record mode so it can intercept webhook events from the stripe API and record them.

fixes DX-6101

### DIFF
--- a/pkg/cmd/playback.go
+++ b/pkg/cmd/playback.go
@@ -140,7 +140,7 @@ func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// --- Setup `stripe listen` and run it now if not in replay-only mode, else listen for changeModeChan.
+	// --- Setup `stripe listen` and run it now if not in replay-only mode, else listen for httpWrapper.ChangeModeChan.
 	lc := newListenCmd()
 	lc.forwardURL = addressString + "/playback/webhooks"
 	startListenCmd := func() {
--- a/pkg/cmd/playback.go
+++ b/pkg/cmd/playback.go
@@ -166,20 +166,6 @@ func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
 			})
 		}
 		listenToModeSwitch()
-		// go func() {
-		// 	modeChan := make(chan string)
-		// 	httpWrapper.OnSwitchMode(modeChan)
-		// 	for {
-		// 		mode := <-modeChan
-		// 		switch strings.ToLower(mode) {
-		// 		case playback.Record, playback.Auto:
-		// 			close(modeChan)
-		// 			startListenCmd()
-		// 		default:
-		// 			continue
-		// 		}
-		// 	}
-		// }()
 	}
 
 	server := httpWrapper.InitializeServer(addressString)
--- a/pkg/cmd/playback.go
+++ b/pkg/cmd/playback.go
@@ -132,7 +132,6 @@ func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
 	}
 
 	// --- Start up the playback HTTP server
-	// TODO(DX-5702): `playback` should handle setup (and teardown) of `stripe listen` as well
 	addressString := pc.address
 	remoteURL := pc.apiBaseURL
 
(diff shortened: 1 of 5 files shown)


pkg/playback/http_playback.go
```diff
@@ -95,6 +95,8 @@ type Server struct {
 
 	log *log.Logger
 
+	SwitchModeChan chan string
```
richardm-stripe, line 98: What do you think about  making the channel private and implementing a public `OnSwitchMode` method for Server that calls its argument? I try to encapsulate channels as much as possible because there's so much you can do with them (close them, listen for messages, send messages, wait for them to be closed) but usually you only intend for them to be used in a particular way (just listening for messages, in this case).

richardm-stripe (changes requested):
This looks good to merge to me except
1. Remove `default_cassette.yaml` (and maybe add it to the .gitignore?) (this is blocking)
2. I Made some food-for-thought i.e. non-blocking comments.
3. PR description should call out how we've tested

pkg/playback/http_playback.go
pepin-stripe: That's a great point, I'll do that indeed!

pkg/cmd/playback.go
```diff
@@ -141,6 +140,31 @@ func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
...
+				mode := <-httpWrapper.SwitchModeChan
+				switch mode {
+				case playback.Record, playback.Auto:
+					startListenCmd()
+				default:
+					continue
+				}
+			}
+		}
+	}()
```
richardm-stripe, line 166: Here we are doing this at the command level, e.g. we are having the playback "command" trigger the listen "command", but I am wondering if you considered doing it further downstream, e.g. have the playback "server" start the proxy "server"?

I don't think it makes much of a difference functionally but you might like the resulting code better. In particular this might allow us to avoid the `lc.forwardURL = addressString + "/playback/webhooks"` bit, which is somewhat awkward, e.g. it is like starting up something external and pointing it at us through the network interface, vs. starting up "our own" proxy server and having access to its events directly.

pepin-stripe: Yeah so originally I was looking into doing that from the playback server, but few things that made me go the command direction:

- The `lc.runListenCmd` internals have a lot of logic (init and start the proxy, etc) and it also uses a lot of cobra specific environment vars (for example `Config.Profile.GetDeviceName()`) so it felt like a lot of things would be duplicated _and_ more complicated.

- It feels like `listen` and `playback` are independent and there shouldn't be a dependency between them. If we change how listen works tomorrow, rename it, etc. `playback` shouldn't really require a refactor. I'm ok having to refactor the higher level logic of this command where everything is setup and being called.

- It is indeed weird to have a command call another command, but actually I see it as the imperative shell of the system, so for me it's ok to have this glue-ish code in there.

what do you think?

What do you think?

pkg/playback/http_playback.go
pepin-stripe: I'm not sure what is the best pattern to do that in Go, can you see what I have and let me know what you think? It doesn't feel very right to me but can't find a better way to handle that...

pkg/playback/http_playback.go
```diff
@@ -262,9 +271,24 @@ func (rr *Server) InitializeServer(address string) *http.Server {
...
 
+// OnSwitchMode blocks on the switchModeChan and returns the value it receives
+// func (rr *Server) OnSwitchMode(send chan<- string) {
+// 	mode := <-rr.switchModeChan
+// 	send <- mode
+// }
+
+// OnSwitchMode blocks on the switchModeChan and returns the value it receives
+func (rr *Server) OnSwitchMode(f func(string)) {
+	go func() {
```
pepin-stripe, line 282: also not sure what the best practice in go is.

calling `go func() ...` inside a function and simply calling this function inline `rr.OnSwitchMode(...)` 

or calling the logic inline without setting up a goroutine inside the function, and having the caller setup a go routing (`go rr.OnSwitchMode(...)`)?

pkg/cmd/playback.go
richardm-stripe: I'm convinced. I think this may be an argument that our "commands" are doing double-duty as our "imperative shell" and our "command-line parsing" layer, but changing this would be out of scope.

pkg/playback/http_playback.go
richardm-stripe: This works!

The other thing you can do is have `OnSwitchMode` put the "handler" (`f`) on the struct, and then start a goroutine up from `NewServer` that reads the channel and fires the handler if set. That might have better behavior w.r.t. if `OnSwitchMode` were to be called multiple times.

richardm-stripe: I think this also only fires the handler once, but previously it would call it multiple times?

richardm-stripe (approved):

pkg/cmd/playback.go
pepin-stripe: yeah I agree - the "commands" should probably just do what cobra offers - the command interaction/args parsing, and then we should have another layer for the actual business logic. Will file a task to look into it at some point!

pkg/playback/http_playback.go
pepin-stripe: Adding the handler on the struct would imply you want an array of handlers in case `OnSwitchMode` is called multiple times, right?

I think this solution is probably good enough for now - I don't love it and hopefully will be able to cleanup later on when we refactor _everything_ !

PR #527, opened 2020-10-29
rename --load-from-webhooks-api to --use-configured-webhooks and keep…

pepin-stripe:
… backwards compatibility

 ### Reviewers
r? @richardm-stripe 
cc @stripe/developer-products

 ### Summary
--load-from-webhooks-api is a very confusing name. We are hoping that --use-configured-webhooks will be a bit more clear.


richardm-stripe (approved):
LGTM

Issue #489, opened 2020-07-22
command to retrieve Stripe Account name


TheVaporTrail:
### Problem
I would like to verify that the cli commands I am about to run against my Stripe account are going to be performed on the correct account. I would like my shell script to present the account name and also "Live"/"Test" and ask for confirmation before continuing with the commands.

I have three Stripe accounts ("Dev", "Test", and "Production"), and Production has "Test" and "Live". 

The script I have either creates the products and subscription levels (with a series of "products create" and "prices create" calls) or retrieves them and reports the values. 

### Feature
A Stripe cli command that would retrieve the account name and the account mode ("test" or "live").

### Examples
I set-up the script so that it requires STRIPE_API_KEY in the environment. The "Project prefix" and "Project name" come from my AWS configuration, which is unique for Dev, Test, and Prod.

```
$./src/.../.../stripeProducts.sh -r

Polygonia: Stripe Billing Configuration

Project Info
  Project prefix: polygoniadev-mobilehub-XXXXXXX
  Project name: PolygoniaDev2

STRIPE_API_KEY must be set
It must match the local Polygonia project directory (Dev, Test, or Production)
$
```
With a STRIPE_API_KEY set the script will report the required values (the two lines after each "Architecture" row) and the values reported from Stripe (the lines after "Reported"). 

I decided to use the metadata in the products and prices to help verify I am reporting from the correct Stripe account.

```
$./.../.../stripeProducts.sh -r

Polygonia: Stripe Billing Configuration

Project Info
  Project prefix: polygoniadev-mobilehub-XXXXXXXX
  Project name: PolygoniaDev2

  Retrieving product list

Hobbyist:
  Architecture
  Price: 400
  Nickname: Hobbyist/monthly
  Reported
  Product Id: prod_ZZZZZZZZZZZZZZZZZZZ
    Mode: TEST
    Metadata project: PolygoniaDev2
    Metadata level: Hobbyist
    Name: Polygonia Design Suite - Hobbyist
  Price Id: price_ZZZZZZZZZZZZZZZZZZZ
    Price: 400
    Metadata project: PolygoniaDev2
    Metadata level: Hobbyist

Artist:
  Architecture
  Price: 900
  Nickname: Artist/monthly
  Reported
  Product Id: prod_ZZZZZZZZZZZZZZZZZZZ
    Mode: TEST
    Metadata project: PolygoniaDev2
    Metadata level: Artist
    Name: Polygonia Design Suite - Artist
[...]

PR #533, opened 2020-11-07
Display account info before resource command execution

ritikbhatia:
### Reviewers
r? @richardm-stripe @tomer-stripe @pepin-stripe 
cc @stripe/developer-products

 ### Summary
This PR is a potential and partial solution to Issue #489 
This PR results in displaying the **mode** (live/test) and **dashboard display name** before execution of a resource command (like `stripe customers create ....` etc.)
It also confirms whether the user wants to proceed with the command or not.
An image for the example output is attached

 ### Future Work
I plan to add this display info and confirmation with the rest of the Stripe CLI commands as well as add a flag in case the user wants to skip this confirmation and account information display.
Confirmation Denied
Confirmed


ritikbhatia:
The latest push has now added a `--skip-confirmation` flag to allow the user to skip the confirmation option for a resource command (as it can be tedious if it is asked for every command)
Attached is a screen snip of using `--skip-confirmation` flag
Skip Confirmation

ritikbhatia:
Changed `--skip-confirmation` flag to `--confirm-account` flag (because asking for confirmation repeatedly can be cumbersome)

So, default is that the command will be executed without asking for confirmation.
In case the user is unsure, he/she can can execute the command with `--confirm-account` flag to confirm the **Stripe account** and the **mode (test/live)** the command is being executed on.

Example: `stripe customers create --description "First Customer" --confirm-account`

richardm-stripe:
Hi @ritikbhatia, thanks for the pull request! This is a substantial change so we'll need to look a bit closely to make sure this is the right choice. @auchenberg-stripe and @tomer-stripe, do you have opinions?

pepin-stripe:
Hi @ritikbhatia, thanks for the PR! 

As @richardm-stripe said, please hold until @auchenberg-stripe or @tomer-stripe confirm that it's fine to add this feature.

Meanwhile, a few comments on the actual PR:
- the `DisplayName` and `getUserAccount` logic should probably live outside of the `resource` package, either in a new package or one where it makes sense to have general account-related logic.
- there is a `Config` package where the APIKey and other info is stored for a session. To avoid having `getUserAccount` make an external request each time a command with the `--confirm-account` flag is passed, I'd look into adding the relevant account information into this `Config` object so we keep it cached. 
- we are trying to increase the overall test coverage, so let's add test please!

ritikbhatia:
Hi @pepin-stripe 
Thanks for the detailed suggestions for improving the PR! I will wait for further approval regarding this feature

Once approval is given, I will work to enhance test coverage, incorporate changes you have mentioned as well as any other changes suggested at a later date, by pushing more commits to this PR itself

ritikbhatia:
@pepin-stripe I have removed the logic of retrieving account display name from the `resources` package and added it at the login step itself(helps to avoid repeated requests for getting the account name)
The value is instead written in the `config` upon login and retrieved from there when required

This works for both **interactive and normal** login

pkg/cmd/resource/operation.go
```diff
@@ -69,8 +69,33 @@ func (oc *OperationCmd) runOperationCmd(cmd *cobra.Command, args []string) error
 
 	oc.Parameters.AppendData(flagParams)
 
-	_, err = oc.MakeRequest(apiKey, path, &oc.Parameters, false)
+	if oc.ConfirmAccount {
+		// display account information anf confirm whether user wants to proceed
```
richardm-stripe, line 73:
```suggestion
		// display account information and confirm whether user wants to proceed
```

ritikbhatia: Thanks @richardm-stripe for spotting the mistake! I have committed the change

ritikbhatia:
@richardm-stripe @pepin-stripe I have added tests as suggested
Tests include:
- `DisplayName` function returning correct display name and retrieving account correctly or not (file: `interactive_login_test.go`)

- `display_name` field being correctly written into `Config` of user profile or not (file: `profile_test.go`)

tomer-stripe:
Thanks for the PR! I like the idea but I think there are some challenges on the server-side for this. For retrieving accounts, unfortunately the restricted access keys we generate with the CLI's login flow (`stripe login`) don't have access to /v1/accounts so anyone who newly logs into the CLI will be met with an error:
```
➜ stripe accounts retrieve
{
  "error": {
    "message": "The provided key 'rk_test_foobar' does not have the required permissions for this endpoint on account 'acct_barfoo'.  This is a restricted API key, but the required permissions are not available for use by restricted keys.",
  }
}
```

We're working on a fix for that but I don't know what the timeline is at the moment. What we should likely do is return the name as part of the login flow but we'd need to update the backend code to do that. I don't think this would be too much work but I need to see if anyone on the team can help out. A few folks also suggested the idea of a `whoami` command which could be cool too! (@pepin-stripe and @auchenberg-stripe have been brainstorming ideas).

The two other things I want to consider are that the commands only output JSON by default (so that it keeps it easy to pipe into other tools, like `jq`) and that we only require confirmation on destructive behaviors (like `DELETE`).

For the first point, that likely puts things more in the shape of adding a `--verbose` flag (similar to our `--show-headers` one). For the latter, if we do want to add confirmation we should find a way to reuse the confirm flag from the base request: https://github.com/stripe/stripe-cli/blob/master/pkg/requests/base.go#L101-L103. I noticed our pre-generated delete commands don't do that (which I think they should) but I'm a little hesitant to add it now in case people were using them in automated flows (but we could discuss doing a minor version bump).

@ritikbhatia what do you think?

ritikbhatia:
@tomer-stripe thank you so much for the detailed response and for suggesting some great improvements!

Following are my thoughts regarding the same:

1. For the first point (regarding restricted access keys using the `stripe login` command), this PR enables the retrieval of account information by making use of the methods similar to the ones used in `SuccessMessage` (https://github.com/stripe/stripe-cli/blob/master/pkg/login/login_message.go) during the login flow. Namely, the `AccountDisplayName` is stored in the `config` file that is created as a result of the `writeProfile()` method and it is retrieved from there itself, as and when required. It makes use of the test/live api keys of the user for accessing the `/v1/accounts` endpoint (similar to the `SuccessMessage` as mentioned earlier and I have added a few tests that confirm that account name retrieval works as expected) and hence no issue of an error arising like the one when using restricted keys (though the suggestion of a `whoami` command actually sounds like a great idea for the same and can be used with the `--confirm-account` flag!)

2. The JSON output for piping into other commands is something I completely agree with and should be the standard response for the commands

3. Finally, I agree that adding a separate `--confirm-account` flag does not make particular sense when there is already a `--confirm` flag in the base requests. I will work to reuse this `--confirm` flag directly for more consistency. As suggested by you, it also makes perfect sense to add this confirmation for the more impactful commands like `DELETE` and `POST`

Apart from the above, I personally feel that the addition of this flag as a resolution for issue #489 could be a good addition, one that gives the user the ability to check the logged in account without having to leave the CLI / execute another command (particularly since a logged in session lasts for 90 days)

Hope this helps and is aligned to your opinions regarding the same! Any other suggestions / clarifications to my response would be helpful @richardm-stripe @pepin-stripe @tomer-stripe !

ritikbhatia:
@tomer-stripe As suggested, the latest push now makes use of the `--confirm` flag that is already present in the base requests, that is, if the `--confirm` flag is passed, auto confirmation will happen, otherwise it will request explicit confirmation from user.
Further, this confirmation is only applicable for `DELETE` requests (eg. `stripe customers delete cus_....`). For other HTTP methods, the confirmation will be by passed, and will proceed to execute the command normally.

cc @richardm-stripe @pepin-stripe

ritikbhatia:
@tomer-stripe any other opinions that you might be having / suggestions that I can use to improve the PR?

pkg/login/interactive_login.go
```diff
@@ -27,6 +30,13 @@ func InteractiveLogin(config *config.Config) error {
 
 	config.Profile.DeviceName = getConfigureDeviceName(os.Stdin)
 	config.Profile.TestModeAPIKey = apiKey
+	displayName, errDisplay := DisplayName(nil, stripe.DefaultAPIBaseURL, apiKey)
+
+	if errDisplay != nil {
+		return err
+	}
```
pepin-stripe, line 37: As @tomer-stripe mentioned in a previous comment, there are some limitations to accessing one's account info depending on the api key used (some api keys are restricted and don't have all accesses, more info in the [docs](https://stripe.com/docs/keys#limit-access)), which means that sometimes, despite a successful login, retrieving the display name will error.

We are looking into resolving that on the backend, but have no timeline for now unfortunately.

Meanwhile, I'd suggest updating the logic so it handles cases when the display name can't be retrieve, instead of throwing an error.

pkg/login/interactive_login.go
```diff
@@ -46,6 +56,52 @@ func InteractiveLogin(config *config.Config) error {
...
+
+		account = acc
+	}
+	displayName := account.Settings.Dashboard.DisplayName
+
+	return displayName, nil
+}
+
+// helper method to get the account for DisplayName
+func getAccount(baseURL string, apiKey string) (*Account, error) {
```
pepin-stripe, line 76: As you mentioned, this function is duplicated from [`login_message.go`](https://github.com/stripe/stripe-cli/blob/master/pkg/login/login_message.go#L65-L92). Let's find a good location for this function and have it shared between the files using it.

pkg/login/interactive_login.go
```diff
@@ -46,6 +56,52 @@ func InteractiveLogin(config *config.Config) error {
 	return nil
 }
 
+// DisplayName returns the display name for a successfully authenticated user
+func DisplayName(account *Account, baseURL string, apiKey string) (string, error) {
```
pepin-stripe, line 60: re: next comment about `getAccount`, this should probably be moved to the same file.

pepin-stripe (changes requested):
I think we're getting close! Left a few additional comments inline.

pkg/login/interactive_login.go
ritikbhatia: Sure! For this, in case there is an error in retrieving the access key, I can change it to allow the execution to flow by assigning an empty string / `could not retrieve account name` to the variable `displayName`

pepin-stripe: I think a more conservative approach (aiming to not disrupt any existing workflow for current users) would be to save the `displayName` as `""` and have the confirmation logic handle this case.

IMO saying "could not retrieve account name" might get confusing for users, so it might just be better to have the confirmation text only mention livemode/testmode and no account name.

ritikbhatia: Okay will proceed with setting `displayName` as "" in this case

pkg/login/interactive_login.go
ritikbhatia: Agreed! I think that since it resides in the `login` logic, we can create a separate file in the `login` folder and have `login_message.go` as well as `interactive_login.go` share these functions

ritikbhatia:
@pepin-stripe I have worked on the changes that you requested. Following are the changes made:
- Added a new file named `retrieve_account.go` that other files share to retrieve the account and use the parameters as needed. Also added `retrieve_account_test.go` for testing this file
- Made the default of `displayName` an empty string so that the command continues to be executed and the confirmation prompt will just show **test/live** mode

Hope these changes are well aligned with how you expected it to be!

pkg/login/interactive_login.go
```diff
@@ -27,6 +27,13 @@ func InteractiveLogin(config *config.Config) error {
 
 	config.Profile.DeviceName = getConfigureDeviceName(os.Stdin)
 	config.Profile.TestModeAPIKey = apiKey
+	displayName, err := getDisplayName(nil, stripe.DefaultAPIBaseURL, apiKey)
+
+	if err != nil {
+		displayName = ""
+	}
+
+	config.Profile.DisplayName = displayName
```
pepin-stripe, line 36: `getDisplayName` returns `"", err` when there is an error, so we don't need lines 32-34 which get overwritten by line 36 anyway.

```go
displayName, _ := getDisplayName(nil, stripe.DefaultAPIBaseURL, apiKey)
config.Profile.DisplayName = displayName
```
should be enough

pepin-stripe (approved):

ritikbhatia:
@pepin-stripe thanks for the approval! Could anyone in the Stripe development team / having write access to the repo merge the PR (as I do not have write access to the repo) ?
cc @richardm-stripe @tomer-stripe

ritikbhatia:
@pepin-stripe Thanks for approving the PR! I have made the change in the latest push (removing the redundancy)

Can someone with write access to the repo (@tomer-stripe @richardm-stripe @pepin-stripe ) merge the PR, since I do not have write access to the repo?
Thank you!

pepin-stripe:
@ritikbhatia we just need a 2nd pair of eyes before merging - @tomer-stripe will try to review the PR before EOW. Thanks!

ritikbhatia:
@pepin-stripe sure, no worries at all! Thanks for the heads up :)

ritikbhatia:
Hey @pepin-stripe @tomer-stripe ! May I ask if there are any updates regarding this PR?

tomer-stripe:
@ritikbhatia really sorry, been swamped the past few weeks. I'll try to take a look at it today!

pkg/cmd/resource/operation.go
```diff
@@ -69,8 +70,38 @@ func (oc *OperationCmd) runOperationCmd(cmd *cobra.Command, args []string) error
...
+		}
+
+		// call the confirm command from base request
+		confirmation, err := oc.Confirm()
+		if err != nil {
+			return err
+		} else if !confirmation {
+			fmt.Println("Exiting without execution. User did not confirm the command.")
+			return nil
+		}
```
tomer-stripe, line 96: Nice, thanks for checking this!

pkg/login/retrieve_account.go
```diff
@@ -0,0 +1,55 @@
+package login
```
tomer-stripe, line 1: Nice! Thanks for refactoring this

tomer-stripe (commented):
This is looking good to me! I'm ready to merge this in if there's nothing else on your side

ritikbhatia:
Hi @tomer-stripe 
Thanks for the positive remarks! I would like to confirm there is nothing else from my side

tomer-stripe:
@ritikbhatia great! Quick question I just thought up: does the `-c, --confirm` flag work here? We have that on the existing `delete` commands and want to double check that users have an override for programmatic flows.

ritikbhatia:
@tomer-stripe thanks for the question.
Yup, the `confirm` flag does work on the DELETE commands. The code that I have added reuses this `confirm` flag as suggested in an earlier review.

tomer-stripe:
Awesome, I'm merging it now then! :D

simonpoortman123 (commented):
Simon Poortman
Bondskanselier, koning

Issue #539, opened 2020-11-17
CLI says FATAL when ctrl-c


auchenberg-stripe:
### Problem
Feedback from team testing:

> CLI-related: ^C[Fri, 13 Nov 2020 13:25:13 PST] FATAL Aborting we should really not have it say FATAL when ctrl+c

PR #540, opened 2020-11-18
[blocked] Playback - Eject Cassette when CTRL+C is pressed

pepin-stripe:
### Reviewers
r? @suz-stripe 
cc @stripe/developer-products

 ### Summary

In the scenario where the user only want to run playback, record an interaction and stop the server, having to manually eject the cassette to persist it to disk is annoying.

This PR solves this by leveraging `WithSIGTERMCancel` and calling `ejectCassette`.

This is blocked by #539 which cancels the cassette ejection when `listen` exits with FATAL.

### DIFF
--- a/pkg/cmd/playback.go
+++ b/pkg/cmd/playback.go
@@ -151,6 +151,7 @@ func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
 		startListenCmdLoop(pc.mode, addressString, httpWrapper)
 	}
 
+	// pass the signal to the server.ListenAndServe somehow
 	ctx := context.Background()
 	withSIGTERMCancel(ctx, httpWrapper)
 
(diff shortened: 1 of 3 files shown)


suz-stripe:
just checking in here - does anything else need to be updated in this PR after #542 is merged?

pepin-stripe:
> just checking in here - does anything else need to be updated in this PR after #542 is merged?

Yeah, I filed something on JIRA, basically the way the logic is now, when playback runs listen in the background, we get 2 `WithSIGTERMCancel` setup and conflicting, so we need to cleanup the cobra command to extract the actual listen logic out of it so playback can run listen without it setting up a sigtermcancel.

See DX-6158 on JIRA for me details.

suz-stripe:
ok I'll consider this as still blocked then thanks for explaining 🙇 
ping me when it's ready for final review or maybe I'll end up taking this on depending on logistics 😄

pepin-stripe:
sounds good! That's probably a great onboarding task for listen and playback since it's mostly cleaning up and reorganizing code around!

suz-stripe:
> That's probably a great onboarding task 
roger that! 🙋🏼‍♀️

alejacdd (approved):

PR #541, opened 2020-11-18
return nil instead of log.Fatal when ctx is closed during Listen and Tailer Run

### LINKED ISSUE #539: CLI says FATAL when ctrl-c

pepin-stripe:
### Reviewers
r? @suz-stripe @richardm-stripe
cc @stripe/developer-products

 ### Summary

Fixes #539 

Commit a023543c63cd2719ae3d980ca2173934653773fc introduced `Log.Fatalf` when `ctx.Done()` happens, which logs a FATAL line to the user terminal when they press CTRL+C.

It is not necessary to log FATAL here since closing the process is the intended purpose by the user when pressing CTRL+C

### DIFF
--- a/pkg/logtailing/tailer.go
+++ b/pkg/logtailing/tailer.go
@@ -174,7 +174,7 @@ func (t *Tailer) Run(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			ansi.StopSpinner(s, "", t.cfg.Log.Out)
-			t.cfg.Log.Fatalf("Aborting")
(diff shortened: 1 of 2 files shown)


richardm-stripe:
LGTM

richardm-stripe (approved):

PR #549, opened 2020-12-03
add post-installation hint to log in

suz-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/developer-products

 ### Summary
We received some feedback about the post installation experience of the CLI. Once installed, it's quite common to jump right into running commands with it. Most commands require a user to be logged in and authenticated with Stripe first, so in this scenario they would immediately see an error asking them to log in. That's a bit of a bummer experience right after installing, so it was suggested that we add a post install greeting to remind users to log in first before anything else.

This homebrew log example demonstrates the greeting added as a postinstall step (6th line from the bottom):

```
$: brew install stripe
==> Installing stripe from stripe/stripe-cli
==> Downloading https://github.com/stripe/stripe-cli/releases/download/v1.5.5/stripe_1.5.5_mac-os_x86_64.tar.gz
Already downloaded: /Users/redacted/Library/Caches/Homebrew/downloads/872c5db6f7bd57b70308810fb007edbefb23144cbb98ef4c4580625683a3b6c1--stripe_1.5.5_mac-os_x86_64.tar.gz
==> /usr/local/Cellar/stripe/1.5.5/bin/stripe completion --shell bash
==> /usr/local/Cellar/stripe/1.5.5/bin/stripe completion --shell zsh
==> Postinstalling stripe
❤ Thanks for installing the Stripe CLI! To get started, run `stripe login`
==> Caveats
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d
==> Summary
🍺  /usr/local/Cellar/stripe/1.5.5: 6 files, 20.4MB, built in 5 seconds
```

Along with homebrew, APT and RPM configs also have this greeting added as a postinstall step.

 ### Limitations
1. GoReleaser's Scoop implementation is not complete, so right now we do not have a way of running a postinstall script on Windows. [I have opened a PR](https://github.com/goreleaser/goreleaser/pull/1931) on GoReleaser for this. If that is merged at some point, we can add this greeting to Scoop installations in the future. [Here is what it might look like](https://github.com/suz-stripe/scoop-test/runs/1490247874?check_suite_focus=true#step:3:25), which is better than brew's IMO.
2. I thought of placing the greeting as a brew caveat (as caveats comes after the postflight stanza), however GoReleaser only accepts strings for the caveats attribute. We'd want an execution block instead as we check if the user is logged in before deciding to show the greeting.
3. I'm a little worried that users still won't see this instruction because it won't be the very last log line of output for any of the package manage installation processes. I'm interested to hear some creative ideas on this! We _could_ run `stripe login` during postinstall to halt and make them notice, but that feels a bit aggressive and not idiomatic.
4. It is awfully hard to test this stuff manually 🤔   any pointers?

### TODO 
+ [x] Prompt the user to log in immediately after running a command needing credentials and then optionally continue the command

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -106,9 +106,7 @@ brews:
           if [[ -f $e ]]; then source $e; fi
         }
       EOS
-    post_install: |
-     system bin/"stripe", "postinstall"
-     EOS
+    caveats: "❤ Thanks for installing the Stripe CLI! If this is your first time using the CLI, be sure to run `stripe login` first."
 scoop:
   bucket:
     owner: stripe
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -79,6 +79,8 @@ nfpms:
     formats:
     - deb
     - rpm
+    scripts:
+      postinstall: "scripts/postinstall.sh"
 brews:
   -
     github:
@@ -104,6 +106,9 @@ brews:
           if [[ -f $e ]]; then source $e; fi
         }
       EOS
+    post_install: |
+     system bin/"stripe", "postinstall"
+     EOS
 scoop:
   bucket:
     owner: stripe
--- a/scripts/postinstall.sh
+++ b/scripts/postinstall.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+stripe postinstall
+
--- a/pkg/cmd/postinstall.go
+++ b/pkg/cmd/postinstall.go
@@ -40,7 +40,7 @@ func (pic *postinstallCmd) runPostinstallCmd(cmd *cobra.Command, args []string) 
 	if err != nil {
 		welcomeIcon := color.BrightRed("❤").String()
 		welcomeText := "Thanks for installing the Stripe CLI! To get started, run `stripe login`"
-		fmt.Println(fmt.Sprintf("%s %s", welcomeIcon, welcomeText))
+		fmt.Printf("%s %s\n", welcomeIcon, welcomeText)
 	}
 
 	return nil
--- a/pkg/cmd/postinstall.go
+++ b/pkg/cmd/postinstall.go
@@ -1,0 +1,47 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/ansi"
+	"github.com/stripe/stripe-cli/pkg/config"
+	"github.com/stripe/stripe-cli/pkg/validators"
+)
+
+type postinstallCmd struct {
+	cfg *config.Config
+	cmd *cobra.Command
+}
+
+func newPostinstallCmd(config *config.Config) *postinstallCmd {
+	pic := &postinstallCmd{
+		cfg: config,
+	}
+	pic.cmd = &cobra.Command{
+		Use:     "postinstall",
+		Args:    validators.NoArgs,
+		Short:   "Run some checks after installation of this CLI and prompt user if needed",
+		Example: `stripe postinstall`,
+		Hidden:  true,
+		RunE:    pic.runPostinstallCmd,
+	}
+	return pic
+}
+
+func (pic *postinstallCmd) runPostinstallCmd(cmd *cobra.Command, args []string) error {
+	color := ansi.Color(os.Stdout)
+	_, err := pic.cfg.Profile.GetAPIKey(false)
+
+	// If we can't get the API key, then it's likely that this is a first install rather than an upgrade.
+	// Suggest the user run `stripe login` to get started as a helpful prompt.
+	if err != nil {
+		welcomeIcon := color.BrightRed("❤").String()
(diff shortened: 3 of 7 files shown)


scripts/postinstall.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+stripe postinstall
```
suz-stripe, line 3: will this binary be found if running under just `sh`? I forget if `$PATH` is a local var for shell scripts.

tomer-stripe:
Does using the caveats place the message at the end of the install output? It might be okay that it's a plain string if we say something like "Thanks for installing ... If this is your first using the cli, make sure to run stripe login first ...".

I'm wondering if it would make sense to turn what you have into a `stripe doctor` style command that can actually check if you've logged in (and maybe work as a wrapper for other commands), check API health, etc (but obviously overall this would be a separate task)

suz-stripe:
> Does using the caveats place the message at the end of the install output? 

It places it at the [absolute lowest point you have access to](https://github.com/Homebrew/homebrew-cask/tree/master/doc/cask_language_reference#stanza-order) as a developer. There will be at least one line afterwards for the brew build statistics. I think also the bash / zsh completion caveats _might_ appear after the login caveat as well, but I'll need to test that theory to be sure. I like your idea of just having a "first time?" qualifier to solve the static string limitation. I'll give this a go.

> I'm wondering if it would make sense to turn what you have into a stripe doctor style command t

I also had the same thought train for a later iteration of this 😄

Jauny:
Hello Stripe Team!

I agree with the risk of users not noticing the message, but at least I don't think it hurts to have it. 

Maybe as an additional nice UX, `stripe login` could be automatically run if a user tries to run a command while logged out?
Still not ideal, but at least they wouldn't have to run the command themselves.

suz-stripe:
Ok here is the brew output since switching to using a caveat: [brew install output](https://github.com/suz-stripe/cli-pkg-mngr-testing/runs/1497088541?check_suite_focus=true#step:2:14)

And here is the APT output, which is much more noticeable: [apt install output](https://github.com/suz-stripe/cli-pkg-mngr-testing/runs/1497088558?check_suite_focus=true#step:2:43)

scripts/postinstall.sh
suz-stripe: this was a non-issue

.goreleaser.yml
```diff
@@ -79,6 +79,8 @@ nfpms:
     formats:
     - deb
     - rpm
+    scripts:
+      postinstall: "scripts/postinstall.sh"
```
tomer-stripe, line 83: Do we still need this with the caveat? Is there anyway we can get this to print out plain text? I'm wondering if we can avoid shipping a command that just prints text

scripts/postinstall.sh
tomer-stripe: Instead of running a command, could we have this echo something instead?

tomer-stripe:
@Jauny that's an interesting idea but I'm worried that'll introduce a bit of a weird side-effect since login requires the user to hit enter and then opens up the browser. That might be annoying if someone's running in docker.

pepin-stripe:
> @Jauny that's an interesting idea but I'm worried that'll introduce a bit of a weird side-effect since login requires the user to hit enter and then opens up the browser. That might be annoying if someone's running in docker.

got it. What happens if someone is running in docker and they are not logged in tho?

.goreleaser.yml
suz-stripe: we still need something in the form of a script file for APT/RPM installs. Each package manager requires a different way of running something postinstall

scripts/postinstall.sh
suz-stripe: we can, but I made that a command for the following reasons:

BLUF: leveraging the CLI itself to provide a better user experience / less maintenance burden for us

+ the CLI is already written and built to be a cross platform binary, so leveraging the existing features we needed for this feature in the codebase was convenient for us retain this compatibility reliably when postinstalling on each platform
  - I wanted to use colour on the heart in order to make the output more noticeable and pretty for users, while also respecting users' colour settings, which the CLI has functionality to do. An echo alone couldn't do this so we'd be writing custom scripts for each platform to do this.
  - I did want some smart login checking behaviour so that we didn't have to output a string for folks who were just upgrading their version so they don't have more noise to read through once done.
+ the postinstall command allows us to keep the above mentioned  functionality in one place, rather than writing both bash and powershell / cmd scripts and having to maintain these different postinstall behaviours as a result.

suz-stripe:
> got it. What happens if someone is running in docker and they are not logged in tho?

if we run the login command for the user in docker, it'll swap to interactive login as the browser popup will gracefully fail due to the cli not being able to source a browser within its container environment

scripts/postinstall.sh
suz-stripe: if we remove the coloured heart feature and smart login checking, we'd be able to get away with the echo, and duplicating the same message in the brew caveat as well.

tomer-stripe (approved):

PR #564, opened 2021-01-19
Adding ability to show URL when no browser is detected on linux

### LINKED ISSUE #499: Login fails if xdg-open is installed but fails to open a web browser
### Issue

When running the stripe cli in an environment such as a Github or Visual Studio Codespace, the following can happen to cause the tool to fail to start a web browser but also not print the URL for manual access:

- although remote, it is not an ssh session; SSH_TTY etc are not set, so the client attempts to open a web browser by choosing the `else` path: https://github.com/stripe/stripe-cli/blob/master/pkg/login/client_login.go#L58
- open.Browser uses Start() to spawn xdg-open: https://github.com/stripe/stripe-cli/blob/master/pkg/open/open.go#L17 and notably Start does not wait for it to complete (vs. Run, or doing a subsequent Wait)
- since the user had installed a package which caused xdg-utils to be installed in the codespace, xdg-open was found so Start did not return an error immediately
- xdg-open fails to start a web browser
- stripe client assumes it was successful, as xdg-open started, and polls for the completion of the remote work

### Expected Behavior

Note that xdg-open failed to start a web-browser and print the URL.

### Steps to reproduce

Install xdg-open but no web browsers and run `stripe login`. 

Alternatively based on https://linux.die.net/man/1/xdg-open we might fake out `xdg-open` returning 3 to indicate no browser was found and test it like this, where the URL should be printed but isn't.

```bash
$ cat ./xdg-open
#!/bin/bash
exit 3 

$ chmod +x ./xdg-open
$ PATH=. ./stripe login
Your pairing code is: scenic-poised-gains-swift
Press Enter to open the browser (^C to quit)
⣟ Waiting for confirmation... 
⣷ Waiting for confirmation... 
⣻ Waiting for confirmation... 
```

### Environment

Linux (Ubuntu 20.04 LTS), stripe version 1.5.1

ctrudeau-stripe:
### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products

 ### Summary

When a user runs `stripe login` a browser is needed. Not all environments have one available. This attempts to detect if the environment lacks a browser in linux. If it lacks a browser then show the URL to the user for them to login. 

### Testing

Built the CLI and copied to a docker container running ubuntu without a browser. URL displayed properly. 

Attempts to fix #499 and RUN_DX-1

### DIFF
--- a/pkg/open/open.go
+++ b/pkg/open/open.go
@@ -29,3 +29,24 @@ func Browser(url string) error {
 
 	return nil
 }
+
+// LacksBrowser determines if no browser is set in linux
+func LacksBrowser() bool {
+	if runtime.GOOS != "linux" {
+		return false
+	}
+
+	var err error
+
+	output, err := execCommand("xdg-settings", "get", "default-web-browser").Output()
+
+	if err != nil {
+		return true
+	}
+
+	if string(output) == "" {
+		return true
+	}
+
+	return false
+}
--- a/pkg/open/open.go
+++ b/pkg/open/open.go
@@ -30,21 +30,21 @@ func Browser(url string) error {
 	return nil
 }
 
-// LacksBrowser determines if no browser is set in linux
-func LacksBrowser() bool {
-	if runtime.GOOS != "linux" {
-		return false
+// CanOpenBrowser determines if no browser is set in linux
+func CanOpenBrowser() bool {
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
+		return true
 	}
 
 	output, err := execCommand("xdg-settings", "get", "default-web-browser").Output()
 
 	if err != nil {
-		return true
+		return false
 	}
 
 	if string(output) == "" {
-		return true
+		return false
 	}
 
-	return false
+	return true
 }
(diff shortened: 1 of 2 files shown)


pepin-stripe (approved):
lgtm!

pkg/open/open.go
```diff
@@ -29,3 +29,22 @@ func Browser(url string) error {
 
 	return nil
 }
+
+// LacksBrowser determines if no browser is set in linux
+func LacksBrowser() bool {
```
tjfontaine-stripe, line 34: I think `Lacks` is a bit difficult to parse the context of this API -- Instead I would consider a `CanOpenBrowser` or `HasDefaultBrowser` a bit more straight forward [but then the return value would be different].

pkg/open/open.go
```diff
@@ -29,3 +29,22 @@ func Browser(url string) error {
...
+	if runtime.GOOS != "linux" {
```
tjfontaine-stripe, line 35: also checking only `linux` here may be a bit too restrictive as this would technically work on other platforms that have `xdg-utils` installed as well, could just flip this to only attempt to try if it's not `darwin` or `windows`

ctrudeau-stripe:
ptal @pepin-stripe, addressing language issues from @tjfontaine-stripe

pepin-stripe (approved):
looks good!

PR #568, opened 2021-01-20
move WithSIGTERMCancel from Proby pkg to cobra command

pepin-stripe:
### Reviewers
r? @suz-stripe 
cc @stripe/developer-products

 ### Summary
This avoids printing "FATAL [...]" when a user closes stripe playback with cmd+c.

It is also a first step in term of decoupling cobra cmd and their core functionality, making it easier and cleaner to have commands triggering multiple core modules (for example stripe playback running listen in the background).

after
Screen Shot 2021-01-20 at 11 52 40 AM

before
Screen Shot 2021-01-20 at 11 52 10 AM

### DIFF
--- a/pkg/cmd/listen.go
+++ b/pkg/cmd/listen.go
@@ -5,8 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"os"
+	"os/signal"
 	"strconv"
 	"strings"
+	"syscall"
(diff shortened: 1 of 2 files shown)


pepin-stripe:
ptal @suz-stripe :)

Z2Xsoft (approved):

PR #574, opened 2021-01-21
Extract core proxy logic outside of cobra listen cmd

pepin-stripe:
### Reviewers
r? @suz-stripe 
cc @tomer-stripe 

 ### Summary
Part of DX-6158 we want to decouple the CLI interface and core logic. This does that for the listen cmd and extract all core logic back into proxy.go.

I'd like to hear early feedback about this and see if it makes sense to everyone.


pkg/proxy/proxy.go
```diff
@@ -396,6 +397,93 @@ func New(cfg *Config, events []string) *Proxy {
...
+					CheckRedirect: func(req *http.Request, via []*http.Request) error {
+						return http.ErrUseLastResponse
+					},
+					Timeout: defaultTimeout,
+					Transport: &http.Transport{
+						TLSClientConfig: &tls.Config{InsecureSkipVerify: cfg.SkipVerify},
+					},
+				},
+				Log:             p.cfg.Log,
+				ResponseHandler: EndpointResponseHandlerFunc(p.processEndpointResponse),
```
vcheung-stripe, line 479: To stream events with gRPC, I would want to have a different ResponseHandler. Maybe a field of `ResponseHandler EndpointResponseHandlerFunc` could be added to the `Config` struct?

pepin-stripe: Yeah we need to update that. We need to be able to both continue to send the webhook response to stripe and at the same time stream that resp to gRPC.

Inside `processEndpointResponse` we already check `if p.webSocketClient != nil {` to send the msg through the websocket, so maybe just adding the gRPC stream handler on `p` so we can also check `if p.gRPCService` or something?

There is also only 1 `EndpointResponseHandlerFunc` type used in the CLI, and it's this occurrence, so I'm not sure why we have it dynamic this way. There's probably a better way of doing this.

vcheung-stripe: I'm ok with having something like `if p.gRPCService ` inside of `processEndpointResponse`. If we go with this approach, then I can add that field when I implement the gRPC method. Then this LGTM.

pepin-stripe: Seems good enough for me too.

vcheung-stripe (commented):
This is really great and I'm glad the visitor logic was reusable! I feel like this will make `proxy` a lot easier to reuse in the code. I just have some comments mainly about the types and style.

pkg/cmd/logs/tail.go
```diff
@@ -244,55 +245,57 @@ func (tailCmd *TailCmd) convertArgs() error {
...
 				ansi.StopSpinner(s, "Ready! You're now waiting to receive API request logs (^C to quit)", logger.Out)
-			case logtailing.Done:
+			case visitor.Done:
 				ansi.StopSpinner(s, "", logger.Out)
 			}
 			return nil
 		},
-		VisitLog: func(le logTailing.LogElement) error {
+		VisitLog: func(le visitor.LogElement) error {
+			log, _ := le.Log.(logtailing.EventPayload)
```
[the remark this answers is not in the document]
pepin-stripe: agreed!

pkg/cmd/listen.go
```diff
@@ -96,181 +101,146 @@ Stripe account.`,
...
+				ansi.StartSpinner(s, "Session expired, reconnecting...", logger.Out)
+			case visitor.Ready:
+				ansi.StopSpinner(s, fmt.Sprintf("Ready! Your webhook signing secret is %s (^C to quit)", ansi.Bold(se.Data[0])), logger.Out)
+			case visitor.Done:
+				ansi.StopSpinner(s, "", logger.Out)
+			}
+			return nil
+		},
+		VisitLog: func(le visitor.LogElement) error {
+			stripeEvent, _ := le.Log.(proxy.StripeEvent)
```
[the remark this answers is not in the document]
pepin-stripe: agreed!

pkg/visitor/visitor.go
```diff
@@ -22,14 +22,15 @@ type ErrorElement struct {
...
-	Log EventPayload
+	Log interface{}
 
 	MarshalledLog string
 }
 
 // StateElement is the current state of the stream: loading, ready, etc.
 type StateElement struct {
 	State state
+	Data  []string
```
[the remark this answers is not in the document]
pepin-stripe: I kinda hacked this one, but I feel like it's probably good enough for our 1 use case, and can be updated when we need.

pkg/visitor/visitor.go
```diff
@@ -1,4 +1,4 @@
-package logtailing
+package visitor
```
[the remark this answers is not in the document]
pepin-stripe: I moved it to its own package because it feels like this would be the de-facto way to communicate between the caller and the core logic? But I guess it all could live in websocket for now indeed.

vcheung-stripe (approved):
LGTM, thanks for making the changes!

pkg/cmd/listen.go
```diff
@@ -96,181 +101,149 @@ Stripe account.`,
...
 
-	return fmt.Sprintf(
-		"%s://%s%s%s",
-		f.Scheme,
-		f.Host,
-		strings.TrimSuffix(f.Path, "/"), // avoids having a double "//"
-		destination.Path,
-	)
+			if strings.ToUpper(format) == outputFormatJSON || printJSON {
+				fmt.Println(de.Marshaled)
```
tomer-stripe, line 228: Do we want this to print the line or go to the logger?

pkg/proxy/proxy.go
```diff
@@ -45,18 +49,29 @@ type EndpointRoute struct {
...
+	// URL to which events are forwarded to
+	ForwardURL string
+	// Headers to inject when forwarding events
+	ForwardHeaders []string
+	// URL to which Connect events are forwarded to
+	ForwardConnectURL string
+	// Headers to inject when forwarding Connect events
+	ForwardConnectHeaders []string
+	// UseConfiguredWebhooks loads webhooks config from user's account
+	UseConfiguredWebhooks bool
```
tomer-stripe, line 66: This is all specific to just webhooks, right? Is this shared with logs tailing?

pkg/proxy/proxy.go
```diff
@@ -387,25 +401,71 @@ func (p *Proxy) processEndpointResponse(evtCtx eventContext, forwardURL string,
...
 		cfg.Log = &log.Logger{Out: ioutil.Discard}
 	}
 
+	// if no events are passed, listen for all events
+	if len(cfg.Events) == 0 {
+		cfg.Events = []string{"*"}
+	} else {
+		for _, event := range cfg.Events {
+			if _, found := validEvents[event]; !found {
+				fmt.Printf("Warning: You're attempting to listen for \"%s\", which isn't a valid event\n", event)
```
tomer-stripe, line 416: Should this go to the logger?

pkg/proxy/proxy.go
```diff
@@ -387,25 +401,71 @@ func (p *Proxy) processEndpointResponse(evtCtx eventContext, forwardURL string,
...
+		}
+	}
+
+	// build endpoint routes
+	var endpointRoutes []EndpointRoute
+	if cfg.UseConfiguredWebhooks {
+		// build from user's API config
+		endpoints := getEndpointsFromAPI(cfg.Key, cfg.APIBaseURL)
+		if len(endpoints.Data) == 0 {
+			// TODO: this log needs to be handled differently
```
tomer-stripe, line 427: What's this todo?

pkg/proxy/proxy.go
pepin-stripe: proxy is only for listen, so specific to webhooks. This has only be moved from the cobra cmd file to the proxy.go file

pkg/cmd/listen.go
pepin-stripe: this should probably go to the logger yes, will fix!

pkg/proxy/proxy.go
pepin-stripe: indeed this one should go to the logger

pkg/proxy/proxy.go
pepin-stripe: cleaned up!

pkg/cmd/listen.go
pepin-stripe: actually, this is the actual events being printed, and using the logger changes the output (timestamp, log level, etc).
We know that some users are probably parsing those logs (especially with json format) and changing the format will probably be a backward incompatible change. Do we want to?

pepin-stripe: cc @tomer-stripe I'd like to hear what you think? I'm not sure what's the implication of using the logger vs fmt here, as it's all getting printed to the terminal from the cobra command for the user. All this logic has mostly been moved, so it's probably safer to keep `fmt` where it's been used?

tomer-stripe: So I'm thinking more broadly for the gRPC work -- we probably wouldn't want that to print anything if they're running in gRPC mode right? (since all the output should go to the gRPC stream itself). A log-level output would be helpful for debugging though!

It's probably a larger scope ask but I think what the CLI actually needs is a separate package that handles all output and it's just automatically done based on how it's being called (gRPC, CLI, debug, etc).

pepin-stripe: Probably makes sense at some point to do some additional cleanup - I think eventually what we want is
- core logic only _returns_ values/errors
- client (cobra, gRPC, etc) processes returns and handles it however it wants (prints, log, etc).

That being said, this section is actually a `cobra` visitor, so it's only ran when listen is ran through the CLI, so we always do want to print here. This loc will never be called when listen is ran through gRPC.

PR #575, opened 2021-01-21
Adding ablity to check for expected failures with a trigger execution

gracegoo-stripe:
### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products

 ### Summary
There are trigger events that expect certain failures to occur such as `charge.failed`. This change allows us to specify an expected error code within a fixture and compare that against the response body returned from the request. If the codes match, we will no longer be treating this as an error. If the request fails for any other reason, the error will propagate as it did before.

After:
Screen Shot 2021-01-22 at 12 42 29 PM
Before:
Screen Shot 2021-01-22 at 12 43 20 PM


pkg/fixtures/fixtures.go
```diff
@@ -105,10 +106,11 @@ func NewFixture(fs afero.Fs, apiKey, stripeAccount, baseURL, file string) (*Fixt
 // defined to populate the user's account
 func (fxt *Fixture) Execute() error {
 	for _, data := range fxt.fixture.Fixtures {
+		fmt.Println()
```
pepin-stripe, line 109: is that needed?

pkg/fixtures/fixtures.go
```diff
@@ -118,6 +120,21 @@ func (fxt *Fixture) Execute() error {
 	return nil
 }
 
+func errWasExpected(err error, expectedErrorType string) bool {
+	if rerr, ok := err.(requests.RequestError); ok {
+		if e, ok := rerr.Body["error"].(map[string]interface{}); ok {
+			if t, ok := e["type"]; ok && t == expectedErrorType {
```
pepin-stripe, line 126: I wonder if there is a way to simplying this instead of having to next 3 levels of ifs just to deconstruct our error. This diff introduces `RequestError` for its purpose so we can probably make it in a way to makes it easier.

pkg/fixtures/fixtures.go
```diff
@@ -118,6 +120,21 @@ func (fxt *Fixture) Execute() error {
...
+				fmt.Printf("Expected Failure occurred. StatusCode: %d, Error Type: %s\n", rerr.StatusCode, t)
+				if message, ok := e["message"]; ok {
+					fmt.Printf("Error Message: %s\n", message)
```
pepin-stripe, line 129: `errWasExpected` should probably just return bool and not print anything.

pkg/requests/base.go
```diff
@@ -39,6 +40,21 @@ func (r *RequestParameters) SetStripeAccount(value string) {
...
+type RequestError struct {
+	msg        string
+	StatusCode int
+	Body       map[string]interface{}
+}
+
+func (e RequestError) Error() string {
+	out, err := json.MarshalIndent(e.Body, "", "  ")
+	if err != nil {
+		panic(err)
```
pepin-stripe, line 53: for safety maybe we should instead still return a string saying something like "unexpected error trying to marshal error's json body" or something.

pepin-stripe (commented):
Looks good overall, a few comments.

Also in terms of UX, triggering a "success" event looks like this:
```
$ stripe trigger charge.succeeded
Setting up fixture for: charge
```

So I think triggering a failure event should look the same. For the user, triggering the event should also look like a success (the event was successfully triggered) and printing that a failure was registered will be confusing. IMO we should print the same thing for `charge.success` and `charge.failed` since the _command_ was successful in either case.

pkg/fixtures/fixtures.go
pepin-stripe: so I got a little excited to try and figure out how to do that in Go, here's an idea of how to clean that up a bit.

1. We create structs to represent the expected _error_ response from the API.
```
// this could also be RequestBody and have fields for both error and success responses but not sure this is needed
type RequestErrorBody struct {
    Error RequestError `json:"error"`
}

type RequestError struct {
    Message string `json:"message"`
    Code string `json:"code"`
}
```

2. Now we can Marshal `body` into a `RequestErrorBody` and have all field natively accessible
```
var reqErrorBody RequestErrorBody
json.Unmarshal(body, &reqErrorBody)
```
now we can check if reqErrorBody has the correct code
```
if err != nil && !err.Error.Code != data.ExpectedErrorType {
  return err
}
```

gracegoo-stripe:
> So I think triggering a failure event should look the same. For the user, triggering the event should also look like a success (the event was successfully triggered) and printing that a failure was registered will be confusing. IMO we should print the same thing for `charge.success` and `charge.failed` since the _command_ was successful in either case.

That's fair, I'll update this in the next rev!

pkg/fixtures/fixtures.go
gracegoo-stripe: It was added to make it a bit easier to differentiate the outputs for the events that run multiple fixtures. However, with a simpler output this will not be needed anymore!

pkg/fixtures/fixtures.go
gracegoo-stripe: Thanks for demoing this to us, I'll incorporate it into my next change!

pkg/requests/base.go
gracegoo-stripe: Sounds good. In that case, is the goal to print an error message for this particular request but still allow the event to continue executing?

pepin-stripe: I think we should avoid `panic` in general and gracefully handle errors.

In that particular example, we're trying to stringify an error, so if json.MarshalIndent fails I would just return a simpler string, such as 
```
fmt.Printf("Request failed, %v", err)
```

pepin-stripe (approved):
Yes that looks good! I like the idea of keeping `requestErrorBody` and `requestErrorContent` scoped inside `compileRequestError`!

PR #587, opened 2021-02-02
Fix fixture logic which parsed large numbers into scientific notation

ianjabour-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary
When we unmarshal fixtures from a file we represent `params` as an empty interface. This means that numeric values are parsed as `float64`.

In our logic which translates the values to request params we used the `%v` for float which is translated to `%g` under the hood. The `%g` format will use:
> %e for large exponents, %f otherwise

For more context reference the [docs](https://golang.org/pkg/fmt/) and the comment that was added.

### Testing
 
I was able to reproduce the failure in a test case and verify that it is now resolved.

Screen Shot 2021-02-02 at 12 59 02 PM

### DIFF
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


suz-stripe (approved):
@ianjabour-stripe  LGTM
thank you for a) the test b) the new documentation 💯 A+

PR #589, opened 2021-02-08
Implement triggers and fixtures with custom data

tomelm:
### Reviewers
r? @suz-stripe @pepin-stripe @gracegoo-stripe 
cc @stripe/developer-products @ob-stripe @tjfontaine-stripe 

 ### Summary
So, I was watching the Super Bowl and for reasons I cannot entirely explain was hit with some inspiration for #295. We'd discussed wanting to have dynamic flags (like `--customer-email fry@stripe.com`) but that's not going to doable with the compilation of the CLI and shell autocompletion (plus, I think it would get pretty ambiguous quickly for some more advanced trigger files).

Instead what I think we should consider is a few generic flags that support the "query style" paths we currently support in the fixtures themselves to use params across multiple requests.

The interface for users here would expose four additional flags on `stripe fixtures` and `stripe triggers`:
1. `--skip <name>` - let users skip entire fixture steps based on the name. Example: `--skip cust_bender --skip char_bender`
2. `--override <query>` - override currently specified params. Example: `--override customer:email=fry_stripe.com`
3. `--add <query>` - add completely new parameters that are not currently defined in the fixture. Example: `--add customer:robot.type="Bending robot"` (which would convert to json of `{robot: {type: "Bending robot"}}`)
4. `--remove <query>` - remove parameters defined in the fixture

The gist of how I'm implementing this is by building maps of structure:
`customer:foo.bar.baz=abc_123`
```
{
  <name:customer>: {
    <path:foo>: {
      <path:bar>: {
        <path:baz>: "abc_123"
      }
    }
  }
}
```

These maps are then merged against the existing params dictionary. Removals still need to be implemented and will have to work differently.

### Testing
Some manual testing:
```
➜ stripe-dev trigger customer.created --skip customer
Skipping fixture for: customer

➜ stripe-dev trigger charge.succeeded --override charge:amount=50
Running fixture for: charge
```

Added automated tests too!

### To do
If this seems like a good idea, then:
- [x] implement additions
- [x] implement removals
- [x] clean up the code
- [x] automated testing

### For a future iteration
- [ ] print statements to acknowledge changed parameters
- [ ] `--save` flag to write what resulting fixture/trigger would look like
- [ ] some form of debug mechanisms, maybe some print/read flags to more cleanly explain what's happening.


pepin-stripe:
Looks interesting I like how it's handled and the dot interface seems good.

I'm wondering how skipping fixtures works; I thought that creating a customer was necessary to have an id to trigger an update. How is that happening when we skip that?

Doolattaj:
I think this jumping is good

tomer-stripe:
@pepin-stripe for skipping, you'd need to pair it with an override. With how fixtures currently work, if we take an update scenario it's roughly:
1. create customer
2. use id from step 1 to update an object

If you already have an object though, you could skip the first step entirely and use an override:
```
$ stripe trigger customer.updated --skip customer_create --override customer:id=foo
```

tomelm:
Thinking about this a little more, I'm thinking I should clean up this file/code a bit as I implement the feature. There's a lot of weird recursive parsing logic and I want to make sure we're exposing good interfaces to build on this in the future without making it brittle 🤔

pkg/fixtures/fixtures.go
```diff
@@ -95,27 +93,79 @@ func NewFixture(fs afero.Fs, apiKey, stripeAccount, baseURL, file string) (*Fixt
...
+	for _, f := range fxt.fixture.Fixtures {
+		if _, ok := data[f.Name]; ok {
+			if err := mergo.Merge(&f.Params, data[f.Name]); err != nil {
+				fmt.Println(err)
+			}
+		}
+	}
+}
+
+// Remove todo implement - removes fields from the fixture
```
vcheung-stripe, line 137: Is this still todo? It looks complete to me

vcheung-stripe (approved):
LGTM, I like the interface for Fixture and test coverage looks good. 🚢

pkg/fixtures/fixtures.go
```diff
@@ -497,33 +275,81 @@ func (fxt *Fixture) updateEnv(env map[string]string) error {
...
 
-	if r, didMatch := matchFixtureQuery(value); didMatch {
-		isQuery = true
-		match := r.FindStringSubmatch(value)
-		query = fixtureQuery{Name: match[1], Query: match[2], DefaultValue: match[3]}
+	return false
+}
+
+// buildRewrites takes a slice of json queries and values then builds
+// them into a map to later by merged. We work through the entire
```
pepin-stripe, line 290: I think you meant "to later _be_ merged" ?

pepin-stripe (approved):
LGTM overall. Only concern/question (not blocking) I have is if it's going to be confusing for users that `--skip` needs to go with `--override` and the type of error message a user will get if they `--skip` with no override. Will it be clear or generic?

pkg/fixtures/fixtures.go
```diff
@@ -497,33 +275,81 @@ func (fxt *Fixture) updateEnv(env map[string]string) error {
...
-	// separator for `name:json_path`. Additionally, default value will
-	// be specified after the `|`.
-	// example: ${name:json_path|default_value}
-	r := regexp.MustCompile(`\${([^\|}]+):([^\|}]+)\|?([^/\n]+)?}`)
-	if r.Match([]byte(value)) {
-		return r, true
+// pop returns the first item and the rest of the list minus the first item
+// From: https://github.com/golang/go/wiki/SliceTricks#pop
+func pop(list []string) (string, []string) {
+	return list[len(list)-1], list[:len(list)-1]
```
etsai-stripe, line 340: @tomer-stripe i think this pops the last item, and return the list minus the last item? should we update the comment or code?

etsai-stripe: @tomer-stripe also if we do want to pop the last item, do we still need to `reverse` it at line314?

tomer-stripe:
Closing this PR since @etsai-stripe finished it in a separate one!

PR #594, opened 2021-02-18
run scripts in bash for double bracket compat

suz-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary
Our scripts were breaking due to a `sh` env incompatibility with double bracket conditionals on GH Actions. I changed the shebangs to `bash` for the affected scripts in order to fix this.


vcheung-stripe (approved):
The code change LGTM, but I don't have context on the environment our action runs in.

PR #598, opened 2021-02-21
[do not merge] Update to go1.16

tomelm:
### Reviewers
r? @suz-stripe @pepin-stripe @gracegoo-stripe 
cc @stripe/developer-products

 ### Summary
Update to go1.16. This also updates dependencies and linter.

~Do not merge until https://github.com/Homebrew/homebrew-core/pull/71289 is resolved, otherwise we won't be able to do local dev.~

Homebrew  has updated to go 1.16 so we can merge once we feel good about the release steps working.

PR #599, opened 2021-02-21
Update to go1.16 and split releaser to separate builds on Linux, macOS, and Windows

tomelm:
### Reviewers
r? @suz-stripe 
cc @stripe/developer-products @pepin-stripe @gracegoo-stripe 

 ### Summary
I can't figure out how to only diff on my other PR so this contains some of the same commits as #598. Do not merge until #598 is mergeable. 

Before we can use the `embed` feature in go1.16 we need to make sure our build processes compile against 1.16 instead of 1.13.x.

We've had to use the mailchain goreleaser Docker image to make sure that macOS compilation works correctly (#439). Unfortunately, that's based on [dockercore/golang-cross](https://github.com/docker/golang-cross) which hasn't updated past go1.13 and doesn't seem like they're going to in the near future (see https://github.com/docker/golang-cross/pull/54 and https://github.com/docker/golang-cross/issues/64).

This PR splits our build and release into two flows:
1. for macOS
2. for everything else

Building explicitly for macOS will make sure we don't have the DNS-related issues in #439, but also makes it easier for us to build for Apple Silicon (including building a universal binary, which requires [`lipo`](https://ss64.com/osx/lipo.html)) and signing/notarizing the binary (which requires some macOS-specific tooling from Xcode). I'm not sure if we should just move everything to macOS at the risk of introducing potential other compatibility issues for Linux or Windows, but having split workflows doesn't seem like a bad idea at the moment.

Plus, overall this will reduce our build dependency chain on third parties, giving us more control.

edit: in case it's not obvious I just got one of the M1 laptops and have been playing around with the ARM compilation 😆


themar7777 (approved):

themar7777 (changes requested):
Windows/remove

tomelm:
ptal @suz-stripe 
cc @tjfontaine-stripe @pepin-stripe

I updated this PR over the other (#598). I started working to pull the Dockerfile in and as I got deeper into it, there were some errors occurring around installing `docker-ce`. There are workarounds (see [this](https://unix.stackexchange.com/questions/363048/unable-to-locate-package-docker-ce-on-a-64bit-ubuntu)) but I was getting increasingly worried about how many work arounds we were introducing, it felt like it was growing very brittle.

I figure split build processes might be the least risky option long term. I've been testing it in another repo ([here](https://github.com/tomelm/stripe-cli-temp)) and the release ["worked"](https://github.com/tomelm/stripe-cli-temp/actions/runs/628379424) (minus a few failures; the `mac-build` failure was because it tried to upload to the stripe homebrew repo, the windows issue looks like it was rate limited). The release itself actually worked and only created one released for the version tag ([see v0.0.10](https://github.com/tomelm/stripe-cli-temp/releases/tag/v0.0.10)).

I did split this out to `mac`, `windows`, and `linux` builds. Instead of one checksums file, I created one for each platform (there were overwrite issues). `arm` actually _does not_ work yet because the macOS 11.0 environment is private at the moment (see https://github.com/actions/virtual-environments/issues/2486).

.goreleaser.linux.yml
```diff
@@ -0,0 +1,55 @@
...
+    ldflags:
+      - -s -w -X github.com/stripe/stripe-cli/pkg/version.Version={{.Version}}
+    binary: stripe
+    env:
+      - CGO_ENABLED=1
+      - CC=x86_64-w64-mingw32-gcc
+      - CXX=x86_64-w64-mingw32-g++
+    main: ./cmd/stripe/main.go
+    goos:
+      - windows
```
pepin-stripe, line 30: is that supposed to be windows since this is the config file for the linux build?

.github/workflows/release.yml
```diff
@@ -1,42 +1,67 @@
...
-      -
-        name: Upload to Bintray
-        run: |
-          scripts/publish-deb-to-bintray.sh
-          scripts/publish-rpm-to-bintray.sh
-        shell: bash
+          go-version: 1.16
+      -
+        name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
```
pepin-stripe, line 22: seems like this is all using the correct go v of 1.16 _but_ I thought we were supposed to use goreleaser-xcgo to allow cgo compilation for some mac-related issues (dns related I think)? Did you confirm that we can now use goreleaser directly and everything works? (@suz-stripe has more context)

pepin-stripe (changes requested):

.goreleaser.linux.yml
tomer-stripe: Nope, need to remove this, good catch!

.github/workflows/release.yml
tomer-stripe: The `xcgo` setup was needed predominantly because we were compiling a mac-native binary on Linux, which required some different tooling. If we run the mac-native compilation on mac-* platforms, it should be fine moving forward!

pepin-stripe (approved):
lgtm!

tomelm:
This is ready! Most recent test run is here: https://github.com/tomelm/stripe-cli-temp/actions/runs/634468413. Note that the test run doesn't include some of the follow up steps such as Homebrew, Docker, Bintray, or VirusTotal. I can't easily test those but none of those steps or code changed here so should be okay!

.github/workflows/release.yml
```diff
@@ -1,30 +1,49 @@
...
     tags:
-      - 'v*'
+      - "v*"
 
 jobs:
-  goreleaser:
+  build-mac:
+    runs-on: macos-latest
+    steps:
+      - name: Code checkout
```
suz-stripe, line 12: we're missing a `with: fetch-depth: 0` in this step

.github/workflows/release.yml
```diff
@@ -33,11 +52,26 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GORELEASER_GITHUB_TOKEN }}
           BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
           BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
-      -
-        name: Upload to VirusTotal
+
+  build-windows:
+    runs-on: windows-latest
+    steps:
+      - name: Code checkout
```
suz-stripe, line 59: we're missing a `with: fetch-depth: 0` in this step

suz-stripe (approved):
This looks great and is a step in the right direction long term for our release process.

I initially had concerns around what would happen if three different goreleaser release runs would cause issues for the GitHub release step with regards to the tag, the changelog and the archives attached but looking at your practice run repo it looks like it seems to work fine. I am delighted but also confused as to how goreleaser handles this so gracefully. I'm guessing it checks for an existing tag / published release and just pushes the builds to an existing release if it finds one? Perhaps GitHub also smoothes this over with idempotence in their API.

As for the Bintray / Virustotal steps, I assume that the GitHub release archives needed should be available as long as these upload steps are run in the right os context.

Going forward, I recommend that we think about building dry run abilities into our scripts so that we can easily test our scripts that are run outside of goreleaser. For example, our bintray and virustotal scripts could check for a dry run env var or flag, and run all steps except for uploading. This way we can ensure the other important parts of the script work such as ensuring env vars are correctly supplied, the script has the right permissions and dependencies installed, and that the GitHub release download URLs resolve correctly to files so we can spot issues with our goreleaser config changes.

This dry run could also extend to running the release workflow against a temp repo for cli / brew / scoop etc. Just some thoughts! 💭

PR #611, opened 2021-03-14
Use golang's embed feature for triggers

tomelm:
### Reviewers
r? @stripe/developer-products
cc @stripe/developer-products

 ### Summary
Use golang's embed feature and remove unused dependencies: https://golang.org/pkg/embed/

Tested manually locally:
```
stripe-cli on  tomer/embed-fixtures via 🐹 v1.16.1
➜ stripe-dev trigger customer.created
Setting up fixture for: customer
```


pepin-stripe (approved):
wow this is amazing!

I'm wondering how that works- With the files at `pkg/fixtures/`, are they "packaged" in the binary twice now? (once as _files_ and once in memory as embedded?)

I was expecting those files to be moved somewhere where the compilation wouldn't be adding those files in the binary but here it seems like they might be still included? Or Golang is smart enough to not package what was embedded?

tomer-stripe:
The go compiler has always generally ignored non-go files, which is why we needed to have additional steps to convert strings into binary data. With this feature the compiler handles this natively and behind the scenes

PR #614, opened 2021-03-16
auto-create "required resources" on stripe sample create

richardm-stripe:
### Reviewers
r? @stripe/developer-products

 ### Summary
Paired on this with @cjavilla-stripe, who requested this feature.

Today `stripe samples create` very conveniently will copy over a user's secret key/publishable key/webhook signing secret into the `.env` file of the Stripe Sample.

For some samples, this is all the configuration the sample needs, and it's ready to go! But some samples -- often billing-related -- have additional dependencies. They require the user to create a resource on their account in testmode, and add the corresponding id to the `.env` as well. At best, this is inconvenient. At worst, this is confusing, since it is not always very obvious what kind of resource you will need, or what API call you will need to make to create it -- especially for users who are just trying out Stripe for the first time.

This PR begins to solve that problem by introducing the concept of a "required resource" which can be automatically created and configured during `stripe samples create`.

Here's how it works:
1. Each stripe sample defines their "requiredResources" inside `.cli.json` ([example](https://github.com/stripe-samples/checkout-single-subscription/blob/master/.cli.json#L4-L13)).
2. A hardcoded lookup table inside the stripe-cli code defines a POST request capable of generating that type of resource ([code](https://github.com/stripe/stripe-cli/pull/614/files#diff-1775bc38562f60c6dfb1d49400580f915200dd7cb40866e3552bf19181cb7f0bR313-R336)).
3. When the user runs `stripe samples create <name>`, if the stripe sample defines any resources that have not already been created, they receive a prompt asking permission to automatically create them.
image
4. If they select yes, then a POST request is made to testmode underneath the currently active profile. If the post request is successful, the value of the "id" field in the response is written to the user's profile in their stripe-cli config (so that it can be re-used on future invocations of `stripe samples create`)
5. When the `.env` file is copied from the example, any of the required resource IDs present in the user's profile are written as environment variables there as well (like is already happening for secret key/publishable key)

Notes:
- There's a hardcoded list here in stripe-cli. If somebody wants to add a new type of "required resource", they will need to add an entry to that list.
- If the user deletes one of these auto-created resources but doesn't clear it from their stripe-cli config, they won't see the prompt anymore and `stripe samples create` will copy over an id that doesn't exist and so the sample will not work.

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -13,15 +13,16 @@ import (
 
 // Profile handles all things related to managing the project specific configurations
 type Profile struct {
-	DeviceName             string
-	ProfileName            string
-	APIKey                 string
-	LiveModeAPIKey         string
-	LiveModePublishableKey string
-	TestModeAPIKey         string
-	TestModePublishableKey string
-	TerminalPOSDeviceID    string
-	DisplayName            string
+	DeviceName              string
+	ProfileName             string
+	APIKey                  string
+	LiveModeAPIKey          string
+	LiveModePublishableKey  string
+	TestModeAPIKey          string
+	TestModePublishableKey  string
+	TerminalPOSDeviceID     string
+	DisplayName             string
+	StripeSampleResourceIDs map[string]string
 }
 
 // CreateProfile creates a profile when logging in
@@ -140,6 +141,25 @@ func (p *Profile) GetDisplayName() string {
 	return ""
 }
 
+// GetStripeSampleResourceID gets the id from the config for a Stripe Sample "required resource"
+// with a particular name.
+func (p *Profile) GetStripeSampleResourceID(resourceName string) string {
+	if err := viper.ReadInConfig(); err == nil {
+		return viper.GetString(p.GetConfigField(resourceName))
+	}
+
+	return ""
+}
+
+// SetStripeSampleResourceID sets the id in the config for a Stripe Sample "required resource"
+// with a particular name.
+func (p *Profile) SetStripeSampleResourceID(resourceName string, id string) {
+	if p.StripeSampleResourceIDs == nil {
+		p.StripeSampleResourceIDs = map[string]string{}
+	}
+	p.StripeSampleResourceIDs[resourceName] = id
+}
+
 // GetTerminalPOSDeviceID returns the device id from the config for Terminal quickstart to use
 func (p *Profile) GetTerminalPOSDeviceID() string {
 	if err := viper.ReadInConfig(); err == nil {
@@ -206,6 +226,10 @@ func (p *Profile) writeProfile(runtimeViper *viper.Viper) error {
 
 	if p.DisplayName != "" {
 		runtimeViper.Set(p.GetConfigField("display_name"), strings.TrimSpace(p.DisplayName))
+	}
+
+	for name, id := range p.StripeSampleResourceIDs {
+		runtimeViper.Set(p.GetConfigField(name), strings.TrimSpace(id))
 	}
 
 	runtimeViper.MergeInConfig()
--- a/pkg/config/profile_test.go
+++ b/pkg/config/profile_test.go
@@ -18,6 +18,9 @@ func TestWriteProfile(t *testing.T) {
 		ProfileName:    "tests",
 		TestModeAPIKey: "sk_test_123",
 		DisplayName:    "test-account-display-name",
+		StripeSampleResourceIDs: map[string]string{
+			"stripe_samples_price_recurring_basic_id": "pr_xyz",
+		},
 	}
 
 	c := &Config{
@@ -42,6 +45,7 @@ func TestWriteProfile(t *testing.T) {
 [tests]
   device_name = "st-testing"
   display_name = "test-account-display-name"
+  stripe_samples_price_recurring_basic_id = "pr_xyz"
   test_mode_api_key = "sk_test_123"
 `
 	require.EqualValues(t, expectedConfig, string(configValues))
--- a/pkg/samples/list.go
+++ b/pkg/samples/list.go
@@ -108,7 +108,7 @@ func (s *Samples) GetSamples(mode string) map[string]*SampleData {
 		return list
 	}
 
-	// Reduce the amount of request to GitHub
+	// Reduce the number of requests to GitHub
 	var noNetwork bool
 	switch mode {
 	case "list":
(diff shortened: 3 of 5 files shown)


gracegoo-stripe (approved):
Nice this looks awesome! I checked out the PR and played around with it and it's much nicer than the previous experience. 

Just some comments/questions regarding the notes you mentioned:
1. Is there an internal guide for creating stripe samples that we can update so it's less likely the sample authors forget? And I suppose also to advertise the feature to the authors. 
2. I'm not sure what the best approach here is -- how the error is surfaced to the users depends on how each sample handles the missing resource, right? Maybe this is also something we can add on the documentation so the samples make it clear to the user that they have to delete a stale ID in their CLI config.

richardm-stripe:
Thanks for the review @gracegoo-stripe, I'm letting @cjavilla-stripe adopt this.

Re: stale IDs:
  1. We could just not cache the IDs in the user's profile. This would potentially create a lot of spammy products though if they have to clone multiple times.
  2. We could *always* prompt the user for permission. If they say "yes" and there are no IDs in their profile, then we create. If there *are* IDs in their profile, we could do a GET first to confirm it still exists, and then a POST only if it hadn't been deleted.

tomer-stripe:
@richardm-stripe @cjavilla-stripe I was actually thinking about this over the weekend! I'm a little concerned about hardcoding the resources like this though and have a slightly different suggestions: what if we have a `fixtures` folder for each sample and the CLI could detect and execute those on clone? This interface means that changing data would require a new CLI release and is going to make it hard for the user to run this on their own.

cjavilla-stripe:
@tomer-stripe we chatted briefly about using a rough fixtures approach where the fixtures methods would be defined in the sample .cli.json, but very curious to hear any ideas you may have about how we could get the specific output of a fixture into .env.

richardm-stripe:
@tomer-stripe I'm a bit worried about doing `git clone` and letting the contents have explicit control over HTTP requests sent to Stripe API with the users' credentials. We do hardcode testmode (for now), so if you think this is overkill, I'm happy to defer

Engineering-wise, there are some advantages to having the smarts inside the CLI, I think. In this PR, resolving a "required resource" means just a path and some form data (for a POST), if we ever wanted to slightly depart from this (make a GET, not a POST;  make a series of requests, not just one; fetch a property from the response other than "id", etc.), it would be very easy to write Go code that just does something different for a particular name. If the smarts live in the `.cli.json`, then we will need a small JSON DSL for expressing these behaviors, which could be annoying.

OTOH this might not be worse than the release burden, WDYT?

richardm-stripe:
Maybe even simpler than all this would be to put all the smarts into the sample repos and include a Bash setup script in every Stripe sample that runs CLI commands and can munge `.env`. Probably not secure to trigger it automatically, but `stripe samples create` could tell the user about it.

tomer-stripe:
@cjavilla-stripe we would definitely need some changes to fixtures. The command interface is pretty basic but if we leverage the underlying fixture code within the CLI (instead of invoking the command), we would have access to all the request and response data we store: https://github.com/stripe/stripe-cli/blob/master/pkg/fixtures/fixtures.go#L57. We'd still need a little bit of work but if we load the json file into a `Fixture`, then run `#Execute`, we could add a function that takes a JSON query to pull specific data out.

@richardm-stripe I think that's a fair concern! We'd want to think through the interface a bit. A combination of the fixture being explicitly called out and a prompt in the CLI stating something like "We detected some fixtures in this sample, would you like to run the following commands to bootstrap the data?" could help. Plus, also, strong +1 to only testmode only.

pepin-stripe:
Well I'm glad I went through this conversation because I didn't know much about fixtures and now I do!

I kind of agree with the idea of leveraging fixtures, because it seems like it's doing exactly what we need (list of requests we need to do to create/pull resources, potentially with dependencies on each others). 

I also think that a _sample_ should be responsible for _its own dependencies_, else we create a dependency between the 2 systems.

To address @richardm-stripe concerns - is your concern about a malicious fixture json file to be inserted into the cloned repo, despite it being a stripe-owned repo? Is that a real possibility (as in, more than a malicious fixture inserter into the CLI directly)?

pepin-stripe:
Another question about storing the IDs locally and risk of stale ids. Is it necessary to store ids to know if a resource exist? Can we GET the most recent resource instead and ask prompt the user to run the POST to create one if none exist on the test account?

Richard already mentioned doing a GET to validate IDs, so maybe we don't have to store IDs at all?

pepin-stripe:
yeah agreed with that!

cjavilla-stripe:
Thanks for the feedback, all. 

Down to brass tacks :)

> We'd still need a little bit of work but if we load the json file into a Fixture, then run #Execute, we could add a function that takes a JSON query to pull specific data out.

Wondering if we would want some structure in [.cli.json](https://github.com/stripe-samples/accept-a-card-payment/blob/master/.cli.json) that specified the fixture command and json query to set an env variable like:

```json
{
    "name": "accept-a-card-payment",
    "configureDotEnv": true,
    "integrations": [
      {
        "name": "using-webhooks",
        ".env": {
          "PRICE_ID": {
            "fixture": "./create-price.json",
            "query": ".id"
          },
        },
        "clients": ["web", "android", "ios"],
        "servers": ["java", "node", "node-typescript", "php-slim", "php", "python", "ruby"]
      },
      {
        "name": "without-webhooks",
        "clients": ["web", "android", "ios"],
        "servers": ["java", "node", "node-typescript", "php-slim", "php", "python", "ruby"]
      },
      {
        "name": "decline-on-card-authentication",
        "clients": ["web"],
        "servers": ["java", "node", "node-typescript", "php-slim", "php", "python", "ruby"]
      }
    ]
  }
```

Then a file in the root of the sample or in like `/fixtures/create-price.json` in the sample that looked like:

```json
{
  "_meta": {
    "template_version": 0
  },
  "fixtures": [
    {
      "name": "price",
      "path": "/v1/prices",
      "method": "post",
      "params": {
        "currency": "usd",
        "unit_amount": 1999,
        "product_data": {
          "name": "Stripe Sample Product",
          "description": "(created for Stripe Sample)"
        },
      }
    }
  ]
}
```

salimchakht (commented):
okay

richardm-stripe:
Closing -- can reopen if either of us has time to try the suggested approach

giulianalembo (commented):
Ha trovato prodotti su stripe

PR #615, opened 2021-03-17
Add format flag to listen cmd

gracegoo-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/developer-products

 ### Summary
The --format flag is a string that currently only takes in JSON as a valid input option, similar to the log tailing command.
It's intended to replace the --print-json for the listen cmd. The difference in behavior between --format JSON and --print-json is that the new flag will stripe out all formatting in the payload. 

While working on the event streaming feature in VSCode, the stream was getting partial json because the json was being pretty-printed. This PR updates the listen cmd to behave the same way as the log tail command. 

### Testing
With the deprecated flag:
Screen Shot 2021-03-17 at 9 54 54 AM

New flag:
Screen Shot 2021-03-17 at 9 52 56 AM

No flag:
Screen Shot 2021-03-17 at 9 54 12 AM

### DIFF
--- a/pkg/cmd/listen.go
+++ b/pkg/cmd/listen.go
@@ -66,7 +66,7 @@ Stripe account.`,
 	lc.cmd.Flags().BoolVarP(&lc.latestAPIVersion, "latest", "l", false, "Receive events formatted with the latest API version (default: your account's default API version)")
 	lc.cmd.Flags().BoolVar(&lc.livemode, "live", false, "Receive live events (default: test)")
 	lc.cmd.Flags().BoolVarP(&lc.printJSON, "print-json", "j", false, "Print full JSON objects to stdout.")
-	lc.cmd.Flags().MarkDeprecated("print-json", "Please use `--format JSON` instead")
+	lc.cmd.Flags().MarkDeprecated("print-json", "Please use `--format JSON` instead and use `jq` if you need to process the JSON in the terminal.")
 	lc.cmd.Flags().StringVar(&lc.format, "format", "", `Specifies the output format of webhook events
 	Acceptable values:
 		'JSON' - Output webhook events in JSON format`)
--- a/pkg/cmd/listen.go
+++ b/pkg/cmd/listen.go
@@ -66,10 +66,10 @@ Stripe account.`,
 	lc.cmd.Flags().BoolVarP(&lc.latestAPIVersion, "latest", "l", false, "Receive events formatted with the latest API version (default: your account's default API version)")
 	lc.cmd.Flags().BoolVar(&lc.livemode, "live", false, "Receive live events (default: test)")
 	lc.cmd.Flags().BoolVarP(&lc.printJSON, "print-json", "j", false, "Print full JSON objects to stdout.")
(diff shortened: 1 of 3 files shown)


pkg/cmd/listen.go
```diff
@@ -64,7 +65,11 @@ Stripe account.`,
 	lc.cmd.Flags().StringVarP(&lc.forwardConnectURL, "forward-connect-to", "c", "", "The URL to forward Connect webhook events to (default: same as normal events)")
 	lc.cmd.Flags().BoolVarP(&lc.latestAPIVersion, "latest", "l", false, "Receive events formatted with the latest API version (default: your account's default API version)")
 	lc.cmd.Flags().BoolVar(&lc.livemode, "live", false, "Receive live events (default: test)")
-	lc.cmd.Flags().BoolVarP(&lc.printJSON, "print-json", "j", false, "Print full JSON objects to stdout")
+	lc.cmd.Flags().BoolVarP(&lc.printJSON, "print-json", "j", false, "Print full JSON objects to stdout.")
+	lc.cmd.Flags().MarkDeprecated("print-json", " This flag is deprecated. Please use `--format JSON` instead")
```
vcheung-stripe, line 69: Looks like the message `Flag --print-json is deprecated` is already printed.

pkg/cmd/listen.go
```diff
@@ -64,7 +65,11 @@ Stripe account.`,
...
+	lc.cmd.Flags().StringVar(&lc.format, "format", "", `Specifies the output format of request logs
```
vcheung-stripe, line 70: I think this might want to be
```diff
- Specifies the output format of request logs
+ Specifies the output format of events
```

vcheung-stripe (changes requested):
I have some suggestions for the output strings 😄

gracegoo-stripe:
> I have some suggestions for the output strings 😄

Oh my goodness, thank you for catching these!

vcheung-stripe (commented):
The messages look good to me. I can't speak for the rest of it so passing back r? @tomer-stripe

pepin-stripe:
LGTM but same questions as @tomer-stripe for case sensitivity. As a user I think I'd default to `--format json` since I'm usually used to pass in arguments in lower case

gracegoo-stripe:
> This looks good! Two quick questions:
> 
> 1. is `JSON` case sensitive? It probably shouldn't be but I can't remember what we do in request logs
> 2. did you test this with piping into `jq`? We should make sure that still works (and potentially mention that in the deprecation message too)

1. They are not but I can update them!
2. I did just now and that does work. I can suggest this in the deprecation message as well, that's a good idea. 
 
Screen Shot 2021-03-18 at 9 39 05 AM

gracegoo-stripe:
Okay, I updated the flag check to be case insensitive for both `listen` and `logs tail`. Also added colorized output to the `listen` output because that looked fun:

Screen Shot 2021-03-18 at 11 55 58 AM

Screen Shot 2021-03-18 at 11 53 28 AM

And the updated deprecation message:

Screen Shot 2021-03-18 at 11 59 00 AM

pepin-stripe (approved):
LOVE IT!

vcheung-stripe (approved):

Issue #616, opened 2021-03-22
Fixtures: Incorrect parameter list is sent to the query


pleshevskiy:
I created custom fixture, because I wanted to test checkout session with subscription, but something seems to have gone wrong.

### Issue
The following code is my fixture. I have `plan_EhGOJoLHVbjdhS` with 10$ /month and `plan_EhG7axLZKYDylg` with 100$ /month. 100 + 10 = 110$ = 11000 cents. 

```
{
    "_meta": {
        "template_version": 0
    },
    "fixtures": [
        {
            "name": "checkout_session",
            "path": "/v1/checkout/sessions",
            "method": "post",
            "params": {
                "success_url": "https://httpbin.org/post",
                "cancel_url": "https://httpbin.org/post",
                "mode": "subscription",
                "payment_method_types": [
                    "card"
                ],
                "line_items": [
                    {
                        "price": "plan_EhGOJoLHVbjdhS",
                        "quantity": 1
                    },
                    {
                        "price": "plan_EhG7axLZKYDylg",
                        "quantity": 1
                    }
                ]
            }
        },
        {
            "name": "payment_page",
            "path": "/v1/payment_pages",
            "method": "get",
            "params": {
                "session_id": "${checkout_session:id}"
            }
        },
        {
            "name": "payment_method",
            "path": "/v1/payment_methods",
            "method": "post",
            "params": {
                "type": "card",
                "card": {
                    "token": "tok_visa"
                },
                "billing_details": {
                    "email": "stripe@example.com"
                }
            }
        },
        {
            "name": "payment_page_confirm",
            "path": "/v1/payment_pages/${payment_page:id}/confirm",
            "method": "post",
            "params": {
                "payment_method": "${payment_method:id}",
                "expected_amount": 11000
            }
        }
    ]
}
```

After running the command I get an error

```
Request failed, status=400, body={
  "error": {
    "code": "checkout_amount_mismatch",
    "message": "The computed invoice amount does not match the latest invoice on the subscription.",
    "param": "expected_amount",
    "type": "invalid_request_error"
  }
}
```

When I change `expected_amount` to 10000 all works but... in dashboard I see only one paid plan (`plan_EhG7axLZKYDylg`). When I switch `line_items` and change `expected_amount` to 1000 it steel works. In dashboard I see another one paid plan (`plan_EhGOJoLHVbjdhS`)

---

When I create checkout session via curl (see code bellow) all works correct

```bash
[...]

PR #617, opened 2021-03-22
Fixing index management for params of type arrays of hashes in fixtures

gracegoo-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary
Addresses https://github.com/stripe/stripe-cli/issues/616

Arrays of hashes in request params are differentiated by their index number. Our fixture param parser was not incrementing the index for hash elements within an array, which resulted in the hashes getting overwritten if there were more than one of them in the array. This is why the expected amount by the server was different from what the user expected in that issue. \ This change adds in logic to increment the index counter as we iterate through an array of hashes. 

It also removes an unnecessary parameter in parseArray. I think it's unnecessary because we'd want to restart the counter within the scope of a single array of hashes. I tried to find a real life example of this to verify but wasn't able to find an API ref that has something like an array of an array of hashes.

### Testing
Tested with a custom fixture file. The same one provided in this [issue](https://github.com/stripe/stripe-cli/issues/616), with the plans replaced with my own. 
Before this change: 
Screen Shot 2021-03-22 at 4 29 00 PM

After this change:
Screen Shot 2021-03-22 at 4 30 59 PM

Curl request for comparison: 
Screen Shot 2021-03-22 at 4 32 04 PM

Also added logic to our unit test to verify array of hash types.


vcheung-stripe (commented):
This LGTM since we paired on this, but I think more eyes on this would be great because I don't know what all the API request shapes look like. ptal @gracegoo-stripe

pkg/fixtures/fixtures.go
```diff
@@ -274,22 +274,23 @@ func (fxt *Fixture) parseMap(params map[string]interface{}, parent string, index
 	return data
 }
 
-func (fxt *Fixture) parseArray(params []interface{}, parent string, index int) []string {
+func (fxt *Fixture) parseArray(params []interface{}, parent string) []string {
```
pepin-stripe, line 277: I think a docstring for this function would help in the future (should have been already here).
Reading the function gets a bit confusing if someone is trying to figure out what shape can be expected for `params` and `fmt.Sprintf("%s[]=%s", parent, fxt.parseQuery(v.String())))` is pretty confusing to parse as a human.

gracegoo-stripe: Good callout, I'll try to add some more documentation here!

gracegoo-stripe: Updated, ptal @pepin-stripe

pepin-stripe: That's much better! 

There is a typo (`elemenent`) and I think it'd be a useful to show an example of possible input since the way url params are passed (`"address[city]=New New York"`, `"tax_id_data[1][value]=value_1"`, etc) can be pretty confusing.

not a blocker tho!

gracegoo-stripe: Agreed, added those examples. ptal @pepin-stripe

pepin-stripe (approved):

pkg/fixtures/fixtures.go
pepin-stripe: looks great now thanks :)

PR #618, opened 2021-03-26
Unsupported event type: Show more detailed error message

Fonata:
### Reviewers
r? @tomer-stripe
cc @stripe/developer-products

### Summary
I executed the following command in my Bash:
```bash
stripe trigger payout.failed
```

The command failed with the output "event payout.failed is not supported."

I was not sure if the message came from the Stripe servers (e.g. no endpoint configured for event) or my own app server. Only after researching that neither is probable, I decided to search the Stripe CLI source. I'd like to save other developers some time.


pepin-stripe (approved):
This is looks great, thanks a lot!

PR #638, opened 2021-04-21
Add gRPC server with `Version` method, available through `stripe daemon`

vcheung-stripe:
### Reviewers
cc @stripe/developer-products 

 ### Summary

Note that this is merging into a grpc branch, not master.

Add everything needed for a gRPC server, along with the `Version` method to verify it works. I'm happy to break this up if it's too large.

- Added a daemon command to start the gRPC server
Screen Shot 2021-04-21 at 2 20 18 PM
Screen Shot 2021-04-21 at 2 07 50 PM
- Implemented `Version` as the first method to verify it works.
- Added simple auth: accept only requests that have a `sec-x-stripe-cli` header
- Added a make target `make protoc` to compile proto files and generate docs. Example doc:
Screen Shot 2021-04-21 at 2 08 21 PM
- Updated GH action to validate that the proto files were compiled

### Testing
- Added test for `Version`
- Added tests for the middleware checking for the `sec-x-stripe-cli` header
- Locally invoked `Version` from a client in vscode-stripe for manual testing


Makefile
```diff
@@ -30,7 +30,7 @@ cover: test
 
 # gofmt and goimports all go files
 fmt:
-	find . -name '*.go' | while read -r file; do gofmt -w -s "$$file"; goimports -w "$$file"; done
+	find . -path ./rpc -prune -false -o -name '*.go' | while read -r file; do gofmt -w -s "$$file"; goimports -w "$$file"; done
```
vcheung-stripe, line 33: Don't format generated grpc code in `./rpc`.

pkg/rpcserver/middleware.go
```diff
@@ -0,0 +1,53 @@
...
+const requiredHeader = "sec-x-stripe-cli"
+
+// Only allow requests from clients with a sec-x-stripe-cli header
+func authorize(ctx context.Context) error {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return status.Errorf(codes.InvalidArgument, "Retrieving metadata failed")
+	}
+
+	if _, ok := md[requiredHeader]; !ok {
```
pepin-stripe, line 22: how does that header work? will we ever care about the header value and need to validate, or do we just want the header's presence? In which case, why is this necessary?

pkg/rpcserver/rpc_server.go
```diff
@@ -0,0 +1,127 @@
...
+		<-interruptCh
+		onCancel()
+		cancel()
+	}()
+	return ctx
+}
+
+// Run starts the gRPC server
+func (srv *RPCServer) Run(ctx context.Context) error {
+	ctx = withSIGTERMCancel(ctx, func() {
```
pepin-stripe, line 68: I think we need to think about this a bit more. That's how we handle cmd+c right now in other functions, but it's creating weird race conditions when 2 different commands are calling each others (such as Playback internally running Listen).

`withSIGTERMCancel` is purely a command-line specific UX and I'm wondering if this should be the one, single bit of additional logic we could allow in the cobra command.

pkg/rpcserver/rpc_server.go
```diff
@@ -0,0 +1,127 @@
...
+
+	fmt.Fprintln(os.Stderr, string(configOutput))
+
+	go func() {
+		if err := grpcServer.Serve(lis); err != nil {
+			srv.cfg.Log.Fatalf("Failed to serve gRPC server %s: %v", lis.Addr().String(), err)
+		}
+	}()
+
+	for range ctx.Done() {
```
pepin-stripe, line 123: interesting. Is it better to do that vs `<- ctx.Done()`?

pkg/rpcserver/rpc_server.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+		srv.cfg.Log.Fatalf("Failed to listen on %s. Unexpected error: %v", address, err)
+	}
+
+	grpcServer := grpc.NewServer(
+		grpc.UnaryInterceptor(serverUnaryInterceptor),
+		grpc.StreamInterceptor(serverStreamInterceptor),
+	)
+
+	rpc.RegisterStripeCLIServer(grpcServer, srv)
```
pepin-stripe, line 106: I'm wondering if we can avoid the double server naming (`grpcServer` and `RPCServer`) which gets pretty confusing especially when you try to debug or talk to someone, and use another naming instead?

Since the command is called `daemon` could we replace `RPCServer` by `RPCDaemon`?

pepin-stripe: Or `RPCService`? It seems like the autogen for `RegisterStripeCLIServer` is calling `RegisterService`

pkg/rpcserver/middleware.go
vcheung-stripe: It's necessary to prevent malicious websites from making requests to the user's daemon. Their browser won't allow adding a header with prefix `sec-`, but other clients like vscode-stripe can. I can dm you with the details if you'd like.

pkg/rpcserver/rpc_server.go
vcheung-stripe: Agree, also we should probably DRY it

pkg/rpcserver/rpc_server.go
vcheung-stripe: Honestly I'm not sure, I'm not familiar enough with channels 😅

pkg/rpcserver/middleware.go
pepin-stripe: Playing devil's advocate: Are we concerned about other malicious clients?

Please DM details, I'd love to read more on that!

pkg/rpcserver/rpc_server.go
vcheung-stripe: Agree -- it was very confusing to me even to write this

pkg/rpcserver/rpc_server.go
pepin-stripe: If your goal here is to block until `ctx.Done()` is being closed, I think you can simply do `<-ctx.Done()`. There are a few examples in other commands that use the same pattern.

pkg/rpcservice/rpc_service.go
```diff
@@ -0,0 +1,111 @@
...
+
+	rpc.RegisterStripeCLIServer(srv.grpcServer, srv)
+
+	if err := srv.grpcServer.Serve(lis); err != nil {
+		srv.cfg.Log.Fatalf("Failed to serve gRPC server on %s: %v", lis.Addr().String(), err)
+	}
+}
+
+func (srv *RPCService) createListener() net.Listener {
+	address := "127.0.0.1:" // empty port => an available port is automatically chosen
```
tjfontaine-stripe, line 75: Since you use `net.Listen("tcp"` later, I think we should use `IPv6loopback` here instead of `127.0.0.1` (side point, it makes me sad that golang has a constant for IPv6 loopback but not IPv4)

pkg/rpcservice/rpc_service.go
```diff
@@ -0,0 +1,111 @@
...
+			srv.cfg.Log.Fatalf("Failed to listen on %s. %s is an invalid port.", address, addrError.Addr)
+		}
+
+		// Port is already in use
+		var syscallErr *os.SyscallError
+		if errors.As(err, &syscallErr) && errors.Is(syscallErr.Err, syscall.EADDRINUSE) {
+			srv.cfg.Log.Fatalf("Failed to listen on %s. Port is already in use.", address)
+		}
+
+		srv.cfg.Log.Fatalf("Failed to listen on %s. Unexpected error: %v", address, err)
```
tjfontaine-stripe, line 100: The only other error I can think of at the moment would be "no free ports", but this should be fine as is.

pkg/rpcservice/version.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+
+	"github.com/stripe/stripe-cli/pkg/version"
+	"github.com/stripe/stripe-cli/rpc"
+)
+
+// Version returns the version of the Stripe CLI
+func (srv *RPCService) Version(ctx context.Context, req *rpc.VersionRequest) (*rpc.VersionResponse, error) {
+	return &rpc.VersionResponse{Version: version.Version}, nil
```
tjfontaine-stripe, line 12: you might want to add things like the git commit sha here

tjfontaine-stripe (approved):

pkg/rpcservice/rpc_service.go
vcheung-stripe: Updated with `IPv6loopback` and printing the host and port in separate fields:

Screen Shot 2021-04-28 at 10 59 55 AM

pkg/rpcservice/version.go
vcheung-stripe: I think I could add this field in a future PR.

PR #646, opened 2021-04-28
Implement `stripe daemon` and gRPC server with initial subset of commands

vcheung-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

This is the main development branch of the stripe daemon feature, which exposes CLI commands in a gRPC server.

TODO:
- [x] Add `stripe daemon` command
- [x] Authenticate requests with a Sec- header
- [x] Generate documentation for protobuf files
- [x] Add CI for protobuf generated files
- [x] stripe version method
- [x] stripe samples methods
- [x] stripe login methods
- [x] stripe listen method
- [x] stripe logs tail method
- [x] stripe triggers method
- [x] stripe events resend method
- [ ] ~stripe status method~ won't do
- [ ] ~Add telemetry for commands invoked in the daemon~ TODO in the near future
- [x] commands should return UNAUTHENTICATED error if logged out
- [ ] ~remove documentation~ TODO after the CLI ref is up
- [x] add logging to the interceptors

This branch currently contains these PRs:
- #638 add `stripe daemon` command, auth, docs generation, CI, and `Version` method
- #647 implement `SamplesList` method
- #648 implement `SampleConfigs` method
- #658 implement `SampleCreate` method
- #655 implement `Login` and `LoginStatus` methods
- #659 implement `TriggersList` method
- #663 implement `Trigger` method
- #667 implement `LogsTail` method
- #669 implement `EventsResend` method
- #676 implement `Listen` method
- #677 return `UNAUTHENTICATED` error code for some commands
- #678 write debug logs when methods are invoked

PR #652, opened 2021-05-03
Refactor stripe samples commands

vcheung-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary

- Hoisted the spinners, user IO, and error handling up to the cobra command
  - This removes IO concerns from the samples package
- Added new fields to the `Samples` struct to make it easier to work with 
- Updated and added a couple tests in the samples package

Verified manually by running stripe samples list and creating a variety of stripe samples with and without --force-refresh, with and without a path.

Motivated by https://github.com/stripe/stripe-cli/pull/646, which will be rebased on this.

### DIFF
--- a/pkg/cmd/samples/create.go
+++ b/pkg/cmd/samples/create.go
@@ -1,6 +1,7 @@
 package samples
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
@@ -113,6 +114,8 @@ func (cc *CreateCmd) runCreateCmd(cmd *cobra.Command, args []string) error {
 			if res.PostInstall != "" {
 				fmt.Println(res.PostInstall)
 			}
+		default:
+			return errors.New("an unknown error occurred during sample creation")
 		}
 	}
 
--- a/pkg/cmd/samples/create.go
+++ b/pkg/cmd/samples/create.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 
+	"github.com/manifoldco/promptui"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
@@ -65,7 +66,12 @@ func (cc *CreateCmd) runCreateCmd(cmd *cobra.Command, args []string) error {
 		Git:    gitpkg.Operations{},
 	}
 
-	if _, ok := sample.GetSamples("create")[args[0]]; !ok {
+	samplesList, err := sample.GetSamples("create")
+	if err != nil {
+		return err
+	}
+
+	if _, ok := samplesList[args[0]]; !ok {
 		errorMessage := fmt.Sprintf(`The sample provided is not currently supported by the CLI: %s
 To see supported samples, run 'stripe samples list'`, args[0])
 		return fmt.Errorf(errorMessage)
@@ -105,7 +111,7 @@ To see supported samples, run 'stripe samples list'`, args[0])
 	// depending on whether or not it's. Additionally, this
 	// identifies if the sample has multiple integrations and what
 	// languages it supports.
-	err := sample.Initialize(selectedSample)
+	err = sample.Initialize(selectedSample)
 	if err != nil {
 		switch e := err.Error(); e {
 		case git.NoErrAlreadyUpToDate.Error():
@@ -130,10 +136,11 @@ To see supported samples, run 'stripe samples list'`, args[0])
 	// directory, the user needs to select which integration they
 	// want to work with (if selectedSamplelicable) and which language they
 	// want to copy
-	err = sample.SelectOptions()
+	selectedConfig, err := PromptSampleConfig(sample)
 	if err != nil {
 		return err
 	}
+	sample.SelectedConfig = *selectedConfig
 
 	// Setup to intercept ctrl+c
 	c := make(chan os.Signal, 1)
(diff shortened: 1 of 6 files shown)


ob-stripe:
I've removed myself from the team. Re-rolling.

r? @stripe/developer-products

pkg/cmd/samples/create.go
```diff
@@ -65,7 +66,12 @@ func (cc *CreateCmd) runCreateCmd(cmd *cobra.Command, args []string) error {
 		Git:    gitpkg.Operations{},
 	}
```
pepin-stripe, line 67: maybe out of scope of this diff, but to me this is an implementation detail that should live inside the package, not called here.

pkg/cmd/samples/create.go
```diff
@@ -130,10 +136,11 @@ To see supported samples, run 'stripe samples list'`, args[0])
 	// directory, the user needs to select which integration they
 	// want to work with (if selectedSamplelicable) and which language they
 	// want to copy
-	err = sample.SelectOptions()
+	selectedConfig, err := PromptSampleConfig(sample)
 	if err != nil {
 		return err
 	}
+	sample.SelectedConfig = *selectedConfig
```
pepin-stripe, line 143: I think this function still has a lot of implementation details that should be moved into the samples package.

Things such as checking if files/dir exists, if the sample is already up to date, etc.

pkg/cmd/samples/create.go
```diff
@@ -181,3 +188,97 @@ To see supported samples, run 'stripe samples list'`, args[0])
...
+	return selectedIntegration, nil
+}
+
+func serverSelectPrompt(servers []string) (string, error) {
+	selected, err := selectOptions("server", "What server would you like to use", servers)
+	if err != nil {
+		return "", err
+	}
+
+	return selected, nil
```
pepin-stripe, line 283: all those are fine here indeed. all the Prompts, Options, etc. (user input) should live here. The Cobra cmd should only have the responsibility to take in user input, and then pass it to the samples package as arguments to a "create" func.

pkg/cmd/samples/list.go
```diff
@@ -27,13 +29,13 @@ func NewListCmd() *ListCmd {
...
 the CLI.`,
-		Run: listCmd.runListCmd,
+		RunE: listCmd.runListCmd,
 	}
 
 	return listCmd
 }
 
-func (lc *ListCmd) runListCmd(cmd *cobra.Command, args []string) {
+func (lc *ListCmd) runListCmd(cmd *cobra.Command, args []string) error {
```
pepin-stripe, line 38: yes this one looks great, except previous comment about having to do `samples.Samples{...}` seems that it's an implementation detail to me (not 100% convinced tho). Especially since we only pass defaults to it...

It should probably just be `samples.GetSamples("list")` without having to init a sample

pepin-stripe (approved):
So... I dont' wanna block gRPC but I feel like this cleanup could be a bit more complete.

It seems like this PR is moving cobra-related logic out of the samples package, which is great and probably enough to unblock gRPC.

What I think would be great to see if also cleaning up package-related logic out of the cobra command (see inline comments for more details).

Basically image calling `Samples.Create` or whatever from another place than the cobra function - all logic that would need to be duplicated from the command is probably good candidates to be extracted into the package.

pkg/cmd/samples/create.go
vcheung-stripe: Agree, a single function that creates the sample and returns (path, error) would be ideal. I think this will make grpc implementation simpler as well.

pkg/cmd/samples/create.go
vcheung-stripe: Do you mean the construction of the Samples object? I can change it to follow this pattern using `New` : https://github.com/stripe/stripe-cli/blob/3fbf545adac182d1374bbe04e5c84a22f0a04851/pkg/logtailing/tailer.go#L94-L107

pepin-stripe: No, more about the fact that you have to initialize a `sample` object, only with default, to then call functions that feel like they could be package function instead of instance functions.

Basically instead of doing
```
sample := samples.Sample{}
sample.CreateSample()
```

it seems like the interface would be cleaner with
```
samples.CreateSample()
```
directly.

pepin-stripe: This is not blocking by the way, mostly just nit/personal perf.

pkg/cmd/samples/create.go
vcheung-stripe: I took a look at this and I think it's in scope because the grpc method will need this logic.

However, a problem with moving out the file existence check, sample is up to date check, etc, is that the command wants to print messages to the terminal while those checks are running. If we move out that logic, we would have to think about where to put the spinner and print statements.

If we want to move the logic to the samples package but keep the spinner and print statements in the cobra command, we could maybe use a channel? The `samples.Create` method could write a status to the channel: Downloading, Copying, Configuring, etc, and the cobra command could read the status and write the spinner and print statements asynchronously. What do you think of this approach?

pepin-stripe: Agreed, I think channels is the way to go here to stream the output.

vcheung-stripe: Sounds good I will add this!

vcheung-stripe:
I split sample creation into two methods: 

- `samples.GetSampleConfig` returns a sample's config
  - The cobra command calls this so it knows what options to prompt the user in the terminal
  - The grpc method can call this to return the available options to a grpc client
- `samples.Create` returns a struct over a channel: current state of the routine, error, destination path, and post install message
  - The cobra command reads from the channel to print spinners and status messages
  - The grpc method doesn't care about the state, but it needs the destination path and post install message to return to the client

I'm not sure how clear this pattern is, but it will make implementing the grpc method simpler because there will be less duplication. ptal @pepin-stripe

pkg/cmd/samples/create.go
```diff
@@ -59,125 +54,161 @@ func (cc *CreateCmd) runCreateCmd(cmd *cobra.Command, args []string) error {
...
+	go samples.Create(
+		cc.cfg,
+		selectedSample,
+		selectedConfig,
+		destination,
+		cc.forceRefresh,
+		resultChan,
+	)
+
+	for res := range resultChan {
```
pepin-stripe, line 94: you might want a `default` case for safety

pepin-stripe (approved):
LGTM! Nice work :)

pkg/cmd/samples/create.go
vcheung-stripe: Added a default case, ptal @pepin-stripe

pepin-stripe: did you push it? I don't see it.

vcheung-stripe: Yep here it is:
https://github.com/stripe/stripe-cli/pull/652/files#diff-54c4064d2b4ba0c610ea0511c0faa5109f448125b2b307a35ee5c53939c0310aR117-R119

pepin-stripe: wtf it didnt appear before I clicked that link! LGTM 🚢

vcheung-stripe:
r-codeowners?

PR #657, opened 2021-05-05
Fix samples create always failing

vcheung-stripe:
### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products

 ### Summary

In https://github.com/stripe/stripe-cli/pull/652, I made a mistake in the switch statement. Because I didn't include all the cases, we would hit the default case every time. The default case returns an error so sample creation would fail every time.

This adds those unused cases to the switch statement and also renames them to be more accurate.


pepin-stripe (approved):
LGTM

PR #658, opened 2021-05-06
Implement `SampleCreate` method for gRPC

vcheung-stripe:
### Reviewers
cc @stripe/developer-products

 ### Summary

Implement `SampleCreate` method for gRPC. This uses the refactored methods from https://github.com/stripe/stripe-cli/pull/652.

Generated documentation:
Screen Shot 2021-05-05 at 5 04 48 PM
Screen Shot 2021-05-05 at 5 04 57 PM

I tested this on an example typescript client:
```ts
  try {
    const sampleCreateRequest = new SampleCreateRequest()
      .setSampleName("accept-a-card-payment")
      .setIntegrationName("using-webhooks")
      .setClient("web")
      .setServer("java")
      .setPath(
        "/Users/vcheung/stripe/test-grpc-client/samples/accept-a-card-payment"
      );
    const response = await new Promise<SampleCreateResponse>(
      (resolve, reject) => {
        client.sampleCreate(sampleCreateRequest, (error, response) => {
          if (error) {
            reject(error);
          } else {
            resolve(response);
          }
        });
      }
    );
    console.log(
      response.getPostInstall() || "Successfully created sample",
      sampleCreateRequest.getSampleName(),
      response.getPath()
    );
  } catch (e) {
    console.error(e);
  }
```

I successfully created a couple samples:
Screen Shot 2021-05-05 at 4 59 19 PM Screen Shot 2021-05-05 at 4 59 08 PM


pkg/rpcservice/sample_create.go
```diff
@@ -0,0 +1,82 @@
...
+	for res := range resultChan {
+		if res.Err != nil {
+			return nil, res.Err
+		}
+		if res.State == samples.Done {
+			return &rpc.SampleCreateResponse{
+				Path:        res.Path,
+				PostInstall: res.PostInstall,
+			}, nil
+		}
```
pepin-stripe, line 42: could there be another state return or something that we care about or need to handle?

pepin-stripe (approved):
LGTM

pkg/rpcservice/sample_create.go
vcheung-stripe: Good question. Definitely the client wants to know about success and failure. I don't think the client should be concerned about other states because they are internal details, like files being copied and stuff like that.

PR #660, opened 2021-05-07
Refactor trigger command

vcheung-stripe:
### Reviewers
r? @gracegoo-stripe
cc @stripe/developer-products

 ### Summary

- Move the logic from the `stripe trigger` cobra command into its own method, `fixtures.Trigger()`. This method will be reused in the `Trigger` gRPC method.
- Make `fixture.Execute()` return the names of the requests made. For example, if you run `stripe trigger customer.created`, then `fixture.Execute()` returns ["customer", "customer_deleted"]. This list is ignored by the cobra command, but the gRPC method response will use it.

This is merging into master. Afterwards I'll rebase https://github.com/stripe/stripe-cli/pull/646 on it and add `Trigger` for grpc.

I ran tests and manually ran stripe trigger with different args and flags to verify nothing changed.

### DIFF
--- a/pkg/cmd/fixtures.go
+++ b/pkg/cmd/fixtures.go
@@ -61,7 +61,7 @@ func (fc *FixturesCmd) runFixturesCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	err = fixture.Execute()
+	_, err = fixture.Execute()
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


gracegoo-stripe (approved):
Looks great!

pepin-stripe (approved):
LGTM!

PR #661, opened 2021-05-07
Refactor logs tail command with channel and visitors

vcheung-stripe:
### Reviewers
r? @pepin-stripe
cc @stripe/developer-products

 ### Summary

This decouples user IO and the receiving of logs.
- The `Tailer` is concerned with sending data on a channel. The data is either a log, an error, a warning, or a connection state.
- The cobra command, as a client of the `Tailer`, receives these different data types and processes them using a visitor object.

This pattern enables `Tailer` to be used by any class that wants to do something with logs, for example a gRPC service that wants to stream logs, errors, and connection state to a gRPC client. All the class has to do is read from the channel and provide its own visitor object.

This pattern might also work well for https://github.com/stripe/stripe-cli/pull/574. This is quite boilerplate heavy, so I am open to other solutions as well.

### DIFF
--- a/pkg/cmd/logs/tail.go
+++ b/pkg/cmd/logs/tail.go
@@ -167,18 +167,18 @@ func (tailCmd *TailCmd) runTailCmd(cmd *cobra.Command, args []string) error {
 
 	logger := log.StandardLogger()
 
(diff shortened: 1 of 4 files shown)


pepin-stripe (approved):
didn't know about the visitor pattern, neat :)

Issue #662, opened 2021-05-07
panic: runtime error: index out of range [1] with length 1


tekord:
### Issue

The following command with a combination of 'forward-to' and 'headers' options produces the runtime error:

```
stripe listen --forward-to http://my-app.local/stripe/webhook --headers ACT-AS-USER=1

panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/stripe/stripe-cli/pkg/proxy.convertToMapAndSanitize(0xc0003d2690, 0x1, 0x1, 0x156af40)
        D:/a/stripe-cli/stripe-cli/pkg/proxy/endpoint.go:190 +0x245
github.com/stripe/stripe-cli/pkg/proxy.NewEndpointClient(0xc00001e4b0, 0x23, 0xc0003d2690, 0x1, 0x1, 0x0, 0xc0001a97d0, 0x1, 0x1, 0xc000242160, ...)
        D:/a/stripe-cli/stripe-cli/pkg/proxy/endpoint.go:152 +0x88
github.com/stripe/stripe-cli/pkg/proxy.New(0xc000554750, 0xc0001a97d0, 0x1, 0x1, 0x2)
        D:/a/stripe-cli/stripe-cli/pkg/proxy/proxy.go:410 +0x3f1
github.com/stripe/stripe-cli/pkg/cmd.(*listenCmd).runListenCmd(0xc000146630, 0xc0001d9400, 0xc0001b4340, 0x0, 0x4, 0x0, 0x0)
        D:/a/stripe-cli/stripe-cli/pkg/cmd/listen.go:171 +0x5a5
github.com/spf13/cobra.(*Command).execute(0xc0001d9400, 0xc0001b4300, 0x4, 0x4, 0xc0001d9400, 0xc0001b4300)
        C:/Users/runneradmin/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:852 +0x472
github.com/spf13/cobra.(*Command).ExecuteC(0x1b40540, 0x70d, 0x60, 0xc000036060)
        C:/Users/runneradmin/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:960 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
        C:/Users/runneradmin/go/pkg/mod/github.com/spf13/cobra@v1.1.3/command.go:897
github.com/stripe/stripe-cli/pkg/cmd.Execute()
[...]

PR #664, opened 2021-05-10
Updating the help docs for the header params in the listen command.

gracegoo-stripe:
### Reviewers
cc @stripe/developer-products
r? @stripe/developer-products 

 ### Summary
A user opened an issue recently because they were getting an index out of bounds error when trying to use the headers param.
They were using the wrong delimiter for separating the Key and Value of the header, updating the docs to make that more clear for the users.

https://github.com/stripe/stripe-cli/issues/662

### Testing

Before:
Screen Shot 2021-05-10 at 4 04 45 PM

After:
Screen Shot 2021-05-10 at 4 04 24 PM

### DIFF
--- a/pkg/cmd/listen.go
+++ b/pkg/cmd/listen.go
@@ -58,10 +58,10 @@ Stripe account.`,
 		RunE: lc.runListenCmd,
 	}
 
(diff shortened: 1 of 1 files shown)


gracegoo-stripe:
> LGTM!
> 
> We should also add them in the CLI ref too. Would you mind doing that as well?
> https://stripe.com/docs/cli/listen#listen-connect-headers

Absolutely!

PR #665, opened 2021-05-11
patch quickstart to use new base_url attr for reader hostname

suz-stripe:
### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products

 ### Summary
The P400 readers have gone through certificate updates for the [new PQDN format](https://stripe.com/docs/terminal/readers/verifone-p400#reader-has-ip-address-and-can-communicate-with-stripe,-but-not-with-your-point-of-sale-application), which adds a random string to the hostname in order to make them fully unique. This was done because using private IPs alone create much higher chances of collisions. The reader firmware now rejects ingress requests the from old hostname format due to the resulting certificate mismatch.

Quickstart was still using the old hostname format and so the feature was broken. I updated it to extract the `base_url` to use in combination with the private IP. 

I manually tested all three flows:
1. Registering new reader, with no existing readers on account
2. Registering new reader, with existing readers on account
3. Choosing an existing registered reader from list

This change was a bit fiddly and required me to move some stuff around / refactor the registration flow, so HMU if you have any questions around this 🙇

### DIFF
--- a/pkg/terminal/p400/rabbit_service.go
+++ b/pkg/terminal/p400/rabbit_service.go
@@ -64,7 +64,7 @@ func CallRabbitService(tsCtx TerminalSessionContext, method string, methodConten
 	formattedIP := strings.Join(strings.Split(tsCtx.IPAddress, "."), "-")
 	t := &Transport{}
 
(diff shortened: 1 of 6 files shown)


pepin-stripe (approved):
not much context but LGTM :)

PR #667, opened 2021-05-12
Implement `LogsTail` gRPC method

vcheung-stripe:
### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products

 ### Summary

Using #661, implement `LogsTail` gRPC method. We can do something similar for `Listen`.

`LogsTail` is a server stream*. The client sends a request that contains filters. The server streams the connection state and log objects. The client can cancel the stream any time.

*Previously, I had this as a bidirectional stream because I wanted the client to be able to send a special message to end the stream. However I didn't like that design (client would be given too much flexibility to send messages), so I changed it to a server stream. In this new design, clients will have to (1) emit a CANCELLED error on the stream and (2) handle a CANCELLED error from the server.

Documentation:
Screen Shot 2021-05-12 at 10 28 24 AM
Screen Shot 2021-05-12 at 10 28 37 AM
Screen Shot 2021-05-12 at 10 28 51 AM

The response is one of two types:
Screen Shot 2021-05-12 at 10 29 02 AM
Screen Shot 2021-05-12 at 10 28 59 AM

An example typescript client:
```ts
    const stream = client.logsTail(new LogsTailRequest());

    stream.on("data", (response: LogsTailResponse) => {
      // handle the two types of responses
      if (response.hasState()) {
        console.log(response.getState());
      } else if (response.hasLog()) {
        console.log(response.getLog());
      }

      // client cancels the stream
      stream.cancel();
    });

    stream.on("error", (err) => {
      // ignore the CANCELLED error because it's expected
      if ((err as grpc.ServiceError).code !== grpc.status.CANCELLED) {
        console.error(err);
      }
    });

    stream.on("end", () => {
      console.log("ended");
    });

```


pkg/rpcservice/logs_tail.go
```diff
@@ -0,0 +1,199 @@
...
+type ITailer interface {
+	Run(context.Context) error
+}
+
+var createTailer = func(cfg *logtailing.Config) ITailer {
+	return logtailing.New(cfg)
+}
+
+// LogsTail returns a stream of API logs
+func (srv *RPCService) LogsTail(req *rpc.LogsTailRequest, stream rpc.StripeCLI_LogsTailServer) error {
```
pepin-stripe, line 22: if `StripeCLI_LogsTailServer` an auto generated name? We don't usually use snake case in Go. `StripeCLILogsTailServer` would be better.

pkg/rpcservice/logs_tail.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+}
+
+func createVisitor(stream *rpc.StripeCLI_LogsTailServer) *logtailing.Visitor {
+	return &logtailing.Visitor{
+		VisitError: func(ee logtailing.ErrorElement) error {
+			return ee.Error
+		},
+		VisitLog: func(le logtailing.LogElement) error {
+			logError := rpc.LogsTailResponse_Log_Error{
```
pepin-stripe, line 77: will `VisitLog` always be called with an error? I'm confused here.

pkg/rpcservice/logs_tail.go
```diff
@@ -0,0 +1,199 @@
...
+				Content: &rpc.LogsTailResponse_State_{
+					State: stateResponse,
+				},
+			})
+			return nil
+		},
+	}
+}
+
+func getFiltersFromReq(req *rpc.LogsTailRequest) *logtailing.LogFilters {
```
pepin-stripe, line 123: that's... annoying!

pkg/rpcservice/logs_tail.go
pepin-stripe: I'm confused at the signature of this function - what's the entry point for clients?

In your tests in go, you do `client.LogsTail(ctx, &rpc.LogsTailRequest{})` and in javascript you pass `client.LogsTail(req)`. 

Seems like the same function is called differently, never with `stream`. Can you clarify?

pepin-stripe (approved):
overall looks good, I just need clarification on the gRPC signature and how client.LogsTail is called differently in the go tests and in the javascript example.

I trust you for it to work, so not blocking.

pkg/rpcservice/logs_tail.go
vcheung-stripe: - Yes, `StripeCLI_LogsTailServer` is autogenerated so I can't change it.  The format is `<service name>_<method name>Server`.
- I can see the confusion. The signatures are different because they are not the same method. `RPCService#LogsTail` runs on the gRPC server and is completely transparent to clients. On the other hand, `StripeCLIClient#LogsTail` is the entry point for gRPC clients and is a different method entirely.

pkg/rpcservice/logs_tail.go
vcheung-stripe: No, it won't always be called with an error. It's probably better if I change it to send `Error: nil` if there's no error.

pkg/rpcservice/logs_tail.go
vcheung-stripe: Yeah!

PR #669, opened 2021-05-13
Implement `EventsResend` method

vcheung-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary

Implement `EventsResend`:
- It takes everything [`stripe events resend`](https://stripe.com/docs/cli/events/resend) takes except flags that only make sense in a terminal like color output or verbose level
- It resends the event with the provided params
- It returns the event object

Documentation:
Screen Shot 2021-05-12 at 6 22 48 PM

Screen Shot 2021-05-12 at 6 22 53 PM

Screen Shot 2021-05-12 at 6 23 01 PM

Example typescript client:
```ts
  try {
    const request = new EventsResendRequest()
      .setEventId("evt_1IqQYIFzgcm7CokLiuFAaniT")
      .setWebhookEndpoint("we_1IdIf9Fzgcm7CokLBo2yMzqF")
      .setStripeAccount("acct_1IMLXEFzgcm7CokL")
      .setExpandList(["request"])
      .setIdempotency("foo")
      .setLive(false);
    const response = await new Promise<EventsResendResponse>(
      (resolve, reject) => {
        client.eventsResend(request, (error, response) => {
          if (error) {
            reject(error);
          } else {
            resolve(response);
          }
        });
      }
    );
    console.log(response.toObject());
  } catch (e) {
    console.error(e);
  }

```


pkg/requests/base.go
```diff
@@ -20,24 +20,24 @@ import (
...
 // AppendData appends data to the request parameters.
 func (r *RequestParameters) AppendData(data []string) {
-	r.data = append(r.data, data...)
+	r.Data = append(r.Data, data...)
 }
 
 // SetStripeAccount sets the value for the `Stripe-Account` header.
 func (r *RequestParameters) SetStripeAccount(value string) {
-	r.stripeAccount = value
+	r.StripeAccount = value
```
tomer-stripe, line 40: Why do we export these? Part of the reason we have the `AppendData` and `SetStripeAccount` functions to set the data without exporting the actual field.

tomer-stripe (commented):
I don't have major feedback, might be worth rerolling to someone else who has more context. The one thing I was wondering about was having all the struct fields exported now which we had previously avoided. Are we doing that in other places too?

pkg/rpcservice/events_resend.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	var evt stripeEvent
+
+	err = json.Unmarshal(stripeResp, &evt)
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := structpb.NewStruct(evt.Data)
```
pepin-stripe, line 76: neat package!

pepin-stripe (approved):
LGTM

PR #674, opened 2021-05-20
Move endpoint response output out of Proxy and EndpointClient

vcheung-stripe:
### Reviewers
r? @pepin-stripe
cc @stripe/developer-products

 ### Summary

We missed moving the endpoint response output out of Proxy and EndpointClient. To move them, we introduce some new types so the visitor can handle them properly.

This refactor incidentally enables Proxy to return endpoint responses over the channel. Perhaps stripe daemon could leverage that in the Listen method :D


pepin-stripe (approved):
thanks for fixing that! LGTM

PR #676, opened 2021-05-20
Implement `Listen` method for gRPC

vcheung-stripe:
### Reviewers
r? @pepin-stripe
cc @stripe/developer-products

 ### Summary

Add `Listen` for gRPC, which takes the flags that `stripe listen` takes and returns a stream of
- state: the websocket connection state
- stripe_event: a Stripe event received by the CLI
- endpoint_response: one of
  - data: the response from the forward-to endpoint
  - error: an error involving the response (either failed to post or failed to read the response body)

This was made possible by #574! The Proxy class is very transparent after that refactor which is excellent. All I had to do was pass it the right params and handle the output. 🎉 

### Testing
I added several automated tests and tested this extensively on my local machine with different params. My typescript client that I tested this with:
```ts
try {
    const request = new ListenRequest()
      .setForwardTo("localhost:4242/webhook")
      .setHeadersList(["foo:bar", "bar:foo"])
      .setLatest(true)
      .setLive(false)
      .setUseConfiguredWebhooks(false)
      .setSkipVerify(true);
    const stream = client.listen(request);

    stream.on("data", (response: ListenResponse) => {
      if (response.hasState()) {
        console.log(response.getState());
      } else if (response.hasStripeEvent()) {
        console.log(response.getStripeEvent()?.toObject());
      } else if (response.hasEndpointResponse()) {
        console.log(response.getEndpointResponse()?.toObject());
      }
    });

    stream.on("error", (err) => {
      if ((err as grpc.ServiceError).code !== grpc.status.CANCELLED) {
        console.error(err);
      }
    });

    stream.on("end", () => {
      console.log("ended");
    });
  } catch (e) {
    console.error(e);
  }
```

I verified the server correctly sends an error when it fails to forward to the endpoint response.

I ran this client with the `accept-a-card-payment` sample running in the background listening for `payment_intent.succeeded `. I verified the sample received the event and headers. I verified the client received the endpoint's responses, whether it was 200 or 400.


docs/rpc/commands.md
```diff
@@ -6,10 +6,22 @@
 - [commands.proto](#commands.proto)
```
vcheung-stripe, line 6: Note: this file will be removed in the future because we should put docs in the CLI ref instead.

pkg/rpcservice/listen.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return err
+	}
+
+	key, err := srv.cfg.UserCfg.Profile.GetAPIKey(req.Live)
+	if err != nil {
+		return err
+	}
+
+	// validate forward-urls args
```
pepin-stripe, line 42: This seems like it is proxy-specific validation and should be handling by the proxy package (maybe directly by proxy.Init) instead of gRPC?

pkg/rpcservice/listen.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+	} else if req.UseConfiguredWebhooks && len(req.ForwardTo) == 0 {
+		return status.Error(codes.InvalidArgument, "load_from_webhooks_api requires a location to forward to with forward_to")
+	}
+
+	logger := log.StandardLogger()
+	proxyVisitor := createProxyVisitor(&stream)
+	proxyOutCh := make(chan websocket.IElement)
+
+	p, err := createProxy(&proxy.Config{
```
pepin-stripe, line 58: this is where an err should be re turned if one of the arguments is invalid (cc previous comment on urls validation). 

If there are any challenges with having the url validation extracted from here, I'd at least make it happen inside `createProxy`

pkg/rpcservice/listen.go
```diff
@@ -0,0 +1,200 @@
...
+				switch data.Resp.Request.Method {
+				case http.MethodDelete:
+					httpMethodResponse = rpc.ListenResponse_EndpointResponse_Data_HTTP_METHOD_DELETE
+				case http.MethodGet:
+					httpMethodResponse = rpc.ListenResponse_EndpointResponse_Data_HTTP_METHOD_GET
+				case http.MethodPost:
+					httpMethodResponse = rpc.ListenResponse_EndpointResponse_Data_HTTP_METHOD_POST
+				default:
+					httpMethodResponse = rpc.ListenResponse_EndpointResponse_Data_HTTP_METHOD_UNSPECIFIED
+				}
```
pepin-stripe, line 158: nit but I wonder if this could be a bit cleaner with a map and getter function, so we could load the correct rpc method inline

```
HttpMethod: getRPCMethodFromRequestMethod(data.Resp.Request.Method)
```

Or even further, could we have a `buildListenResponse` method that takes data and build the whole thing, so the case statement is a bit cleaner?

```
VisitData: func(de websocket.DataElement) error {
    switch data := de.Data.(type) {
    case proxy.StripeEvent:
        (*stream).Send(buildStripeEventResponse(data))
    case proxy.EndpointResponse:
        (*stream).Send(buildEndpointResponse(data))
    default:
        [...]
```

pepin-stripe: would also make testing a bit simpler as each independent function can be tested in isolation

pepin-stripe (approved):
LGTM but see comments - I think the visitor code could get cleaned up to make it easier to read. Lots of inline logic that makes it harder to parse.

pkg/rpcservice/listen.go
vcheung-stripe: Agreed, it should be in proxy.Init!

PR #677, opened 2021-05-20
Return Unauthenticated error if user isn't logged in for commands tha…

vcheung-stripe:
…t require it

 ### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products
 
 ### Summary

Right now, there are methods that require logging in before they can run: `EventsResend`, `Listen`, `LogsTail`, or `Trigger`. If a gRPC client invokes these commands before logging in, they will receive the default gRPC error code: `UNKNOWN`.

In this PR, we make the methods return the gRPC error code `UNAUTHENTICATED` instead. This allows gRPC clients to easily handle unauthentication. For example, if the Stripe vscode extension is the client and sees this error code, it could prompt the user log in again.

I had to "unrefactor" trigger.go a little to get this to work.

Tested this manually for all four methods.

This should be the last of the big gRPC changes! Everything after will either be maintenance (docs, CI, logging etc.) or unforeseen bug fixes.

### DIFF
--- a/pkg/cmd/trigger.go
+++ b/pkg/cmd/trigger.go
@@ -61,9 +61,14 @@ func (tc *triggerCmd) runTriggerCmd(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	apiKey, err := Config.Profile.GetAPIKey(false)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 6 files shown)


pepin-stripe (approved):
lgtm

PR #678, opened 2021-05-21
Add debug logs for gRPC method invocation

vcheung-stripe:
### Reviewers
r? @stripe/developer-products
cc @stripe/developer-products

 ### Summary

For debugging purposes, write debug logs when gRPC methods are invoked. Move the gRPC config output (in the screenshot, this is the host and port number on the first line) from stderr to stdout. For convenience, stderr will be reserved for logs only.

Screen Shot 2021-05-20 at 5 15 06 PM

### DIFF
--- a/pkg/rpcservice/middleware.go
+++ b/pkg/rpcservice/middleware.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


pepin-stripe (approved):
LGTM

PR #680, opened 2021-05-26
create WebhookEventAck message type and send it when receiving a webh…

pepin-stripe:
…ook event

This should be the one and only PR on the CLI side for the new Ack system.
Created a new `WebhookEventAck` type which gets wrapped by `OutgoingMessage` the same way `WebhookResponse` does.
When a `WebhookEvent` is received, once parsed without errors, the CLI sends a `WebhookEventAck` with webhook_id and webhook_conversation_id back to the server.

I think that should be enough to handle acknowledgements on the server side.

 ### Reviewers
r? @suz-stripe
cc @stripe/developer-products

 ### Summary


suz-stripe (approved):

pepin-stripe:
r-codeowners?

PR #715, opened 2021-07-28
Send close frame when terminating websocket connection

gracegoo-stripe:
### Reviewers
r? @suz-stripe 
cc @stripe/developer-products

 ### Summary
This change adds a new function that sends a proper close frame during the websocket connection termination. 
The codes are from https://datatracker.ietf.org/doc/html/rfc6455#section-7.4. 

### Testing
I added a log on the server that outputs the code and reason received on the server side. If you have other suggestions for testing let me know and I will run through those as well. 

Without this change we seem to have been getting a code anyway: 
```
DX_RUN-377: beginning close: reason="" code=1000
DX_RUN-220: beginning finalize_close of websocket: merchant_token=acct_28DT589O8KAxCGbLmxyZ websocket_id="cliws_{22078}_OKpaGryJBTOluaZzRd83q2euKIxYzJWJRTFksySl9zClrJbxuz" websocket_state=2
REDIS-EVENT-LOOP: Going to exit since we have no active conns: hash_slot=10000
```

With this change: 
```
DX_RUN-377: beginning close: reason="Connection Done" code=1000
DX_RUN-220: beginning finalize_close of websocket: merchant_token=acct_28DT589O8KAxCGbLmxyZ websocket_id="cliws_{5879}_QXr9LtFVuuxstuSnKvtrDCeOCeSFnBM2ckSnWU4nszo3lvcitG" websocket_state=2

```

### DIFF
--- a/pkg/websocket/client.go
+++ b/pkg/websocket/client.go
@@ -519,7 +519,7 @@ const (
 
 	defaultReconnectInterval = 60 * time.Second
 
-	defaultCloseDelayPeriod = 10 * time.Second
+	defaultCloseDelayPeriod = 1 * time.Second
 
 	defaultWriteWait = 1 * time.Second
 )
--- a/pkg/websocket/client.go
+++ b/pkg/websocket/client.go
@@ -167,7 +167,7 @@ func (c *Client) Run(ctx context.Context) {
 			}).Debug("Resetting the connection")
 			close(c.stopReadPump)
 			close(c.stopWritePump)
-			c.SendCloseFrame(ws.CloseServiceRestart, "Resetting the connection")
+			c.SendCloseFrame(ws.CloseNormalClosure, "Resetting the connection")
 			c.wg.Wait()
 		}
 	}
--- a/pkg/websocket/client.go
+++ b/pkg/websocket/client.go
@@ -144,14 +144,14 @@ func (c *Client) Run(ctx context.Context) {
 			close(c.send)
 			close(c.stopReadPump)
 			close(c.stopWritePump)
-
+			c.SendCloseFrame(ws.CloseNormalClosure, "Connection Done")
 			return
 		case <-c.done:
 			close(c.send)
 			close(c.stopReadPump)
 			close(c.stopWritePump)
 			close(c.NotifyExpired)
-
+			c.SendCloseFrame(ws.CloseNormalClosure, "Connection Done")
 			return
 		case <-c.notifyClose:
 			c.cfg.Log.WithFields(log.Fields{
@@ -159,6 +159,7 @@ func (c *Client) Run(ctx context.Context) {
 			}).Debug("Disconnected from Stripe")
 			close(c.stopReadPump)
 			close(c.stopWritePump)
+			c.SendCloseFrame(ws.CloseGoingAway, "Server closed the connection")
 			c.wg.Wait()
 		case <-time.After(c.cfg.ReconnectInterval):
 			c.cfg.Log.WithFields(log.Fields{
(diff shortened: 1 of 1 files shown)


pkg/websocket/client.go
```diff
@@ -144,35 +144,49 @@ func (c *Client) Run(ctx context.Context) {
...
 
-			c.wg.Wait()
+		err := c.conn.WriteControl(ws.CloseMessage, message, time.Now().Add(time.Second))
+		if err != nil {
+			c.cfg.Log.WithFields(log.Fields{
+				"prefix": "websocket.Client.Run",
+				"error":  err,
+			}).Debug("Error while trying to send close frame")
 		}
+		c.conn.Close()
```
tjfontaine-stripe, line 189: is it safe/reasonable to send the close frame and close the connection at approximately the same time? just looking at https://github.com/gorilla/websocket/blob/master/examples/echo/client.go it doesn't seem to do that (though I can imagine wanting a deadline not to be exceeded)

also for my own curiosity what's the difference between WriteControl and WriteMessage?

gracegoo-stripe: Looking into the safety question. 

Re: the diff between WriteControl and WriteMessage, I initially used WriteControl because the docs say that a close is a type of control message: https://github.com/gorilla/websocket/blob/master/conn.go#L73

I think the only difference is that WriteControl is concurrency-safe: https://pkg.go.dev/github.com/gorilla/websocket#hdr-Control_Messages

gracegoo-stripe: I think you're right, this other example has a wait period too: https://github.com/gorilla/websocket/blob/5ade364dff1f03abefdae800ed19074bf707b08f/examples/command/main.go#L75-L78

https://github.com/gorilla/websocket/blob/5ade364dff1f03abefdae800ed19074bf707b08f/examples/command/main.go#L38-L39

tjfontaine-stripe: ok that makes sense, thanks for the clarifications

gracegoo-stripe: Thanks so much for pointing that out btw! https://github.com/gorilla/websocket/issues/624#issuecomment-675728014

suz-stripe (approved):
LGTM 👍 thanks for carefully unwinding the logic around all the closure scenarios

gracegoo-stripe:
As I was updating the PR to add the close delay, I noticed that we already had a config for writeWait that we use for the WriteMessage method. I re-used that here but the value is 10 seconds instead of the 1 requested in the ticket. Is it okay if I change wait time of all writes to 1 second? Or create a separate config just for writing control messages?

suz-stripe:
@gracegoo-stripe that's a really good question. I'm not sure why we wait 10 seconds for a message write. Unless I'm mistaken that write wait is on the operating system socket side and does not include the remote / network transmission. Therefore I think it's reasonable to reduce that wait.

suz-stripe (approved):

pkg/websocket/client.go
```diff
@@ -503,7 +519,9 @@ const (
 
 	defaultReconnectInterval = 60 * time.Second
 
-	defaultWriteWait = 10 * time.Second
+	defaultCloseDelayPeriod = 10 * time.Second
```
suz-stripe, line 522: Sorry just caught this – this seems like a long time to sleep for. Given that most writes across a network should complete within 1 second even on a slow connection, it seems safe to reduce this number as well in order to ensure a speedy close sequence while still giving ample time for the server to receive the close frame. Thoughts?

suz-stripe (changes requested):

pkg/websocket/client.go
gracegoo-stripe: Works for me! Updated

suz-stripe (approved):

Issue #718, opened 2021-07-30
Fixture parameter value with environment variable is truncated


zmk-punchbowl:
### Issue
Using a Stripe fixture to set up a webhook, I'm using an environment variable to help build the `url` parameter. 

```
{
  "_meta": {
    "template_version": 0
  },
  "fixtures": [
    {
      "name": "create_webhook_endpoint",
      "path": "/v1/webhook_endpoints",
      "method": "post",
      "params": {
        "url": "${.env:BASE_API_URL}/hook/stripe",
        "enabled_events": [
          "customer.subscription.created",
          "customer.subscription.deleted",
          "customer.subscription.updated",
          "customer.subscription.pending_update_applied",
          "price.created",
          "price.updated"
        ]
      }
    }
  ]
}
```

The fixture succeeds, but the created webook `url` only includes the `BASE_API_URL` value (e.g. `https://myapi.example.com`) and not the `/hook/stripe` portion. It appears to truncate the text that comes after that environment variable.

### Expected Behavior
I expect the webhook `url` to reflect the full value defined in the fixture.

### Steps to reproduce
Run the above fixture (with that environment variable set)

### Traceback
(it succeeds)
```
Setting up fixture for: create_webhook_endpoint
```

### Environment
macOS
CLI version 1.7.0

gracegoo-stripe:
Thanks for the bug report! I've merged a fix for it here: https://github.com/stripe/stripe-cli/pull/719

PR #719, opened 2021-07-30
[718] Prevent truncation of env var queries

gracegoo-stripe:
### Reviewers
r? @ianjabour-stripe @suz-stripe 
cc @stripe/developer-products

 ### Summary
A user reported a bug where variables with a query embedded were losing all information outside of that query.
For example, a webhook create fixture with the url param set to `"${.env:BASE_API_URL}/hook/stripe"` was being saved with just the base API because we only keep the result of the query. 

I moved the .env query logic out and added a string replace to add the value back into the original query string. I only did this for the .env variables because I figured this didn't make as much sense for the resourceIds from responses but let me know if we think this will be useful for all query values, because I can also attempt to just move this logic outside of `parseQuery`

https://github.com/stripe/stripe-cli/issues/718

### Testing 
Added a unit test and also verified with the manual fixture

### DIFF
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -44,7 +44,7 @@ type fixture struct {
 }
 
 type fixtureQuery struct {
-	Match        string
+	Match        string // The substring that matched the query pattern regex
 	Name         string
 	Query        string
 	DefaultValue string
@@ -401,7 +401,8 @@ func (fxt *Fixture) parseQuery(queryString string) (string, error) {
 				return value, nil
 			}
 
-			fmt.Printf("value:  %s\n", value)
+			// Handle the case where only a substring of the original queryString was a query.
+			// Ex: ${.env:BLAH}/blah/blah
 			value = strings.ReplaceAll(queryString, query.Match, envValue)
 			return value, nil
 		}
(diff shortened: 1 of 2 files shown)


pkg/fixtures/fixtures.go
```diff
@@ -392,26 +395,16 @@ func (fxt *Fixture) parseQuery(value string) (string, error) {
...
-			if envValue == "" {
-				fmt.Printf("No value for env var: %s\n", key)
+			envValue, err := getEnvVar(query)
+			if err != nil || envValue == "" {
 				return value, nil
 			}
-			return envValue, nil
+
+			// Handle the case where only a substring of the original queryString was a query.
+			// Ex: ${.env:BLAH}/blah/blah
```
ianjabour-stripe, line 405: This comment was really helpful to understanding this

pkg/fixtures/fixtures.go
```diff
@@ -454,6 +447,30 @@ func (fxt *Fixture) parseQuery(value string) (string, error) {
 	return value, nil
 }
 
+func getEnvVar(query fixtureQuery) (string, error) {
```
ianjabour-stripe, line 450: I think this definitely makes sense to be it's own function. I can see it being used elsewhere eventually.

ianjabour-stripe (approved):
LGTM - Awesome that you were able to get a fix for this 😄

PR #727, opened 2021-08-06
Use `--forward-to` as default value of `--forward-connect-to` as documented

john-h-k:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary
The message said it but it was not true and would lead to confusing errors such as below when `--forward-to` was provided but `--forward-connect-to` was not.

```sh
[ERROR] Failed to POST: Post "http:///stripe/payment-webhook": http: no Host in request URL
```

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -453,6 +453,14 @@ func Init(cfg *Config) (*Proxy, error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


john-h-k:
@kwu-stripe fixes the other half of #723

pepin-stripe (approved):
Good catch, thank you for fixing this!

This will probably resolve #733 also!

PR #744, opened 2021-09-15
Send forward-to URLs when connecting via `stripe listen`

ianjabour-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary
Updated the Stripe Auth client to optionally send the forward and connect forward URLs to the CLI API service when connecting via `stripe listen`.

I wrote a test to verify the query params get properly included in the request and manually verified in production as well:

#### `stripe listen` with forward urls set
forwarding_urls

### `stripe listen` without forward urls set
no_forwarding_urls


vcheung-stripe (approved):
Code change and test lgtm.

ianjabour-stripe:
r-codeowners?

PR #747, opened 2021-09-17
Updating all instances of context.TODO() with context from caller

gracegoo-stripe:
### Reviewers
r? @bernerd-stripe 
cc @stripe/developer-products

 ### Summary
Updating all instances of context.TODO() to grab the context passed in by the caller. 
For runtime, this means that we pass in the context that's instantiated from the main() function. Most end up just being context.Background() which is no different from context.TODO(), but some get the context that has cancel set if the calling cmd set it. 

For tests, I just passed in an empty context. 

Motivation: This will allow us to use this object to pass along telemetry context for the upcoming migration. 

### Testing
Compiling and unit testing.
I also manually ran through some of my workspace commands.

### DIFF
--- a/pkg/login/poll.go
+++ b/pkg/login/poll.go
@@ -28,7 +28,7 @@ type PollAPIKeyResponse struct {
 }
 
 // PollForKey polls Stripe at the specified interval until either the API key is available or we've reached the max attempts.
-func PollForKey(pollURL string, interval time.Duration, maxAttempts int) (*PollAPIKeyResponse, *Account, error) {
+func PollForKey(ctx context.Context, pollURL string, interval time.Duration, maxAttempts int) (*PollAPIKeyResponse, *Account, error) {
 	var response PollAPIKeyResponse
 
 	if maxAttempts == 0 {
@@ -52,7 +52,7 @@ func PollForKey(pollURL string, interval time.Duration, maxAttempts int) (*PollA
 
 	var count = 0
 	for count < maxAttempts {
-		res, err := client.PerformRequest(context.TODO(), http.MethodGet, parsedURL.Path, parsedURL.Query().Encode(), nil)
+		res, err := client.PerformRequest(ctx, http.MethodGet, parsedURL.Path, parsedURL.Query().Encode(), nil)
 		if err != nil {
 			return nil, nil, err
 		}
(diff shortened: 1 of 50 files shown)


pkg/login/poll.go
```diff
@@ -52,7 +52,7 @@ func PollForKey(pollURL string, interval time.Duration, maxAttempts int) (*PollA
 
 	var count = 0
 	for count < maxAttempts {
-		res, err := client.PerformRequest(context.TODO(), http.MethodGet, parsedURL.Path, parsedURL.Query().Encode(), nil)
+		res, err := client.PerformRequest(ctx, http.MethodGet, parsedURL.Path, parsedURL.Query().Encode(), nil)
```
bernerd-stripe, line 55: 🙌 Love to see this!

pkg/rpcservice/logs_tail.go
```diff
@@ -54,7 +54,7 @@ func (srv *RPCService) LogsTail(req *rpc.LogsTailRequest, stream rpc.StripeCLI_L
 		NoWSS:      false,
 	})
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(stream.Context())
```
bernerd-stripe, line 57: Good find here, this is another great change!

pkg/rpcservice/sample_create_test.go
```diff
@@ -27,7 +27,13 @@ func TestSampleCreateSucceeds(t *testing.T) {
...
 	}
 
-	createSample = func(config *config.Config, sampleName string, selectedConfig *samples.SelectedConfig, destination string, forceRefresh bool, resultChan chan<- samples.CreationResult) {
+	createSample = func(ctx context.Context,
+		config *config.Config,
+		sampleName string,
+		selectedConfig *samples.SelectedConfig,
+		destination string,
+		forceRefresh bool,
+		resultChan chan<- samples.CreationResult) {
```
bernerd-stripe, line 36: Nit: go style typically prefers all arguments on one line or all arguments on their own line as below:

```suggestion
	createSample = func(
		ctx context.Context,
		config *config.Config,
		sampleName string,
		selectedConfig *samples.SelectedConfig,
		destination string,
		forceRefresh bool,
		resultChan chan<- samples.CreationResult,
	) {
```

bernerd-stripe (approved):

pkg/rpcservice/sample_create_test.go
gracegoo-stripe: Gotcha, updated!

PR #745, opened 2021-09-15
Display error message if flags were given without commands

etsai-stripe:
### Reviewers
r? @richardm-stripe 
cc @stripe/developer-products

 ### Summary

issue: https://github.com/stripe/stripe-cli/issues/349

```
st-etsai1:stripe-cli etsai$ stripe-dev
The official command-line tool to interact with Stripe.

Usage:
  stripe [command]

Webhook commands:
  listen                        Listen for webhook events
  trigger                       Trigger test webhook events

Stripe commands:
  logs                          Interact with Stripe API request logs
  status                        Check the status of the Stripe API

Resource commands:
  get                           Quickly retrieve resources from Stripe
  charges                       Make requests (capture, create, list, etc) on charges
  customers                     Make requests (create, delete, list, etc) on customers
  payment_intents               Make requests (cancel, capture, confirm, etc) on payment intents
  ...                           To see more resource commands, run `stripe resources help`

Other commands:
  community                     Chat with Stripe engineers and other developers
  completion                    Generate bash and zsh completion scripts
  config                        Manually change the config values for the CLI
  feedback                      Provide us with feedback on the CLI
  fixtures                      Run fixtures to populate your account with data
  help                          Help about any command
  login                         Login to your Stripe account
  logout                        Logout of your Stripe account
  open                          Quickly open Stripe pages
  samples                       Sample integrations built by Stripe
  serve                         Serve static files locally
  version                       Get the version of the Stripe CLI

Flags:
      --api-key string        Your API key to use for the command
      --color string          turn on/off color output (on, off, auto)
      --config string         config file (default is $HOME/.config/stripe/config.toml)
      --device-name string    device name
  -h, --help                  help for stripe
      --log-level string      log level (debug, info, trace, warn, error) (default "info")
  -p, --project-name string   the project name to read from for config (default "default")
  -v, --version               Get the version of the Stripe CLI

Use "stripe [command] --help" for more information about a command.
st-etsai1:stripe-cli etsai$ stripe-dev --color on
The official command-line tool to interact with Stripe.

Usage:
  stripe [command]

Webhook commands:
  listen                        Listen for webhook events
  trigger                       Trigger test webhook events

Stripe commands:
  logs                          Interact with Stripe API request logs
  status                        Check the status of the Stripe API

Resource commands:
  get                           Quickly retrieve resources from Stripe
  charges                       Make requests (capture, create, list, etc) on charges
  customers                     Make requests (create, delete, list, etc) on customers
[...]

PR #748, opened 2021-09-17
bug fix for --color warning message

etsai-stripe:
### Reviewers
r? @gracegoo-stripe 
cc @stripe/developer-products

 ### Summary

previous PR https://github.com/stripe/stripe-cli/pull/745 checks if a flag was given without command, and shows an error message when it happens. this breaks the other flags like `--version` flag experience when it does not require any command

the original issue was that the `--color` flag was confusing and it was not apparent to users that it needed a command

we will only prompt the user to include a command when `--color` flag was given
cobra also did not output any error message when user enter `stripe --color`
other scenarios where a flag was incorrectly entered would be caught by cobra

Screen Shot 2021-09-17 at 1 47 52 PM
Screen Shot 2021-09-17 at 1 51 22 PM

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -102,9 +102,10 @@ func Execute(ctx context.Context) {
 
 		os.Exit(1)
 	} else {
-		userInput := strings.Join(os.Args[1:], " ")
-		if strings.Contains(userInput, "--") {
-			fmt.Println("You provided flags but did not specify any command.")
+		userInput := os.Args[1:]
+		// --color on/off/auto
+		if len(userInput) == 2 && userInput[0] == "--color" {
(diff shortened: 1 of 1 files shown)


gracegoo-stripe (approved):
LGTM, thank you so much for fixing this so quickly!

pepin-stripe:
Can you explain what the bug was before and why we're moving to having a message specifically for `--color` instead of any flag please?

etsai-stripe:
@pepin-stripe i have added the issue background to the PR description. thanks!

gracegoo-stripe:
In addition, the original error message would show up even if the user had provided the command as well:
```
➜  stripe-cli git:(gracegoo-config-list-fix) stripe-workspace trigger balance.available --color=on
Setting up fixture for: payment_intent
Running fixture for: payment_intent
You provided flags but did not specify any command.
```

PR #752, opened 2021-09-23
CLI Telemetry Migration

gracegoo-stripe:
### Reviewers
r? @pepin-stripe, @bernerd-stripe 
cc @stripe/developer-products

 ### Summary
This change updates the telemetry delivery mechanism. Previously, we used to send telemetry data by adding it to the requests we make to Stripe. This decouples the telemetry from API requests which will allow us to send finer grain data. This change creates a new AnalyticsTelemetry client that sends events to r.stripe.com which is a Stripe-owned service that strictly deals with processing non-PII telemetry events.

As this change gets adopted, we will be getting data from both pipelines depending on which version the user has. To prepare for the migration, we already created a data job that can consume from both sources and consolidate it to a single table. 

### Testing
Updated unit tests and ran some manual tests


pkg/stripe/analytics_telemetry_test.go
```diff
@@ -0,0 +1,156 @@
...
+func TestSetMerchant(t *testing.T) {
+	tel := InitContext()
+	merchant := "acct_zzzzzz"
+	tel.SetMerchant(merchant)
+	require.Equal(t, merchant, tel.Merchant)
+}
+
+// AnalyticsClient Tests
+func TestSendAPIRequestEvent(t *testing.T) {
+	os.Setenv("STRIPE_CLI_TELEMETRY_OPTOUT", "0")
```
gracegoo-stripe, line 36: Not too happy about this... let me know if anyone has any ideas

pkg/stripe/analytics_telemetry.go
```diff
@@ -0,0 +1,169 @@
...
+		data.Set("event_value", eventValue)
+		data.Set("created", fmt.Sprint((time.Now().Unix())))
+
+		return a.sendData(ctx, data)
+	}
+	return nil, nil
+}
+
+func (a *AnalyticsTelemetryClient) sendData(ctx context.Context, data url.Values) (*http.Response, error) {
+	if telemetryOptedOut(os.Getenv("STRIPE_CLI_TELEMETRY_OPTOUT")) {
```
bernerd-stripe, line 133: I think this would be less surprising if this check was moved into `main.go`, so that the context only contains a telemetry client if we're supposed to be sending telemetry. Since the `sendTelemetryEvent` below skips sending events if the context doesn't contain the `TelemetryClientKey` key, I think it would "just work"?

That would also address your concern below about not being happy about setting env vars inside the tests -- instead you could test each case by either adding the client to the context or not.

pkg/stripe/analytics_telemetry.go
```diff
@@ -0,0 +1,169 @@
...
+	WG         *sync.WaitGroup
+	HTTPClient *http.Client
+}
+
+//
+// Public functions
+//
+
+// InitContext initializes an instance of CLIAnalyticsEventContext
+func InitContext() *CLIAnalyticsEventContext {
```
bernerd-stripe, line 66: `NewContext` would be a more natural name in Go for this function.

cmd/stripe/main.go
```diff
@@ -2,10 +2,27 @@ package main
...
+	httpClient := &http.Client{
+		Timeout: time.Second * 3,
+	}
+	telemetryClient := &stripe.AnalyticsTelemetryClient{WG: waitGroup, HTTPClient: httpClient}
+	contextWithTelemetry := context.WithValue(ctx, stripe.TelemetryClientKey{}, telemetryClient)
+
+	cmd.Execute(contextWithTelemetry)
+
+	// Wait for all telemetry calls to finish before existing the process
+	waitGroup.Wait()
```
bernerd-stripe, line 27: Wait groups can be tricky to use right, but I think you've actually got it right! However, I think it could be simplified and also not leak so much of its implementation.

If you updated the type of `AnalyticsTelemetryClient.WG` to be `sync.WaitGroup` instead of `*sync.WaitGroup`, then you could skip initializing it above, because the zero value of the wait group is usable by default. Then this line could either become `telemetryClient.WG.Wait()`, or you could even add a custom method to telemetry client (Wait? Drain?) and then keep the waitgroup internal.

pkg/stripe/analytics_telemetry.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// SetMerchant sets the merchant on the CLIAnalyticsEventContext object
+func (e *CLIAnalyticsEventContext) SetMerchant(merchant string) {
+	e.Merchant = merchant
+}
+
+// SendAPIRequestEvent is a special function for API requests
+func (a *AnalyticsTelemetryClient) SendAPIRequestEvent(ctx context.Context, requestID string, livemode bool) (*http.Response, error) {
+	a.WG.Add(1)
```
bernerd-stripe, line 96: Could this be moved into `sendData` so it's not necessary to remember to do this in all of the public send methods?

pkg/stripe/analytics_telemetry.go
```diff
@@ -0,0 +1,169 @@
...
+
+//
+// Public types
+//
+
+// TelemetryContextKey is the key for the telemetry context
+type TelemetryContextKey struct{}
+
+// TelemetryClientKey is the key for the telemetry client
+type TelemetryClientKey struct{}
```
bernerd-stripe, line 32: Not a blocker, but it's a best practice to keep these keys unexported, and instead define functions that allow you to get or set the the value. In particular, this lets callers avoid making type assertions, as you've had to do in a few places. For example:

```
func WithEventMetadata(context.Context, *CLIAnalyticsEventMetadata) context.Context
func GetEventMetadata(context.Context) *CLIAnalyticsEventMetadata
```

pkg/stripe/analytics_telemetry.go
```diff
@@ -0,0 +1,169 @@
...
+
+// DefaultTelemetryEndpoint is the default URL for the telemetry destination
+const DefaultTelemetryEndpoint = "https://r.stripe.com/0"
+
+// CLIAnalyticsEventContext is the structure that holds telemetry data context that is ultimately sent to the Stripe Analytics Service.
+type CLIAnalyticsEventContext struct {
```
bernerd-stripe, line 38: Context is an overloaded term here.. perhaps Metadata?

bernerd-stripe (commented):
This looks really great, I think the approach really payed off. I've left some ideas that could help clarify or fit more naturally into the Go ecosystem, but really nothing major!

pkg/stripe/analytics_telemetry.go
gracegoo-stripe: I like it, updated!

pkg/stripe/analytics_telemetry.go
gracegoo-stripe: updated

cmd/stripe/main.go
gracegoo-stripe: That makes sense to me, thanks!

pkg/stripe/analytics_telemetry.go
gracegoo-stripe: Great question! I actually had it in sendData initially but was worried that we could still run into race conditions where the program ends before we get through this setup portion of the public send methods, so figured it's safer to have it as early as possible. Not sure if that is a real concern in practice but if we want to make this more optional, what if I added an `add` and `defer` to `sendData` but also leave them in the public methods too? That way it's not necessary and just a precaution if the author of another method that uses `sendData` wants to add it?

pkg/stripe/analytics_telemetry.go
gracegoo-stripe: I like this much better, thanks!

pkg/stripe/analytics_telemetry.go
gracegoo-stripe: True, updated

pkg/stripe/analytics_telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+	WG         sync.WaitGroup
+	HTTPClient *http.Client
+}
+
+//
+// Public functions
+//
+
+// NewContext initializes an instance of CLIAnalyticsEventContext
+func NewContext() *CLIAnalyticsEventMetadata {
```
bernerd-stripe, line 65: As a reader, I definitely expected `NewContext` to return a `context.Context`.. perhaps `NewEventMetadata` to match `WithEventMetadata` and `GetEventMetadata` below?

pkg/stripe/analytics_telemetry.go
```diff
@@ -0,0 +1,203 @@
...
+// TelemetryClient is an interface that can send two types of events: an API request, and just general events.
+type TelemetryClient interface {
+	SendAPIRequestEvent(ctx context.Context, requestID string, livemode bool) (*http.Response, error)
+	SendEvent(ctx context.Context, eventName string, eventValue string) (*http.Response, error)
+}
+
+// AnalyticsTelemetryClient sends event information to r.stripe.com
+type AnalyticsTelemetryClient struct {
+	BaseURL    *url.URL
+	WG         sync.WaitGroup
```
bernerd-stripe, line 56: Could this be `wg` now so that it's not exported for code outside of the stripe pacakge to use directly?

pkg/stripe/analytics_telemetry.go
bernerd-stripe: That makes sense to me!

bernerd-stripe (approved):
Left ideas for two small improvements, but otherwise looks amazing! Really excited to see this go out 💃

pkg/stripe/analytics_telemetry.go
gracegoo-stripe: Yes!

pkg/stripe/analytics_telemetry.go
gracegoo-stripe: d'oh, that's my bad.. updating

bernerd-stripe (approved):

PR #759, opened 2021-09-29
Adding telemetry to gPRC calls

gracegoo-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

### Summary
This change adds additional logic in the middleware interceptors to add the `telemetryClient` and `eventMetadata` to the context passed into the RPC functions. This allows us to get telemetry events for each separate RPC call instead of sharing one metadata for the whole of the daemon process. 

An instance of an `eventMetadata` is created on each method call to gRPC.

We use the same instance of `telemetryClient` for the duration the RPC service is up. The instance of the `telemeryClient` is passed to the RPC service when it is created by the daemon command. 

### Testing
An example of data sent to the analytics service through manual testing
```
 p_invocation_id=1beaa848-7919-407d-81e5-96aeb9d6c24c p_user_agent="Stripe/v1 stripe-cli/master" p_os=darwin p_cli_version=master p_livemode=false p_generated_resource=false p_event_value= p_client_id=stripe-cli p_event_id=01f20da5-258b-4fe4-8a96-52813f1b9bc3 p_request_id=req_t1MWVF7KlGFz54 p_command_path=/rpc.StripeCLI/LogsTail p_created=1632951950 p_merchant=acct_1HzAHfIMIxZxvOV4 p_event_name="API Request"

p_invocation_id=eb1fb4ac-dff7-4de7-9e11-561b14e1072b p_user_agent="Stripe/v1 stripe-cli/master" p_os=darwin p_cli_version=master p_livemode=false p_generated_resource=false p_event_value= p_client_id=stripe-cli p_event_id=006011c9-c4cf-4a9c-aa6e-aa0ade98a679 p_request_id=req_Sseb1LmF1NhqVj p_command_path=/rpc.StripeCLI/Listen p_created=1632951941 p_merchant=acct_1HzAHfIMIxZxvOV4 p_event_name="API Request"
```

Regarding unit tests: 
I wasn't sure what the best method of adding unit tests to test the new interceptor logic was. Calling `DialContext` allows us to hit the interceptor logic in the middleware.go class but I'm not sure how to assert that the context updated in the way we expected. In the other tests, we pass in a `ctx` directly into the gRPC calls which would bypass the logic we are trying to test. Let me know if anyone has ideas.

### DIFF
--- a/pkg/rpcservice/middleware.go
+++ b/pkg/rpcservice/middleware.go
@@ -52,6 +52,11 @@ func authorize(ctx context.Context) error {
 // 1. The telemetry client from the RPC Service
 // 2. The event metadata
 func updateContextWithTelemetry(ctx context.Context, methodName string, server *RPCService) context.Context {
+	// If the context is nil for whatever reason, create an empty one
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
 	// if getting the config errors, don't fail running the command
 	merchant, _ := server.cfg.UserCfg.Profile.GetAccountID()
 
@@ -73,10 +78,10 @@ func serverStreamInterceptor(
 	log.WithFields(log.Fields{
 		"prefix": "gRPC",
 	}).Debugf("Streaming method invoked: %v", info.FullMethod)
-	if err := authorize(stream.Context()); err != nil {
+	wrappedStream := newWrappedStream(stream, info.FullMethod, srv.(*RPCService))
+	if err := authorize(wrappedStream.Context()); err != nil {
 		return err
 	}
-	wrappedStream := newWrappedStream(stream, info.FullMethod, srv.(*RPCService))
 	return handler(srv, wrappedStream)
 }
 
@@ -90,9 +95,9 @@ func serverUnaryInterceptor(
 	log.WithFields(log.Fields{
 		"prefix": "gRPC",
 	}).Debugf("Unary method invoked: %v, req: %v", info.FullMethod, req)
-	if err := authorize(ctx); err != nil {
+	newCtx := updateContextWithTelemetry(ctx, info.FullMethod, info.Server.(*RPCService))
+	if err := authorize(newCtx); err != nil {
 		return nil, err
 	}
-	newCtx := updateContextWithTelemetry(ctx, info.FullMethod, info.Server.(*RPCService))
 	return handler(newCtx, req)
 }
--- a/pkg/rpcservice/middleware.go
+++ b/pkg/rpcservice/middleware.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	log "github.com/sirupsen/logrus"
+
 	"github.com/stripe/stripe-cli/pkg/stripe"
 
 	"google.golang.org/grpc"
@@ -15,19 +16,20 @@ import (
 
 const requiredHeader = "sec-x-stripe-cli"
 
-// It's not easy to pass values through context for streams
+// WrappedServerStream wraps a ServerSteam so that we can pass values through context.
 // https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware#hdr-Writing_Your_Own
 type WrappedServerStream struct {
 	grpc.ServerStream
 	ctx context.Context
 }
 
+// Context returns the context for this stream.
 func (w WrappedServerStream) Context() context.Context {
 	return w.ctx
 }
 
-func newWrappedStream(server *RPCService, stream grpc.ServerStream, methodName string) grpc.ServerStream {
-	newCtx := initializeTelemetryContext(stream.Context(), methodName, server)
+func newWrappedStream(stream grpc.ServerStream, methodName string, server *RPCService) grpc.ServerStream {
+	newCtx := updateContextWithTelemetry(stream.Context(), methodName, server)
 	return &WrappedServerStream{stream, newCtx}
 }
 
@@ -46,7 +48,10 @@ func authorize(ctx context.Context) error {
 	return nil
 }
 
-func initializeTelemetryContext(ctx context.Context, methodName string, server *RPCService) context.Context {
+// Populate the context with:
+// 1. The telemetry client from the RPC Service
+// 2. The event metadata
+func updateContextWithTelemetry(ctx context.Context, methodName string, server *RPCService) context.Context {
 	// if getting the config errors, don't fail running the command
 	merchant, _ := server.cfg.UserCfg.Profile.GetAccountID()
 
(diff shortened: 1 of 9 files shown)


pkg/rpcservice/rpc_service.go
```diff
@@ -34,6 +35,9 @@ type RPCService struct {
 	cfg *Config
 
 	grpcServer *grpc.Server
+
+	// TelemetryClient to use for sending telemetry events
+	TelemetryClient stripe.TelemetryClient
```
etsai-stripe, line 40: why is `TelemetryClient` capitalized when other class attrs are not?

pkg/rpcservice/middleware.go
```diff
@@ -14,6 +16,23 @@ import (
...
+// WrappedServerStream wraps a ServerSteam so that we can pass values through context.
+// https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware#hdr-Writing_Your_Own
+type WrappedServerStream struct {
+	grpc.ServerStream
+	ctx context.Context
+}
+
+// Context returns the context for this stream.
+func (w WrappedServerStream) Context() context.Context {
+	return w.ctx
```
etsai-stripe, line 28: does it break `WithTelemetryClient` if `w.ctx` is `null`? or can it be `null`?

pkg/rpcservice/middleware.go
```diff
@@ -29,6 +48,21 @@ func authorize(ctx context.Context) error {
...
 }
 
+// Populate the context with:
+// 1. The telemetry client from the RPC Service
+// 2. The event metadata
+func updateContextWithTelemetry(ctx context.Context, methodName string, server *RPCService) context.Context {
+	// if getting the config errors, don't fail running the command
+	merchant, _ := server.cfg.UserCfg.Profile.GetAccountID()
+
+	telemetryMetadata := stripe.NewEventMetadata()
```
etsai-stripe, line 58: could the `telemetryMetada` be empty? if its empty, should we still update the context?

pkg/rpcservice/rpc_service.go
vcheung-stripe: This is for making the `TelemetryClient` accessible publicly for the interceptor right?

vcheung-stripe (commented):
I'm very happy with how decoupled this is from the gRPC logic! Where is the user_agent being set? Or will that be in a future PR?

vcheung-stripe:
Regarding tests: I don't know of a way to test that the context is being set correctly in the gRPC calls. It might be worth unit testing `updateContextWithTelemetry` by itself.

pkg/rpcservice/rpc_service.go
gracegoo-stripe: Yup, that's correct!

pkg/rpcservice/middleware.go
gracegoo-stripe: If `w.ctx` is nil, this function would return a `nil` to the caller to handle.

pkg/rpcservice/middleware.go
gracegoo-stripe: Good question! I don't think it can be since we own the `NewEventMetadata` function and we know that if will create an instance of the metadata. It's possible that not all fields in there would be set as we'd expect but i think we'd still want the metadata in the context in that case so we can still send telemetry events

gracegoo-stripe:
> I'm very happy with how decoupled this is from the gRPC logic! Where is the user_agent being set? Or will that be in a future PR?

Oops that's a great catch, I want to set it with this PR! I'll update shortly

gracegoo-stripe:
> Regarding tests: I don't know of a way to test that the context is being set correctly in the gRPC calls. It might be worth unit testing `updateContextWithTelemetry` by itself.

Sounds good to me

pkg/rpcservice/middleware.go
gracegoo-stripe: On second thought, I don't see a problem with checking here and passing in a `context.Background()` if for some reason the context ends up being nil. I'll make that update

gracegoo-stripe:
Added logic to extract user_agent from the incoming metadata. Example of new params getting sent to r.stripe.com:
```

p_invocation_id=449f13f4-2fd3-4853-bd3e-643b24a06d74 p_user_agent=grpc-node-js/1.3.7 p_os=darwin p_cli_version=master p_livemode=false p_generated_resource=false p_event_value= p_client_id=stripe-cli p_event_id=530821f2-b5f8-4f37-b9fe-4fe80661386d p_request_id=req_hC5y5zlQbs1MhR p_command_path=/rpc.StripeCLI/LogsTail p_created=1633025624 p_merchant=acct_1HzAHfIMIxZxvOV4 p_event_name="API Request"
```

pkg/rpcservice/middleware.go
```diff
@@ -29,6 +48,41 @@ func authorize(ctx context.Context) error {
...
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return ""
+	}
+
+	if len(md["user-agent"]) != 1 {
+		return ""
+	}
+
+	return md["user-agent"][0]
```
vcheung-stripe, line 83: I've seen people `strings.Join` the array of user agents, so I'm wondering if we need to do that. I *think* if your client sets a custom user agent, there could be multiple ones: the default one from the gRPC library, and your custom one. But I'm not sure.

gracegoo-stripe: I need to mess around with sending a custom agent from the client but I assumed it would be something like this since the user_agent is a list:
```
"user_agent":["grpc-node-js/1.3.7", "vscode/2.6.1"]
```

But if people use a strings.Join, it would look something like this? 
```
"user_agent":["grpc-node-js/1.3.7, vscode/2.6.1"]
```
For the former, I need to mess around with it to see if there is an order to these (i.e. the default one always comes first so we'd want the lastest one). For the latter, I'm not quite sure what we'd do there. Let me look into this a bit more.

vcheung-stripe: Oh sorry, I meant that we might need to handle the case where the client sends an array of multiple user agents like `"user_agent":["grpc-node-js/1.3.7", "vscode/2.6.1"]`. In that case, I've seen other gRPC server code join the strings to together into `"grpc-node-js/1.3.7, vscode/2.6.1"` or something, so I was wondering if we'll want our interceptor to do the same. If we join them, I don't think order will matter.

gracegoo-stripe: Ohh, I see what you mean. Yeah I like the idea of including them all. We can choose to filter in them in the hubble dashboard if we want to

vcheung-stripe (approved):
Looks great, thanks for adding user agent 👍

PR #778, opened 2021-11-02
Fix kernel panic when not logged in

### LINKED ISSUE #774: panic when running stripe without first running `stripe login`

### Issue

I was following a tutorial which asks me to run `stripe products list`. I overlooked the part asking to login; so I installed `stripe` via homebrew, and run that. 

This didn't work. I got an error message saying I needed to login first (correct), but this was followed by a go panic. 

After logging in correctly all worked fine.

This is on  a machine that never run the `stripe` CLI tool before, on macOS 11.6 (Intel)

### Expected Behavior

The tool should not panic, but return after printing the "you must log in" error message.  

### Environment

macOS, 11.6, Intel

vcheung-stripe:
### Reviewers
r? @stripe/developer-products
cc @stripe/developer-products

 ### Summary

Fixes #774

### DIFF
--- a/pkg/stripe/client.go
+++ b/pkg/stripe/client.go
@@ -90,7 +90,12 @@ func (c *Client) PerformRequest(ctx context.Context, method, path string, params
 	// RequestID of the API Request
 	requestID := resp.Header.Get("Request-Id")
 	livemode := strings.Contains(c.APIKey, "live")
-	go sendTelemetryEvent(ctx, requestID, livemode)
+
+	// Weird edge case: ctx can be nil if the user runs a command like `stripe products list` but isn't logged in
+	if ctx != nil {
+		go sendTelemetryEvent(ctx, requestID, livemode)
+	}
+
 	return resp, nil
 }
 
--- a/pkg/stripe/client_test.go
+++ b/pkg/stripe/client_test.go
@@ -142,3 +142,21 @@ func TestPerformRequest_ConfigureFunc(t *testing.T) {
 
 	defer resp.Body.Close()
 }
+
+func TestPerformRequest_DoesNotPanicWithNilContext(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	}))
+	defer ts.Close()
+
+	baseURL, _ := url.Parse(ts.URL)
+	client := Client{
+		BaseURL: baseURL,
+	}
+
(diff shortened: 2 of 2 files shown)


ianjabour-stripe (approved):
👍

gracegoo-stripe (commented):
Thanks for fixing this!

pkg/stripe/client.go
```diff
@@ -90,7 +90,12 @@ func (c *Client) PerformRequest(ctx context.Context, method, path string, params
 	// RequestID of the API Request
 	requestID := resp.Header.Get("Request-Id")
 	livemode := strings.Contains(c.APIKey, "live")
-	go sendTelemetryEvent(ctx, requestID, livemode)
+
+	// Weird edge case: ctx can be nil if the user runs a command like `stripe products list` but isn't logged in
```
bernerd-stripe, line 94: I'd say this is alright as a hotfix, but context should really almost never be passed around as `nil`. The real bug is likely somewhere else, where the correct context is either not being propagated, or, if it's not available, it's not being set to [context.Background](https://pkg.go.dev/context#Background) or [context.TODO](https://pkg.go.dev/context#TODO).

If we're not sure where that is (it should be discoverable from the panic stackstace), this is good for now and I'd be happy to pair on a followup task to trace down the source of the nil!

gracegoo-stripe: Yup, you're right and I am looking into the root cause.

It's coming from here: https://github.com/stripe/stripe-cli/blob/0c97aabe2b0602a4b7f95a50df5018dd0279368c/pkg/cmd/root.go#L97

Right now I'm trying to understand how we are supposed to correctly invoke this command with context.
```
loginCommand.ExecuteContext(updatedCtx);
```
does not do what I thought it would

vcheung-stripe: One thing we can do is to replace that line with https://github.com/stripe/stripe-cli/blob/0c97aabe2b0602a4b7f95a50df5018dd0279368c/pkg/cmd/login.go#L43 directly rather than calling `Run` or `Execute`. WDYT?

gracegoo-stripe: Okay, the reason `loginCommand.ExecuteContext(updatedCtx);` doesn't work is because the cobra command is still for trigger and all this does is try to run the trigger command again. I think your solution makes more sense than us trying to mimick another cobra invocation, Vincent!

PR #780, opened 2021-11-02
Updating login logic to call login directly instead of invoking the c…

### LINKED ISSUE #774: panic when running stripe without first running `stripe login`

gracegoo-stripe:
…obra command

If the user is not logged for a command that requires them to be, we automatically try to run the login logic for them. 
The previous logic resulted in us losing the context when running the login command.
Fixes #774

 ### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary

Before:
Screen Shot 2021-11-02 at 12 52 05 PM

After:
Screen Shot 2021-11-02 at 12 53 05 PM

Note the login succeeds but we still return a status 1 because the original run failed. I believe this was the case before as well, should we try to re-run the original command after login automatically?

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/stripe/stripe-cli/pkg/cmd/resource"
 	"github.com/stripe/stripe-cli/pkg/config"
+	"github.com/stripe/stripe-cli/pkg/login"
 	"github.com/stripe/stripe-cli/pkg/stripe"
 	"github.com/stripe/stripe-cli/pkg/useragent"
 	"github.com/stripe/stripe-cli/pkg/validators"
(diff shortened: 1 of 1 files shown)


vcheung-stripe (approved):
LGTM, closing https://github.com/stripe/stripe-cli/pull/778

PR #792, opened 2021-12-02
Show friendly message when API key is expired

bernerd-stripe:
### Reviewers
r? @stripe/developer-products
cc @remi-stripe 

 ### Summary

The restricted API keys created by the CLI expire after 90 days, but any restricted API key can also be forcibly expired by rolling. Previously, when invoking the CLI with an expired key we would render the entire error response body (and also returned a successful exit code):

```
$ stripe accounts retrieve
{
  "error": {
    "code": "api_key_expired",
    "doc_url": "https://stripe.com/docs/error-codes/api-key-expired",
    "message": "Expired API Key provided: rk_test_*********************************************************************************************aaaaaa",
  }
}
$ echo $?
0
```

This updates the CLI to handle the `api_key_expired` error and show a tailored error message:

```
$ stripe accounts retrieve
The API key provided has expired. Obtain a new key from the Dashboard or run `stripe login` and try again.
$ echo $?
1
```

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -85,7 +85,7 @@ func Execute(ctx context.Context) {
 
 		switch {
 		case requests.IsAPIKeyExpiredError(err):
-			fmt.Fprintln(os.Stderr, "The API key provided has expired. Obtain a new key from the Dashboard or run `stripe login` to continue")
+			fmt.Fprintln(os.Stderr, "The API key provided has expired. Obtain a new key from the Dashboard or run `stripe login` and try again.")
 		case isLoginRequiredError:
 			// capitalize first letter of error because linter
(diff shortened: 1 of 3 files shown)


pepin-stripe (approved):
Thanks, that looks great!

The one question I have (non blocking) is, are there other places where we might face an expired api error (stripe listen, etc) where we could leverage that?

bernerd-stripe:
> The one question I have (non blocking) is, are there other places where we might face an expired api error (stripe listen, etc) where we could leverage that?

It looks like yes, but would require some more thought because [stripe listen uses the CLI's client package directly](https://github.com/stripe/stripe-cli/blob/master/pkg/stripeauth/client.go#L73-L93), so it doesn't share any error processing logic with the requests package.

PR #793, opened 2021-12-03
use generic dev tool csat form

etsai-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary

Screen Shot 2021-12-03 at 2 35 43 PM


vcheung-stripe (approved):
Awesome!

PR #798, opened 2021-12-07
fix: deadlock when resetting connection

acomley-stripe:
### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products

 ### Summary
When the timer triggers to reset the connection, a deadlock can occur if the ping write fails during that period.

### DIFF
--- a/pkg/websocket/client.go
+++ b/pkg/websocket/client.go
@@ -438,7 +438,7 @@ func (c *Client) writePump() {
 					c.cfg.Log.Error("write error: ", err)
 				}
 
-				// writing to notifyClose during
+				// writing to notifyClose during a reset will cause a deadlock
 				select {
 				case c.notifyClose <- err:
 					c.cfg.Log.WithFields(log.Fields{
--- a/pkg/websocket/client.go
+++ b/pkg/websocket/client.go
@@ -443,11 +443,11 @@ func (c *Client) writePump() {
 				case c.notifyClose <- err:
 					c.cfg.Log.WithFields(log.Fields{
 						"prefix": "websocket.Client.writePump",
-					}).Debug("ping notifyClose")
+					}).Debug("Failed to send ping; closing connection")
 				case <-c.stopWritePump:
 					c.cfg.Log.WithFields(log.Fields{
 						"prefix": "websocket.Client.writePump",
-					}).Debug("ping stopWritePump")
+					}).Debug("Failed to send ping; connection is resetting")
 				}
 				return
 			}
--- a/pkg/websocket/client.go
+++ b/pkg/websocket/client.go
@@ -82,6 +82,7 @@ type Client struct {
 	conn        *ws.Conn
 	done        chan struct{}
 	isConnected bool
+	isResetting bool
 
 	NotifyExpired chan struct{}
 	notifyClose   chan error
@@ -162,8 +163,10 @@ func (c *Client) Run(ctx context.Context) {
 			c.cfg.Log.WithFields(log.Fields{
 				"prefix": "websocket.Client.Run",
 			}).Debug("Resetting the connection")
+			c.isResetting = true
 			c.Close(ws.CloseNormalClosure, "Resetting the connection")
 			c.wg.Wait()
+			c.isResetting = false
 		}
 	}
 }
@@ -179,7 +182,7 @@ func (c *Client) Close(closeCode int, text string) {
 		err := c.conn.WriteControl(ws.CloseMessage, message, time.Now().Add(c.cfg.WriteWait))
 		if err != nil {
 			c.cfg.Log.WithFields(log.Fields{
-				"prefix": "websocket.Client.Run",
+				"prefix": "websocket.Client.Close",
 				"error":  err,
 			}).Debug("Error while trying to send close frame")
 		}
@@ -437,7 +440,9 @@ func (c *Client) writePump() {
 				if ws.IsUnexpectedCloseError(err, ws.CloseNormalClosure) {
 					c.cfg.Log.Error("write error: ", err)
 				}
-				c.notifyClose <- err
+				if !c.isResetting {
+					c.notifyClose <- err
+				}
 
 				return
 			}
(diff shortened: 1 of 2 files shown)


pkg/websocket/client.go
```diff
@@ -162,8 +163,10 @@ func (c *Client) Run(ctx context.Context) {
 			c.cfg.Log.WithFields(log.Fields{
 				"prefix": "websocket.Client.Run",
 			}).Debug("Resetting the connection")
+			c.isResetting = true
 			c.Close(ws.CloseNormalClosure, "Resetting the connection")
 			c.wg.Wait()
+			c.isResetting = false
```
acomley-stripe, line 169: I don't know if this will always be effective at preventing the deadlock. Possible that I need some proper synchronization between the go routines?

pepin-stripe: This is probably fine - what I'm wondering is 
1. why are we not `wg.Wait`ing for the 2 first switch cases
2. should we set/unset `c.isResetting` in switch cases, just to keep things more consistent
3. depending on how much we want Run and read/writePump to be sync/aware from each other, we could also make the `notifyClose` chan buffered, which would avoid making sends to it blocking. ReadPump and WritePump both send to it, so buffer of len 2 would prob be enough.

I'm honestly not sure what the best is, we can discuss tomorrow 1:1 about the pros/cons of each.

pkg/websocket/client.go
```diff
@@ -437,7 +440,9 @@ func (c *Client) writePump() {
 				if ws.IsUnexpectedCloseError(err, ws.CloseNormalClosure) {
 					c.cfg.Log.Error("write error: ", err)
 				}
-				c.notifyClose <- err
```
bernerd-stripe, line 440: The most reliable way to handle something like this is to use a select over sending and a shutdown signal (or whatever signals that nothing will be reading the other side). In this case, there's [`c.stopWritePump`](https://github.com/stripe/stripe-cli/blob/aed27b56a08c690f9ca1a83dcbecda5e59ef5628/pkg/websocket/client.go#L449), which seems to be closed and re-initialized when the connection is reset.

So perhaps something like this would resolve the deadlock?

```go
select {
case c.notifyClose <- err:
case <-c.stopWritePump:
}
```

I think there might still be a race reading `stopWritePump` from the client struct, however, since it appears to be swapped out without locking or coordination. I might suggest, then, also updating `go c.writePump()` to `go c.writePump(c.stopWritePump)` so writePump can always read the correct instance of the channel from its arguments.

pkg/websocket/client.go
```diff
@@ -162,8 +163,10 @@ func (c *Client) Run(ctx context.Context) {
 			c.cfg.Log.WithFields(log.Fields{
 				"prefix": "websocket.Client.Run",
 			}).Debug("Resetting the connection")
+			c.isResetting = true
```
bernerd-stripe, line 166: I believe this will introduce a data race, since the field is read and written to from separate goroutines without other coordination (locks, channels). I'd expect it to therefore suffer similar problems to the deadlock.

acomley-stripe: i mentioned that in my comment above. I'm going to think about what you said above, because I believe there are a few other race conditions currently that are less frequent.

pkg/websocket/client.go
acomley-stripe: I need the go equivalent of a promise. I want close to only run one-at-a-time, but if it is already running it should return the current running promise if that makes sense.

pkg/websocket/client.go
acomley-stripe: Boolean state was removed.

pkg/websocket/client.go
acomley-stripe: Thanks for this Bernerd! This is the method we went with.

pkg/websocket/client.go
acomley-stripe: Boolean state was removed in favor of select on channels.

pkg/websocket/client.go
```diff
@@ -437,8 +437,18 @@ func (c *Client) writePump() {
 				if ws.IsUnexpectedCloseError(err, ws.CloseNormalClosure) {
 					c.cfg.Log.Error("write error: ", err)
 				}
-				c.notifyClose <- err
 
+				// writing to notifyClose during
```
pepin-stripe, line 441: did you forget to finish the comment?

pepin-stripe: I'd add a quick mention about that deadlock we're fixing

pkg/websocket/client.go
```diff
@@ -437,8 +437,18 @@ func (c *Client) writePump() {
...
+				select {
+				case c.notifyClose <- err:
+					c.cfg.Log.WithFields(log.Fields{
+						"prefix": "websocket.Client.writePump",
+					}).Debug("ping notifyClose")
```
pepin-stripe, line 446: That might get confusing for someone trying to debug logs without context. a message like "No response from ping, closing connection"

and (below):
"no response from ping because connection seems to be closing" or something like that.

pkg/websocket/client.go
acomley-stripe: ha! Yes. Yes, I did.

pkg/websocket/client.go
acomley-stripe: Updated to "Failed to send ping; closing connection" and "Failed to send ping; connection is resetting".

pkg/websocket/client.go
```diff
@@ -437,8 +437,18 @@ func (c *Client) writePump() {
...
+				// writing to notifyClose during a reset will cause a deadlock
+				select {
+				case c.notifyClose <- err:
+					c.cfg.Log.WithFields(log.Fields{
+						"prefix": "websocket.Client.writePump",
+					}).Debug("Failed to send ping; closing connection")
+				case <-c.stopWritePump:
+					c.cfg.Log.WithFields(log.Fields{
+						"prefix": "websocket.Client.writePump",
+					}).Debug("Failed to send ping; connection is resetting")
```
pepin-stripe, line 450: love those!

pepin-stripe (approved):

PR #800, opened 2021-12-13
decoupling playback from listen

lemuel-stripe:
### Reviewers
r? @pepin-stripe 
cc @stripe/developer-products 

 ### Summary

This PR is one possible approach to prevent the crash when running `stripe playback`.

Currently when we run `stripe playback` we get:
```
Starting `stripe listen` to proxy webhooks to playback server...
panic: cannot create context from nil parent

goroutine 9 [running]:
context.WithCancel({0x0, 0x0})
        /usr/local/Cellar/go/1.17.3/libexec/src/context/context.go:234 +0x12f
github.com/stripe/stripe-cli/pkg/cmd.withSIGTERMCancel({0x0, 0x0}, 0x196c9c8)
        /Users/lemuel/stripe/stripe-cli/pkg/cmd/listen.go:173 +0x2f
github.com/stripe/stripe-cli/pkg/cmd.(*listenCmd).runListenCmd(0xc0000cd8c0, 0xc00041ca00, {0xc000059fc0, 0x1, 0x1})
        /Users/lemuel/stripe/stripe-cli/pkg/cmd/listen.go:117 +0xc5
github.com/stripe/stripe-cli/pkg/cmd.startListenCmdLoop.func1()
        /Users/lemuel/stripe/stripe-cli/pkg/cmd/playback.go:198 +0x76
created by github.com/stripe/stripe-cli/pkg/cmd.startListenCmdLoop
        /Users/lemuel/stripe/stripe-cli/pkg/cmd/playback.go:203 +0xdb
exit status 2
```

The call `cmd.withSIGTERMCancel` requires context, but currently none is passed to it implicitly since the new listen command isn't called by root. This PR explicitly passes the context to the listen command.

note: if we simply add the new context as `context.Background()` then there is no crash in playback as well. However, #747 makes it seems that if we can pass context from a parent, we should


pkg/cmd/listen.go
```diff
@@ -114,7 +117,7 @@ func (lc *listenCmd) runListenCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	ctx := withSIGTERMCancel(cmd.Context(), func() {
+	ctx := withSIGTERMCancel(lc.parentContext, func() {
```
lemuel-stripe, line 120: note: if we simply replace `cmd.Context()` with `context.Background()` then there is no crash in playback as well. However, #747 makes it seems that if we can pass context from a parent, we should

pkg/cmd/listen.go
```diff
@@ -43,9 +43,10 @@ type listenCmd struct {
 	skipUpdate            bool
 	apiBaseURL            string
 	noWSS                 bool
+	parentContext         context.Context
```
bernerd-stripe, line 46: [Storing contexts in a struct is generally discouraged](https://github.com/golang/go/wiki/CodeReviewComments#contexts), because it makes it hard to reason about the state of the system or lifetimes of in-flight work. One of the exceptions is if it's necessary to meet an interface you don't control, which would apply here, but I _think_ we should be able to solve it without this.

pkg/cmd/playback.go
```diff
@@ -145,7 +145,7 @@ func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
 
 	// --- Setup `stripe listen` and run it now if not in replay-only mode, else listen for httpWrapper.ChangeModeChan.
 	if !pc.noListen {
-		startListenCmdLoop(pc.mode, addressString, httpWrapper)
+		startListenCmdLoop(pc.cmd, pc.mode, addressString, httpWrapper)
```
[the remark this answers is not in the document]
lemuel-stripe: yes you are right, I think thats a good refactoring of this!

pkg/cmd/listen.go
lemuel-stripe: Yes, I agree & I had seen that and thought that I must be reasoning about this incorrectly

so I spoke with @pepin-stripe just now & was informed that the original playback code was written before #574 which is why we had to work around the Cobra CLI interface rather than with it (running the listen command from within the playback command shouldn't be happening with Cobra)

So next step is to undo all of his PR and call the core proxy logic from the playback command rather than the listen command

pkg/cmd/playback.go
```diff
@@ -190,24 +194,79 @@ func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
...
+	}
+
+	key, err := Config.Profile.GetAPIKey(false)
+	if err != nil {
+		return err
+	}
+
+	ctx := withSIGTERMCancel(cmd.Context(), func() {
+		log.WithFields(log.Fields{
+			"prefix": "proxy.Proxy.Run",
```
pepin-stripe, line 212: we prob want to change that, to add context on the fact that this is proxy ran from playback.

"playback.proxy.Run" might be ok.

pkg/cmd/playback.go
```diff
@@ -190,24 +194,79 @@ func (pc *playbackCmd) runPlaybackCmd(cmd *cobra.Command, args []string) error {
...
 }
 
-func startListenCmdLoop(mode string, address string, httpWrapper *playback.Server) {
-	lc := newListenCmd()
-	lc.forwardURL = address + "/playback/webhooks"
-	startListenCmd := func() {
+func runListen(cmd *cobra.Command, address string) error {
+	version.CheckLatestVersion()
+
+	deviceName, err := Config.Profile.GetDeviceName()
```
pepin-stripe, line 200: not a big deal now that we're focused on making this work and building an MVP, we'd probably want to pass something else here to clarify that the listen command is being ran through playback. I think this is used to create a CLI webhook endpoint and is shown in the dev dashboard.

pepin-stripe (approved):
LGTM!

pkg/cmd/playback.go
lemuel-stripe: sounds good I'll switch it to that

pkg/cmd/playback.go
lemuel-stripe: this sounds like a good clarification to add, I was thinking we could just append `-playback` to `deviceName` for now

pkg/playback/replayer.go
```diff
@@ -160,9 +159,8 @@ func (replayer *Replayer) handler(w http.ResponseWriter, r *http.Request) {
 	go func() {
 		// Note: if there are any errors in processing recorded webhooks here,
 		// we log the error and keep going.
-		replayer.replayLock.Wait()       // only send webhooks after the previous request/response is handled
-		replayer.replayLock.Add(1)       // acquire lock so we can send webhooks
-		defer replayer.replayLock.Done() // release lock when done sending webhooks
+		replayer.replayLock.Lock()         // only send webhooks after the previous request/response is handled
+		defer replayer.replayLock.Unlock() // release lock when done sending webhooks
```
bernerd-stripe, line 163: This LGTM, but I think we might also want to explore (as a followup) a more explicit (non-mutex based) way to implement what's essentially a pipeline (receive request -> send response -> send webhooks -> wait for next request).

bernerd-stripe (approved):

pkg/playback/replayer.go
lemuel-stripe: yup I agree! I began some of that exploration & I'll spend some more time on it!

PR #801, opened 2021-12-14
Create requests.MakeMultiPartRequest to allow files upload

pepin-stripe:
### Reviewers
r? @htheodore-stripe 
cc @stripe/developer-products

 ### Summary

This is the integration of @htheodore-stripe 's patch for files upload.


pkg/requests/base.go
```diff
@@ -297,6 +321,40 @@ func (rb *Base) buildDataForRequest(params *RequestParameters) (string, error) {
...
+		splitDatum := strings.SplitN(datum, "=", 2)
+
+		if len(splitDatum) < 2 {
+			return nil, "", fmt.Errorf("Invalid data argument: %s", datum)
+		}
+
+		key := splitDatum[0]
+		val := splitDatum[1]
+
+		if strings.HasPrefix(val, "@") {
```
htheodore-stripe, line 338:
```suggestion
        // Param values that are prefixed with @ will be parsed as a form file
		if strings.HasPrefix(val, "@") {
```

pkg/requests/base.go
```diff
@@ -297,6 +321,40 @@ func (rb *Base) buildDataForRequest(params *RequestParameters) (string, error) {
 	return encode(keys, values), nil
 }
 
+func (rb *Base) buildMultiPartRequest(params *RequestParameters) (*bytes.Buffer, string, error) {
```
htheodore-stripe, line 324: We need some comments to clarify how to use this method.
```suggestion
// buildMultiPartRequest builds a request that can contain mix of files and plain strings.
// Similar to making a multipart request using curl, add the local filepath to params arg with @ prefix.
// e.g. params.AppendData([]string{"photo=@/path/to/local/file.png"})
func (rb *Base) buildMultiPartRequest(params *RequestParameters) (*bytes.Buffer, string, error) {
```

pkg/requests/base.go
```diff
@@ -173,8 +176,31 @@ func (rb *Base) InitFlags() {
 	rb.Cmd.Flags().MarkHidden("api-base") // #nosec G104
 }
 
+// MakeMultiPartRequest will make a multipart/form-data request to the Stripe API with the specific variables given to it
```
htheodore-stripe, line 179:
```suggestion
// MakeMultiPartRequest will make a multipart/form-data request to the Stripe API with the specific variables given to it.
We need some comments to clarify how to use this method.
```suggestion
// Similar to making a multipart request using curl, add the local filepath to params arg with @ prefix.
// e.g. params.AppendData([]string{"photo=@/path/to/local/file.png"})
```
```

File: pkg/requests/base.go [hunk 4]
```diff
@@ -297,6 +321,40 @@ func (rb *Base) buildDataForRequest(params *RequestParameters) (string, error) {
...
+	for _, datum := range params.data {
+		splitDatum := strings.SplitN(datum, "=", 2)
+
+		if len(splitDatum) < 2 {
+			return nil, "", fmt.Errorf("Invalid data argument: %s", datum)
+		}
+
+		key := splitDatum[0]
+		val := splitDatum[1]
+
```
htheodore-stripe, line 337:
```suggestion
// Param values that are prefixed with @ will be parsed as a form file
```

htheodore-stripe (approved):

PR #807, opened 2022-01-17
Create npm-publish.yml

anabc9999:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary
https://github.com/stripe/stripe-cli/commit/d9a8903b6a502e5119c14acd3285e5364c691ac1


tomer-stripe:
hi! I'm going to close this -- I don't think we can directly publish the CLI to npm without some wrappers to drop it into a code base. If you'd want to add that as well that would be awesome :)

PR #812, opened 2022-01-24
Add fixtures to trigger new BT payment intent events

jtb-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary

Adds fixtures to trigger partially_funded and requires_action. This will allow testing of bank transfer payment methods.

Let me know if there's any hygiene changes I should add - I didn't run any codegen or pull the latest API spec etc.


ianjabour-stripe (approved):
LGTM, thanks for adding these! You shouldn't need any codegen, the Stripe CLI will parse the fixture files you created at runtime 👍

PR #819, opened 2022-02-09
grpc service to create new webhook endpoint

etsai-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

## Github Issue
https://github.com/stripe/vscode-stripe/issues/114

Screen Shot 2022-02-08 at 4 32 39 PM
Screen Shot 2022-02-08 at 4 33 02 PM
Screen Shot 2022-02-08 at 4 33 27 PM
Screen Shot 2022-02-09 at 3 14 13 PM
Screen Shot 2022-02-08 at 4 33 46 PM
Screen Shot 2022-02-08 at 4 34 17 PM
Screen Shot 2022-02-08 at 4 40 03 PM
Screen Shot 2022-02-08 at 4 35 06 PM


pkg/requests/webhook_endpoints.go
```diff
@@ -40,3 +42,39 @@ func WebhookEndpointsList(ctx context.Context, baseURL, apiVersion, apiKey strin
...
+		"enabled_events[]=*",
+		fmt.Sprintf("url=%s", url),
+		fmt.Sprintf("api_version=%s", apiVersion),
+	}
+	if description != "" {
+		data = append(data, fmt.Sprintf("description=%s", description))
+	}
+	if connect {
+		data = append(data, "connect=true")
+	}
```
vcheung-stripe, line 62: I wonder if we should use https://pkg.go.dev/net/url#Values.Encode to url-encode this. I see this being used in other parts of the repo.

pkg/rpcservice/webhook_endpoint_create.go
```diff
@@ -0,0 +1,36 @@
...
+
+	key, err := userConfig.Profile.GetAPIKey(livemode)
+	if err != nil {
+		return nil, err
+	}
+
+	err = requests.WebhookEndpointCreate(
+		ctx,
+		stripe.DefaultAPIBaseURL,
+		stripe.APIVersion,
```
vcheung-stripe, line 24: Seems `stripe.APIVersion` is a constant 2019-03-14. I wonder if it makes sense to use empty string instead so that the webhook endpoint uses the merchant default version? That's how the Dashboard works today, so it could be less surprising I think.

vcheung-stripe (commented):
Looks great! I just had a few questions, but mainly I want to be considerate about what API version we're going to use to create the endpoints.

pkg/requests/webhook_endpoints.go
```diff
@@ -40,3 +42,39 @@ func WebhookEndpointsList(ctx context.Context, baseURL, apiVersion, apiKey strin
...
 }
+
+// WebhookEndpointCreate creates a new webhook endpoint
+func WebhookEndpointCreate(ctx context.Context, baseURL, apiVersion, apiKey, url, description string, connect bool, profile *config.Profile) error {
+	if strings.TrimSpace(url) == "" {
+		return fmt.Errorf("url cannot be empty")
+	}
+
+	data := []string{
+		"enabled_events[]=*",
```
vcheung-stripe, line 53: Also one more thing: Is there any way to track what endpoints were created via this gRPC call? Using `enabled_events[]=*` is sort an anti-pattern, but I think it's okay to do it here while the user is developing. But some insights into how often these endpoints are being used could be helpful for us.

pkg/rpcservice/webhook_endpoint_create.go
```diff
@@ -0,0 +1,36 @@
...
+		req.Url,
+		req.Description,
+		req.Connect,
+		&userConfig.Profile,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &rpc.WebhookEndpointCreateResponse{}, nil
```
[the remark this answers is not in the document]
etsai-stripe: i thought about it but i didnt really need it in my usecase, since vscode would make a refresh call to retrieve the updated list

pkg/requests/webhook_endpoints.go
etsai-stripe: yes there is telemetry for creating endpoints on the vscode side

pkg/requests/webhook_endpoints.go
etsai-stripe: do you mean examples like this? https://github.com/stripe/stripe-cli/blob/master/pkg/login/poll.go#L55
in that case, `client.PerformRequest` takes params as a string. 
im using `base.MakeRequest`, which takes `RequestParameters`. it will transform the params and call `client.PerformRequest` in the end

pkg/requests/webhook_endpoints.go
```diff
@@ -40,3 +42,39 @@ func WebhookEndpointsList(ctx context.Context, baseURL, apiVersion, apiKey strin
...
+// WebhookEndpointCreate creates a new webhook endpoint
+func WebhookEndpointCreate(ctx context.Context, baseURL, apiVersion, apiKey, url, description string, connect bool, profile *config.Profile) error {
+	if strings.TrimSpace(url) == "" {
+		return fmt.Errorf("url cannot be empty")
+	}
+
+	data := []string{
+		fmt.Sprintf("url=%s", url),
+		"enabled_events[]=*",
+		"api_version=''", // empty string so the new webhook endpoint uses the merchant default version
```
vcheung-stripe, line 55: I just tried this out with curl, and It seems like `api_version=""` may not work. OTOH, omitting `api_version` seems to work. Maybe we could have the caller pass an empty string, and this function check `if apiVersion != ""` like we're doing for `description`?

etsai-stripe: i think we can leave it out entirely, because it doesnt make sense for users to create a webhook endpoint for an api version thats different from their merchant account (especially thru vscode ext)

vcheung-stripe: sgtm!

etsai-stripe: i justed tested creating a new webhook endpoint without specifyin the api version. the new endpoint is created with my merchant api

vcheung-stripe (approved):
LGTM!

PR #820, opened 2022-02-15
Replace exec package with execabs

vcheung-stripe:
### Reviewers
r? @stripe/developer-products
cc @stripe/developer-products

 ### Summary

This replacement package always looks for the executable in an absolute path, and throws an error if the executable is in a relative path. Note that this only affects the behavior of certain commands on Windows; other operating systems are unaffected.

Internally, see [here](https://jira.corp.stripe.com/browse/RUN_DX-772) for more info.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -5,11 +5,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
+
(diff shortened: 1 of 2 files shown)


etsai-stripe (approved):

etsai-stripe:
is throwing the error more explicitly the intention here?

vcheung-stripe:
> is throwing the error more explicitly the intention here?

Yeah. The executables we want to run are in a Windows system folder, and we have no reason to run executables outside of that. If that ever happens, then we should throw an error, which this package does.

PR #821, opened 2022-02-16
Add `invoice.paid` to triggers

vcheung-stripe:
### Reviewers
r? @stripe/developer-products
cc @stripe/developer-products

 ### Summary

Add `stripe trigger invoice.paid`.

(Note that this event type is similar to `invoice.payment_succeeded`, except `invoice.paid` is also emitted for out-of-band payments. Due to limitations with fixtures, this means this fixture file is actually identical to the one for `invoice.payment_succeeded`.)


ianjabour-stripe (approved):
LGTM - tested out the fixture and works as expected

PR #822, opened 2022-02-16
Fix --forward-connect-to not forwarding events

vcheung-stripe:
### Reviewers
r? @stripe/developer-products
cc @stripe/developer-products

 ### Summary

Fix a bug where running `stripe listen` with `--forward-connect-to` would not forward events.

#### Before:

Using `--forward-to <url>` only:
✅ direct events are forwarded to `<url>`
✅ connect events are forwarded to `<url>`

Using both `--forward-to <url 1>` and `--forward-connect-to <url 2>`:
✅ direct events are forwarded to `<url 1>`
✅ connect events are forwarded to `<url 2>`

Using `--forward-connect-to <url>` only:
✅ direct events are NOT forwarded to `<url>`
❌ connect events are NOT forwarded to `<url>`

#### After:

Using `--forward-to <url>` only:
✅ direct events are forwarded to `<url>`
✅ connect events are forwarded to `<url>`

Using both `--forward-to <url 1>` and `--forward-connect-to <url 2>`:
✅ direct events are forwarded to `<url 1>`
✅ connect events are forwarded to `<url 2>`

Using `--forward-connect-to <url>` only:
✅ direct events are NOT forwarded to `<url>`
✅ connect events are forwarded to `<url>`

I added unit tests for each of these and tested manually as well.

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -493,22 +493,26 @@ func Init(ctx context.Context, cfg *Config) (*Proxy, error) {
 		if err != nil {
 			return nil, err
 		}
-	} else if len(cfg.ForwardURL) > 0 {
-		// non-connect endpoints
-		endpointRoutes = append(endpointRoutes, EndpointRoute{
-			URL:            parseURL(cfg.ForwardURL),
-			ForwardHeaders: cfg.ForwardHeaders,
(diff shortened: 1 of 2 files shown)


ianjabour-stripe (approved):
💯

PR #844, opened 2022-03-24
plugins refactor

suz-stripe:
### Reviewers
r? @gracegoo-stripe 
cc @stripe/developer-products

 ### Summary
Fixes and improvements:

1. CLI is aware of installed plugins, and lists them as commands in the top level help output
2. Improved error, log output, and debug handling in a few places
3. Opaque URL bug fix
4. Tracing of remote URL fetching
5. Third party spinner package upgrade to address a bug in the package
6. Skip sending event telemetry for plugins and delegate to the plugin itself

### DIFF
--- a/pkg/stripe/analytics_telemetry.go
+++ b/pkg/stripe/analytics_telemetry.go
@@ -231,17 +231,3 @@ func TelemetryOptedOut(optoutVar string) bool {
 
 	return optoutVar == "1" || optoutVar == "true"
 }
-
-// IsPluginCommand returns true if the command invoked is for a plugin
-// flase otherwise
-func IsPluginCommand(cmd *cobra.Command) bool {
-	isPlugin := false
-
-	for key, value := range cmd.Annotations {
-		if key == "scope" && value == "plugin" {
-			isPlugin = true
-		}
-	}
-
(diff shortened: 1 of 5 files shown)


pkg/stripe/analytics_telemetry.go
```diff
@@ -231,3 +231,17 @@ func TelemetryOptedOut(optoutVar string) bool {
 
 	return optoutVar == "1" || optoutVar == "true"
 }
+
+// IsPluginCommand returns true if the command invoked is for a plugin
+// flase otherwise
+func IsPluginCommand(cmd *cobra.Command) bool {
```
gracegoo-stripe, line 237: nit: Would be be better off outside this package? I can see it coming handy for non-telemetry related reasons in the future

gracegoo-stripe (approved):
love it, so much cleaner!! I just have a non-blocking nit so approving

pkg/stripe/analytics_telemetry.go
suz-stripe: this is a great catch and I agree!

gracegoo-stripe (approved):

PR #849, opened 2022-04-06
plugin fixes for windows and root level flags

suz-stripe:
### Reviewers
r? @gracegoo-stripe @etsai-stripe 
cc @stripe/developer-products

 ### Summary
I made an assumption that executables on Windows do not need a file extension in order to run successfully via a golang `exec` child process. This assumption turned out to be incorrect. Adding an urgent temporary patch to fix this with a lighter touch, before deciding on a more permanent solution.

Also modified the plugin cmd template to parse the flags on the cli side but send the raw args to the plugin child, which gets us the best of both worlds wrt flag compatibility.

### DIFF
--- a/pkg/plugins/plugin_test.go
+++ b/pkg/plugins/plugin_test.go
@@ -2,12 +2,22 @@ package plugins
 
 import (
 	"context"
+	"fmt"
 	"os"
+	"runtime"
 	"testing"
 
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/require"
 )
+
+func getBinaryExtension() string {
+	if runtime.GOOS == "windows" {
+		return ".exe"
+	}
+
+	return ""
+}
 
 func TestLookUpLatestVersion(t *testing.T) {
 	fs := setUpFS()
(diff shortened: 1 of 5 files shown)


pkg/plugins/plugin_test.go
```diff
@@ -98,16 +104,18 @@ func TestInstallDoesNotCleanIfInstallFails(t *testing.T) {
 	plugin, _ := LookUpPlugin(context.Background(), config, fs, "appA")
 	err := plugin.Install(context.Background(), config, fs, "2.0.1", testServers.StripeServer.URL)
 	require.Nil(t, err)
-	fileExists, _ := afero.Exists(fs, "/plugins/appA/2.0.1/stripe-cli-app-a")
+	file := fmt.Sprintf("/plugins/appA/2.0.1/stripe-cli-app-a%s", GetBinaryExtension())
```
etsai-stripe, line 107: should we make these file paths as constants?

gracegoo-stripe (approved):

pkg/plugins/plugin_test.go
suz-stripe: worth considering for sure. I think we should look into this change once we know what the permanent fix this windows issue will be.

PR #853, opened 2022-04-14
stripe.Client supports custom list of printable headers

bernerd-stripe:
### Reviewers
r? @acomley-stripe 
cc @stripe/developer-products

 ### Summary

This updates the stripe Client to make the list of headers printed in verbose mode to be configurable. This is not exposed through CLI flags, but is useful for plugins or potential usage of the CLI's stripe client package.

It also applies some light refactoring to use the `http.Transport` interface instead of `*http.Transport`.

### DIFF
--- a/pkg/stripe/client.go
+++ b/pkg/stripe/client.go
@@ -41,6 +41,10 @@ type Client struct {
 	// stdout.
 	Verbose bool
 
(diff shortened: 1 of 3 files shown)


bernerd-stripe:
r-codeowners?

acomley-stripe (approved):

PR #863, opened 2022-05-04
Updating serve dir to be absolute

gracegoo-stripe:
### Reviewers
r? @bernerd-stripe 
cc @stripe/developer-products

 ### Summary
This change converts the directory passed in by the user in `stripe serve` to an absolute directory. Otherwise we get some unintended side effects for windows. https://go.dev/play/p/bI1Rwp8YgxA (Thanks bernerd for the playground!)

In the previous commit, I had a wrapper around the Dir implementation to catch the `"filename, directory name, or volume label syntax is incorrect"` error and return an `fs.ErrInvalid` but that still ends up showing up as a 500 in the logs so I figured it's not really worth having this wrapper because it's not any more friendly of an error message.

# Testing
Tested the command on windows and verified that we get the following error: 
```
Error: open C:\Users\Grace\Documents\repos\stripe-cli\C:: The filename, directory name, or volume label syntax is incorrect.
```


pkg/cmd/serve.go
```diff
@@ -32,9 +33,16 @@ func newServeCmd() *serveCmd {
...
+
+			fmt.Printf("Starting server for directory  %s\n", absoluteDir)
+
 			fmt.Println("Starting static file server at address", fmt.Sprintf("http://localhost:%s", port))
-			http.Handle("/", http.FileServer(http.Dir(dir)))
-			err := http.ListenAndServe(fmt.Sprintf(":%s", port), handlers.LoggingHandler(os.Stdout, http.DefaultServeMux))
+			http.Handle("/", http.FileServer(http.Dir(absoluteDir)))
+			err = http.ListenAndServe(fmt.Sprintf(":%s", port), handlers.LoggingHandler(os.Stdout, http.DefaultServeMux))
 
 			return err
```
bernerd-stripe, line 47: Since you had to change this line anyway, could be a good opportunity to update to:

```suggestion
			return http.ListenAndServe(fmt.Sprintf(":%s", port), handlers.LoggingHandler(os.Stdout, http.DefaultServeMux))
```

bernerd-stripe (approved):

PR #883, opened 2022-05-31
Update gopkg.in/yaml.v2 to v3

vcheung-stripe:
### Reviewers
r? @gracegoo-stripe 
cc @stripe/developer-products

 ### Summary

Fix Dependabot alert which says to upgrade gopkg.in/yaml from v2 to v3.
This is a breaking change, however:
- This package is only used by the playback code, which technically isn't publicly usable.
- The only thing that broke after updating was how the YAML output is indented, which doesn't matter. At least this was all that our existing tests caught.

### DIFF
--- a/pkg/playback/server_test.go
+++ b/pkg/playback/server_test.go
@@ -17,8 +17,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-const defaultLocalAddress = "localhost:8080"
-const defaultLocalWebhookAddress = "localhost:8888"
+const defaultLocalAddress = "127.0.0.1:8080"
+const defaultLocalWebhookAddress = "127.0.0.1:8888"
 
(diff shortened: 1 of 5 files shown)


pkg/playback/server_test.go
```diff
@@ -17,8 +17,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-const defaultLocalAddress = "localhost:8080"
-const defaultLocalWebhookAddress = "localhost:8888"
+const defaultLocalAddress = "127.0.0.1:8080"
+const defaultLocalWebhookAddress = "127.0.0.1:8888"
```
vcheung-stripe, line 21: There was a bug that made `make test` fail unless I changed this, see https://github.com/golang/go/issues/41572#issuecomment-700232101

gracegoo-stripe (approved):

PR #895, opened 2022-06-13
remove old bintray scripts (rip)

suz-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary
We haven't used bintray since the service was shut down a while ago, and we now use artifactory as a replacement. So we can safely delete these scripts now 🙂


gracegoo-stripe (approved):
RIP

PR #899, opened 2022-06-21
Add an HTTP wrapper to the gRPC daemon

tomelm:
### Reviewers
r? @vcheung-stripe @suz-stripe 
cc @stripe/developer-products

 ### Summary
I was working on a small side projects where I was trying to have an app talk to the Stripe CLI's gRPC interface. One problem I had is that the request client only supported HTTP which meant that it couldn't interface with the daemon directly. This PR wraps the gRPC daemon and spins up another server that supports all the gRPC functionality as an HTTP server.


ianjabour-stripe (approved):
This LGTM, will defer to @vcheung-stripe being that he's got more context on the gRPC setup.

pkg/rpcservice/middleware.go
```diff
@@ -43,7 +44,10 @@ func authorize(ctx context.Context) error {
 	}
 
 	if _, ok := md[requiredHeader]; !ok {
-		return status.Errorf(codes.Unauthenticated, fmt.Sprintf("%s header is not supplied", requiredHeader))
+		// Check for the alt header too, which is required for grpc web clients
+		if _, ok := md[altRequiredHeader]; !ok {
+			return status.Errorf(codes.Unauthenticated, fmt.Sprintf("%s header is not supplied", requiredHeader))
```
vcheung-stripe, line 49: nit: do you want to return the alt header in this case?
```suggestion
			return status.Errorf(codes.Unauthenticated, fmt.Sprintf("%s header is not supplied", altRequiredHeader))
```

vcheung-stripe (approved):
LGTM 🎉

pkg/rpcservice/middleware.go
tomer-stripe: yep, good catch!

PR #900, opened 2022-06-21
Verify resource command does not conflict with plugin command

etsai-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

Add a test case to verify newly updated/added resource command does not conflict with existing plugin commands

### DIFF
--- a/pkg/cmd/resources.go
+++ b/pkg/cmd/resources.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	plugin_manifest_url = "https://stripe.jfrog.io/artifactory/stripe-cli-plugins-local/plugins.toml"
+	pluginManifestUrl = "https://stripe.jfrog.io/artifactory/stripe-cli-plugins-local/plugins.toml"
 )
 
 type resourcesCmd struct {
@@ -68,7 +68,7 @@ Use "stripe [command] --help" for more information about a command.
 func getAllPluginCommands() ([]string, error) {
 	var manifest PluginList
 
-	resp, err := http.Get(plugin_manifest_url)
+	resp, err := http.Get(pluginManifestUrl)
 	if err != nil {
 		return []string{}, err
 	}
--- a/pkg/cmd/resources.go
+++ b/pkg/cmd/resources.go
@@ -2,43 +2,15 @@ package cmd
 
 import (
 	"fmt"
-	"io/ioutil"
-	"net/http"
 
-	"github.com/pelletier/go-toml"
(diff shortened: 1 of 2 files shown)


etsai-stripe:
@suz-stripe this will not catch the reverse case though.. ie. new plugin command conflicts with an existing resource command. can we do something similar on the plugin repo?

pkg/cmd/resources.go
```diff
@@ -2,17 +2,45 @@ package cmd
...
+}
+
+// Release is the type that holds release data for a specific build of a plugin
+type Release struct {
+	Arch    string
+	OS      string
+	Version string
+	Sum     string
+}
+
```
suz-stripe, line 43: ~~Are we able to import from the plugins package in the `resources_test.go` file instead to make use of the same types defined in there? Then we can remove these. What do you think?~~

See my note about `getPluginCommands` instead 😃

pkg/cmd/resources.go
```diff
@@ -2,17 +2,45 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
+	"net/http"
 
+	"github.com/pelletier/go-toml"
```
suz-stripe, line 8: did we switch toml packages recently? We normally use `github.com/BurntSushi/toml` IIRC

pkg/cmd/resources.go
```diff
@@ -36,3 +64,30 @@ Use "stripe [command] --help" for more information about a command.
...
+		return []string{}, err
+	}
+
+	var pluginCommands []string
+	for _, plugin := range manifest.Plugins {
+		pluginCommands = append(pluginCommands, plugin.Shortname)
+	}
+
+	return pluginCommands, nil
+}
```
suz-stripe, line 93: this feels like it belongs in the plugins package as a public function in `utilities.go`, and then `resources_test.go` can import and use this as a public function. This means we can also not worry about defining the plugin structs in this file nor the test file.

suz-stripe (commented):
Left some feedback, mostly around location of the code outside of the test file :)

suz-stripe:
> @suz-stripe this will not catch the reverse case though.. ie. new plugin command conflicts with an existing resource command. can we do something similar on the plugin repo?

I think this is a good idea if you can swing it!

pkg/cmd/resources.go
etsai-stripe: i will switch it back

suz-stripe (approved):

PR #901, opened 2022-06-22
adding user configurable read timeouts

joyluu-stripe:
### Reviewers
r? @etsai-stripe 
cc @stripe/developer-products

 ### Summary
Allows user to set timeout in config file using a hidden flag (ex. stripe listen --timeout xx); if not set, defaults to 30 seconds as it was before.

Screen Shot 2022-06-27 at 12 13 34 PM

### DIFF
--- a/pkg/cmd/playback.go
+++ b/pkg/cmd/playback.go
@@ -263,7 +263,6 @@ func runListen(cmd *cobra.Command, address string, wg *sync.WaitGroup) error {
 		SkipVerify:            false,
 		Log:                   logger,
 		NoWSS:                 false,
-		Timeout:               30,
 		Events:                []string{"*"},
 		OutCh:                 proxyOutCh,
 	})
--- a/pkg/cmd/playback.go
+++ b/pkg/cmd/playback.go
@@ -263,6 +263,7 @@ func runListen(cmd *cobra.Command, address string, wg *sync.WaitGroup) error {
 		SkipVerify:            false,
 		Log:                   logger,
 		NoWSS:                 false,
+		Timeout:               30,
 		Events:                []string{"*"},
 		OutCh:                 proxyOutCh,
 	})
--- a/pkg/proxy/endpoint.go
+++ b/pkg/proxy/endpoint.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"regexp"
 	"strings"
-	"time"
 
 	log "github.com/sirupsen/logrus"
 
@@ -22,6 +21,8 @@ type EndpointConfig struct {
 	HTTPClient *http.Client
 
 	Log *log.Logger
+
+	Timeout int64
 
 	ResponseHandler EndpointResponseHandler
 
@@ -142,7 +143,7 @@ func NewEndpointClient(url string, headers []string, connect bool, events []stri
 			CheckRedirect: func(req *http.Request, via []*http.Request) error {
 				return http.ErrUseLastResponse
 			},
-			Timeout: defaultTimeout,
+			Timeout: cfg.HTTPClient.Timeout, // changed this
 		}
 	}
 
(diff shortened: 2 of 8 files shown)


etsai-stripe (changes requested):
we can only need `Timeout` stored in either `Config` or `Profile` (`Profile` is already a field in `Config`). `Config` makes more sense imo, since `Profile` is more user account related values.
what we need is something like
```
func (c *Config) GetTimeout() int64 {
	// return customized timeout or default timeout
}
```

pkg/proxy/proxy.go
```diff
@@ -535,12 +538,13 @@ func Init(ctx context.Context, cfg *Config) (*Proxy, error) {
...
 						return http.ErrUseLastResponse
 					},
-					Timeout: defaultTimeout,
+					Timeout: time.Duration(cfg.Timeout) * time.Second,
 					Transport: &http.Transport{
 						TLSClientConfig: &tls.Config{InsecureSkipVerify: cfg.SkipVerify},
 					},
 				},
 				Log:             p.cfg.Log,
+				Timeout:         cfg.Timeout,
```
etsai-stripe, line 547: this is not needed. timeout is already defined in `HttpClient`

pkg/proxy/endpoint.go
```diff
@@ -23,6 +23,8 @@ type EndpointConfig struct {
 
 	Log *log.Logger
 
+	Timeout int64
```
etsai-stripe, line 26: `Timeout` already exists in `HttpClient`

joyluu-stripe: 👍 removed!

joyluu-stripe:
> we can only need `Timeout` stored in either `Config` or `Profile` (`Profile` is already a field in `Config`). `Config` makes more sense imo, since `Profile` is more user account related values. what we need is something like
> 
> ```
> func (c *Config) GetTimeout() int64 {
> 	// return customized timeout or default timeout
> }
> ```

gotcha, made that change! I still kept the get function in `Profile` since the color config seems to do the same thing (hope that's okay!) also realized that we can keep default timeout for cases where no httpClient is passed in, since that's where we pass timeout through.

etsai-stripe (changes requested):
ohh now i see why you tried to add `timeout` to profile. if you want to follow the way `color` is set on `profile`, lets keep everything consistent. 
1. we dont need `Config.GetTimeout`. we can just keep `Profile.GetTimeout` (sorry for the confusion earlier), but `timeout` field should still live only on config struct like how `color` does.
2. we dont need `timeout` added to `proxy.Config`. we can pass `profile.GetTimeout` to the `HttpClient`
3. add `timeout` flag to rootcmd. see example here: https://github.com/stripe/stripe-cli/blob/master/pkg/cmd/root.go#L141

pkg/cmd/playback.go
```diff
@@ -263,6 +263,7 @@ func runListen(cmd *cobra.Command, address string, wg *sync.WaitGroup) error {
 		SkipVerify:            false,
 		Log:                   logger,
 		NoWSS:                 false,
+		Timeout:               30,
```
joyluu-stripe, line 266: doesn't seem like there's a need to add the private flag to here actually? it seems like there's only one call to `proxy.init` and most paramaters are getting passed default values

pkg/proxy/proxy.go
```diff
@@ -102,6 +102,8 @@ type Config struct {
 	Log *log.Logger
 	// Force use of unencrypted ws:// protocol instead of wss://
 	NoWSS bool
+	// Override default timeout
+	Timeout int64
```
etsai-stripe, line 106: could you please explain why we need to pass this in proxy.config?

joyluu-stripe: ah, i was thinking it would be more consistent with how the other hidden flags' values are passed (like noWSS and apiBaseUrl are passed this way into proxy as well).

PR #902, opened 2022-06-27
run stripe login with project name flag if config not found

joyluu-stripe:
### Reviewers
r? @etsai-stripe 
cc @stripe/developer-products

 ### Summary
When running command with the `--project-name=...` flag, the CLI would run `stripe login...` like so:

Screen Shot 2022-06-24 at 11 13 42 AM

Changes this message so that it instead asks the user to run `stripe login --project-name...`:

Screen Shot 2022-06-27 at 12 35 32 PM

The same message should show up when the shorthand flag `-p` is used, and when multiple flags are used in the command:

Screen Shot 2022-06-29 at 6 10 53 PM

Screen Shot 2022-06-29 at 6 10 36 PM

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -93,6 +93,7 @@ func showSuggestion() {
 func Execute(ctx context.Context) {
 	telemetryMetadata := stripe.NewEventMetadata()
 	updatedCtx := stripe.WithEventMetadata(ctx, telemetryMetadata)
+	userInput := os.Args[1:]
 
 	rootCmd.SetUsageTemplate(getUsageTemplate())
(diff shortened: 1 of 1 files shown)


pkg/cmd/root.go
```diff
@@ -125,12 +128,9 @@ func Execute(ctx context.Context) {
 		}
 
 		os.Exit(1)
-	} else {
```
joyluu-stripe, line 128: i think moving the `userInput` (so i could reuse it) created a lint error for this; since it's just a minor structural change i think it's ok?

etsai-stripe (approved):

PR #903, opened 2022-06-30
add api version header to fixtures

joyluu-stripe:
### Reviewers
r? @etsai-stripe 
cc @stripe/developer-products

 ### Summary

Introducing an `--api-version` flag users can use to set the API version to be used when a fixture/trigger is executed.

using this test fixture to test if specified API Version is correct:

```
{
    "_meta": {
      "template_version": 0
    },
    "fixtures": [
      {
        "name": "cus_jenny_rosen",
        "path": "/v1/customers",
        "method": "post",
        "params": {
          "name": "Jenny Rosen",
          "email": "jenny@rosen.com",
          "source": "tok_visa",
          "address": {
            "line1": "1 Main Street",
            "city": "New York"
          }
      },
      {
        "name": "ch_jenny_charge",
        "path": "/v1/charges",
        "method": "post",
        "params": {
          "customer": "${cus_jenny_rosen:id}",
          "amount": 100,
          "currency": "usd",
          "capture": false
      },
      {
        "name": "capt_bender",
        "path": "/v1/charges/${ch_jenny_charge:id}/capture",
        "method": "post"
      }
    ]
  }
```

Tested by running both `trigger` and `fixture` commands and seeing if logs show the version set by flag:

# Fixture
Screen Shot 2022-06-30 at 3 49 15 PM
Screen Shot 2022-06-30 at 3 55 33 PM

# Trigger
Screen Shot 2022-06-30 at 3 54 31 PM
Screen Shot 2022-06-30 at 3 55 09 PM

Running either without the flag results in using the default version the user has set in their account API settings:

Screen Shot 2022-06-30 at 4 01 22 PM
Screen Shot 2022-06-30 at 4 09 16 PM


pkg/cmd/fixtures.go
```diff
@@ -73,7 +75,7 @@ func (fc *FixturesCmd) runFixturesCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	_, err = fixture.Execute(cmd.Context())
+	_, err = fixture.Execute(cmd.Context(), fc.apiVersion)
```
joyluu-stripe, line 78: could also pass `api-version` flag to `NewFixtureFromFile` like how the other flags are, but it made less sense as version gets set/changed within execute. if it would be better off in the other function though, lmk!

pkg/cmd/trigger.go
```diff
@@ -78,7 +80,7 @@ func (tc *triggerCmd) runTriggerCmd(cmd *cobra.Command, args []string) error {
 
 	event := args[0]
 
-	_, err = fixtures.Trigger(cmd.Context(), event, tc.stripeAccount, tc.apiBaseURL, apiKey, tc.skip, tc.override, tc.add, tc.remove, tc.raw)
+	_, err = fixtures.Trigger(cmd.Context(), event, tc.stripeAccount, tc.apiBaseURL, apiKey, tc.skip, tc.override, tc.add, tc.remove, tc.raw, tc.apiVersion)
```
joyluu-stripe, line 83: `apiVersion` is passed through here because `Trigger` also calls `fixture.execute`

pkg/cmd/fixtures.go
etsai-stripe: i agree it is better to be passed into the execution method directly

pkg/fixtures/fixtures.go
```diff
@@ -315,6 +315,7 @@ func (fxt *Fixture) makeRequest(ctx context.Context, data fixture) ([]byte, erro
 	}
 
 	params, err := fxt.createParams(data.Params)
+	params.SetVersion(apiVersion)
```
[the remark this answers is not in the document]
joyluu-stripe: ooh, i'll do that. the empty string check slipped my mind too, thank you for bringing it up!

etsai-stripe (approved):

PR #918, opened 2022-07-21
DX-7509: redact livemode values in existing configs

etsai-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

- redact livemode values from existing configs
- rename publishable keys to fix the `stripe config --list --project-name <project_name>` issue; viper has a limit on the key name length
- return `login is required` error when livemode keys are not present in the secure storage. this will force users to log in even when they have old livemode key values in existing configs
- add ability to remove livemode values from secure storage

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -166,6 +166,24 @@ func (c *Config) InitConfig() {
 	KeyRing, _ = keyring.Open(keyring.Config{
 		ServiceName: "Stripe CLI Key Storage",
 	})
+
+	// redact livemode values for existing configs
+	if err := viper.ReadInConfig(); err == nil {
+		// if the config file has expires at date, then it is using the new livemode key storage
+		if viper.IsSet(c.Profile.GetConfigField(LiveModeAPIKeyName)) {
+			key := viper.GetString(c.Profile.GetConfigField(LiveModeAPIKeyName))
+			if !IsRedactedAPIKey(key) {
+				c.Profile.WriteConfigField(LiveModeAPIKeyName, RedactAPIKey(key))
+			}
+		}
+
+		if viper.IsSet(c.Profile.GetConfigField(LiveModePubKeyName)) {
+			key := viper.GetString(c.Profile.GetConfigField(LiveModePubKeyName))
+			if !IsRedactedAPIKey(key) {
+				c.Profile.WriteConfigField(LiveModePubKeyName, RedactAPIKey(key))
+			}
+		}
+	}
 }
 
 // EditConfig opens the configuration file in the default editor.
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -33,10 +33,10 @@ const (
 	DisplayNameName            = "display_name"
 	IsTermsAcceptanceValidName = "is_terms_acceptance_valid"
 	TestModeAPIKeyName         = "test_mode_api_key"
-	TestModePublishableKeyName = "test_mode_publishable_key"
+	TestModePubKeyName         = "test_mode_pub_key"
 	TestModeKeyExpiresAtName   = "test_mode_key_expires_at"
 	LiveModeAPIKeyName         = "live_mode_api_key"
-	LiveModePublishableKeyName = "live_mode_publishable_key"
+	LiveModePubKeyName         = "live_mode_pub_key"
 	LiveModeKeyExpiresAtName   = "live_mode_key_expires_at"
 )
 
@@ -122,32 +122,34 @@ func (p *Profile) GetAPIKey(livemode bool) (string, error) {
 		return p.APIKey, nil
 	}
 
-	// If the user doesn't have an api_key field set, they might be using an
-	// old configuration so try to read from secret_key
-	if !livemode {
-		if !viper.IsSet(p.GetConfigField("api_key")) {
-			p.RegisterAlias("api_key", "secret_key")
-		} else {
-			p.RegisterAlias(TestModeAPIKeyName, "api_key")
-		}
-	}
+	var key string
+	var err error
 
 	// Try to fetch the API key from the configuration file
-	if err := viper.ReadInConfig(); err == nil {
-		var key string
-		fieldID := livemodeKeyField(livemode)
-
-		if !livemode {
-			key = viper.GetString(p.GetConfigField(fieldID))
-		} else {
-			key = p.RetrieveLivemodeValue(fieldID)
+	if !livemode {
+		// If the user doesn't have an api_key field set, they might be using an
+		// old configuration so try to read from secret_key
+		if viper.IsSet(p.GetConfigField("secret_key")) {
+			p.RegisterAlias(TestModeAPIKeyName, "secret_key")
+		} else if viper.IsSet(p.GetConfigField("api_key")) {
+			p.RegisterAlias(TestModeAPIKeyName, "api_key")
 		}
 
-		err := validators.APIKey(key)
+		if err := viper.ReadInConfig(); err == nil {
+			key = viper.GetString(p.GetConfigField(TestModeAPIKeyName))
+		}
+	} else {
+		key, err = p.RetrieveLivemodeValue(LiveModeAPIKeyName)
 		if err != nil {
 			return "", err
 		}
+	}
 
+	if key != "" {
+		err = validators.APIKey(key)
+		if err != nil {
+			return "", err
+		}
 		return key, nil
 	}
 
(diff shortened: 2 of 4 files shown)


vcheung-stripe (approved):
LGTM

Istador:
This broke my existing setup using the `latest` docker image (`docker  run  --rm  -it  -v stripe-config://root/.config/stripe/  stripe/stripe-cli:latest  $command`), where I don't have access to the keyring of the host system.

There's no keyring/-chain dialog popping up (like shown in issue #916), but it just redacts the existing live key from the config file:
```
(!) Livemode value found for the field 'live_mode_api_key' in your config file.
Livemode values from the config file will be redacted and will not be used.
No directory provided for file keyring
```

Using `login` again will redact the new key for the config silently without a warning.

I'm using `v1.10.3` instead of the latest `v1.10.4` from now on to mitigate this.

How should I proceed? What do I need to do/change to fix this?

etsai-stripe:
@Istador apologies for the inconvenience. we want to disallow livemode key storage in plain text form entirely. if you would like to use your livemode key without a credential management app, you will need to use the `--api-key` flag for each request

Istador:
> @Istador apologies for the inconvenience. we want to disallow livemode key storage in plain text form entirely. if you would like to use your livemode key without a credential management app, you will need to use the `--api-key` flag for each request

I do understand the reasoning behind the change and why this is preferable from a security standpoint.

But please consider installing a keyring software into the docker image you provide that works out of the box (without a UI). Or provide documentation on how to connect/expose the hosts keyring to the docker container.

Using `--api-key` manually is a way less desirable/convenient/secure alternative. I'd need to manually create/copy the live key around, store it somewhere else, and develop a script around it to add it to the stipe commands automatically, based on if it's for live or test.

It'd also be nice if breaking changes aren't hidden behind patch version numbers.

etsai-stripe:
@Istador all these are very good points. i will look into updating the docker image. thanks!

etsai-stripe:
@Istador  could you please try `apt-get install pass` in your docker?

Istador:
I could try to build and use this on Monday to see if it works:
```Dockerfile
FROM  stripe/stripe-cli:v1.10.4
RUN  apk  add  pass
```

Istador:
Dockerfile:
```Dockerfile
FROM  stripe/stripe-cli:v1.10.4
RUN  apk  add  pass  gpg-agent
COPY  ./entrypoint.sh  /entrypoint.sh
ENTRYPOINT  [ "/entrypoint.sh" ]
```

entrypoint.sh:
```sh
#!/bin/sh
if ! [ -f ~/.gnupg/trustdb.gpg ] ; then
  chmod 700 ~/.gnupg/
  gpg --quick-generate-key stripe-live
fi
if ! [ -f ~/.password-store/.gpg-id ] ; then
  pass init stripe-live
fi
/bin/stripe  $@
```

This entrypoint will create a GPG key the first time it is used (prompts user for a password on the cli).

```
docker build . -t stripe-test
docker run --rm -it -v stripe-config://root/.config/stripe/ -v stripe-gpg://root/.gnupg/ -v stripe-pass://root/.password-store/ stripe-test $command
```

---

But I can't get it to work.

Calling `pass show default.live_mode_api_key` manual in the entrypoint does work (asks me for a password and successfully decrypts the value put there by the stripe-cli).

But the stripe-cli always outputs `gpg: decryption failed: No secret key` for me and asks me to login again.

etsai-stripe:
@Istador 
i dont quite understand how the gpg key is required for `pass` to work. does that have to be added to the docker?
that error message usually means the key pairs dont match. (https://linuxhint.com/solve-gpg-decryption-failed-no-secret-key-error/#:~:text=Your%20public%20key's%20decryption%20key,related%20key%20on%20your%20system.)

with my lack of understanding of how gpg is required by stripe cli, may i naively suggest you generate the gpg key pair locally and copy them into your docker?

Istador:
> i dont quite understand how the gpg key is required for `pass` to work. does that have to be added to the docker?

Only adding `pass` to the image will result in an error, complaining that `pass` needs to be initialized first:
```
Error: You must run:
    pass init your-gpg-id
before you may use the password store.
```

https://linux.die.net/man/1/pass :
> pass is a very simple password store that keeps passwords inside [gpg2](https://linux.die.net/man/1/gpg2)(1) encrypted files inside a simple directory tree residing at ~/.password-store.

The `gpg-agent` is required to actually ask on the cli for the password of the gpg key that is used for encrypting/decrypting the secrets managed by `pass`.

---

> you generate the gpg key pair locally and copy them into your docker?

I haven't tested it, but that should produce the same error in my understanding of it.

etsai-stripe:
@Istador 
i found this article for initializing pass with gpg key. could you please try:
- create one: https://gist.github.com/flbuddymooreiv/a4f24da7e0c3552942ff#create-a-gpg-key
- and then initiate `pass` with the gpg id created from the prev step: https://gist.github.com/flbuddymooreiv/a4f24da7e0c3552942ff#create-a-pass-database

Istador:
> @Istador i found this article for initializing pass with gpg key. could you please try:
> 
> * create a new gpg key: https://gist.github.com/flbuddymooreiv/a4f24da7e0c3552942ff#create-a-gpg-key
> * and then initiate `pass` with the gpg id created from the prev step: https://gist.github.com/flbuddymooreiv/a4f24da7e0c3552942ff#create-a-pass-database

That's exactly what I do/did in the entrypoint. I generate a new `gpg` key and use it to initialize `pass`.

I also tried `--gen-key` instead of `--quick-generate-key` like in the guide. But don't in the entrypoint anymore, because `--quick-generate-key` is simpler and requires less user interaction. It generates a public and private key with a passphrase like `--gen-key` does (using defaults).

I also first implemented a key ID extraction, by `grep`ing the output of `gpg`, to give `pass init` the key ID for the key that was generated. But then I figured out that giving the gpg key an alias instead (`stripe-live`) and passing that one to `pass` is also possible and way simpler to program.

Using `--gen-key` instead of `--quick-generate-key` and init `pass` with an key ID instead of the alias results in the same `gpg: decryption failed: No secret key` error when decrypting the stored secret.

How I do the `gpg` and `pass` setup in the entrypoint are correct in my opinion. Because 1. the `stripe-cli` is able to encrypt the data but fails to decrypt it again and 2. calling `pass` manually to decrypt the file works.

---

I suspect it's an issue with the `stripe-cli` being unable to spawn/display the `gpg-agent` to enter the password on the same tty it is using/occupying. Because in a GUI environment the password prompt to decrypt happens outside of the terminal. (RSA encryption uses the public key that isn't guarded with a password, so that works without a password prompt)

Something I might try out tomorrow to test/verify this: use `pass` to decrypt the secret (which asks for the password) before calling `/bin/stripe` in the entrypoint. That way the `gpg-agent` is able to cache the key and doesn't need to prompt for an password later when the `stripe-cli` runs.

Istador:
> Something I might try out tomorrow to test/verify this: use pass to decrypt the secret (which asks for the password) before calling /bin/stripe in the entrypoint. That way the gpg-agent is able to cache the key and doesn't need to prompt for an password later when the stripe-cli runs.

Ok, that actually worked.

`entrypoint.sh`:
```
[... like before ...]
pass show default.live_mode_api_key >/dev/null
/bin/stripe  $@
```

Calling `pass show default.live_mode_api_key` before the call to stripe will ask for the password using `gpg-agent` and puts the key in the cache. When the stripe cli then later want to access the `gpg` key it doesn't fail with `gpg: decryption failed: No secret key` anymore, but can access the key and use it to request live data.

So I highly suspect that to be an issue with the `stripe-cli` not being able to prompt for the `gpg` password on the same shell/terminal it is running on.

etsai-stripe:
@Istador thank you so much for the patience and working through the issue with us! i will add your method in the README for docker users.

etsai-stripe:
Screen Shot 2022-08-18 at 2 37 10 PM
@Istador does it prompt you this for every stripe cli command you make? even non-live ones?

my `entrypoint.sh`:

```
#!/bin/sh
if ! [ -f ~/.gnupg/trustdb.gpg ] ; then
  chmod 700 ~/.gnupg/
  gpg --quick-generate-key stripe-live
fi
if ! [ -f ~/.password-store/.gpg-id ] ; then
  pass init stripe-live
fi
pass show default.live_mode_api_key >/dev/null
/bin/stripe  $@
```

Istador:
> @Istador does it prompt you this for every stripe cli command you make because of pass show default.live_mode_api_key >/dev/null ? even non-live ones?

Yes, exactly. Because currently in the entrypoint it's called everytime without any logic. E.g. a better way would be to only execute it if the arguments contain `--live` in them.

But adding the `pass show` command there was just to verify that it's probably an issue opening the passphrase prompt from within the following stripe program. (Without the `pass show` the expected behavior would be to be prompted when the stripe cli decides to read the stored secret, for which it needs the passphrase to use the gpg private key, but instead it fails with the `gpg: decryption failed: No secret key` error immediately without a prompt.

PR #919, opened 2022-07-22
show warning for old livemode values

etsai-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -184,31 +184,31 @@ func (p *Profile) GetExpiresAt(livemode bool) (time.Time, error) {
 
 // GetPublishableKey returns the publishable key for the user
 func (p *Profile) GetPublishableKey(livemode bool) (string, error) {
+	var fieldID string
 	var key string
-	var err error
 
 	if livemode {
-		p.redactAllLivemodeValues()
-		key, err = p.retrieveLivemodeValue(LiveModePubKeyName)
-		if err != nil {
-			return "", err
-		}
+		fieldID = LiveModePubKeyName
 	} else {
-		// test mode
-		if err := viper.ReadInConfig(); err == nil {
-			if viper.IsSet(p.GetConfigField("publishable_key")) {
-				p.RegisterAlias(TestModePubKeyName, "publishable_key")
-			}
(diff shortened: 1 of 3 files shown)


pkg/config/profile_livemode.go
```diff
@@ -62,6 +66,25 @@ func (p *Profile) DeleteLivemodeValue(key string) error {
...
+	if err := viper.ReadInConfig(); err == nil {
+		// if the config file has expires at date, then it is using the new livemode key storage
+		if viper.IsSet(p.GetConfigField(LiveModeAPIKeyName)) {
+			key := viper.GetString(p.GetConfigField(LiveModeAPIKeyName))
+			if !isRedactedAPIKey(key) {
+				fmt.Println(color.Yellow(`
+(!) Livemode value found for the field '` + LiveModeAPIKeyName + `' in your config file.
+Livemode values from the config file will be redacted and will not be used.`))
+
+				p.WriteConfigField(LiveModeAPIKeyName, RedactAPIKey(key))
```
vcheung-stripe, line 82: Should we redact LiveModePubKeyName too?

pkg/config/profile.go
```diff
@@ -139,7 +139,8 @@ func (p *Profile) GetAPIKey(livemode bool) (string, error) {
 			key = viper.GetString(p.GetConfigField(TestModeAPIKeyName))
 		}
 	} else {
-		key, err = p.RetrieveLivemodeValue(LiveModeAPIKeyName)
+		p.redactAllLivemodeValues()
```
vcheung-stripe, line 142: Would the values already be redacted by this point? Is this call defensive?

vcheung-stripe (approved):
LGTM! Just had a couple questions for clarity.

pkg/config/profile_livemode.go
etsai-stripe: i did that at first but tomer said we should leave out publishable key since its public facing, which makes sense

pkg/config/profile.go
etsai-stripe: yes its a defensive check in case user manually edits the config tom file again

PR #921, opened 2022-07-25
More detailed API key error

tomelm:
Noticed that we were swallowing config errors when returning API keys if something on the config-level was busted. We probably want that for general usage but I made this configurable depending on the log-level.

### DIFF
--- a/pkg/config/profile_test.go
+++ b/pkg/config/profile_test.go
@@ -120,7 +120,7 @@ func TestAPIKeyLogLevel(t *testing.T) {
 	// For debug mode, the error should complain about a config file missing
 	// since we did not init the config
 	key, err := c.Profile.GetAPIKey(false)
-	assert.EqualError(t, err, `Config File "config" Not Found in "[]"`)
+	assert.ErrorContains(t, err, `stripe/config.toml: no such file or directory`)
 	assert.Equal(t, "", key)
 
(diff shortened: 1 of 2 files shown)


gracegoo-stripe (approved):

pkg/config/profile_test.go
```diff
@@ -101,6 +104,43 @@ test_mode_key_expires_at = '` + expiresAt + `'
...
+	// since we did not init the config
+	key, err := c.Profile.GetAPIKey(false)
+
+	// Little weird but windows returns a different error message than others
+	os := runtime.GOOS
+	switch os {
+	case "windows":
+		assert.ErrorContains(t, err, `config.toml: The system cannot find the file specified.`)
+	default:
+		assert.ErrorContains(t, err, `config.toml: no such file or directory`)
```
mnorth-stripe, line 131: Sometimes these error messages change or vary (e.g., different linux distros) and I'd like us to have a very obvious and actionable signal if/when that happens.

Can we have an assertion somewhere that alerts us if neither of the two cases we know about today are met?

mnorth-stripe (changes requested):
Thanks for making our CLI errors more actionable!

I think you could add one more assertion to the test and then it's ready to ship

tomer-stripe:
Closing this, very old

PR #928, opened 2022-08-03
RUN_DX-1388: append query from webhook endpoint if exists

etsai-stripe:
### Reviewers
r? @gracegoo-stripe 
cc @stripe/developer-products

 ### Summary

append query from registered webhook endpoint to the forward-to url
`stripe listen --load-from-webhooks-api --forward-to localhost:5000`


gracegoo-stripe (approved):
LGTM

PR #941, opened 2022-08-12
Add pending_invoice_items_behavior=include to invoice fixtures

vcheung-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products @karlr-stripe

 ### Summary

In version [2022-08-01](https://stripe.com/docs/upgrades#2022-08-01), the invoice create method no longer includes the invoiceitems by default. This causes `invoice.payment_failed` to actually succeed in 2022-08-01 because the invoice is for zero dollars. To fix this, we set a parameter on that method to include pending invoiceitems in the invoice.

I tested all the fixtures on 2020-08-27 and 2022-08-01 to verify the correct events are emitted in both versions.


tomer-stripe (approved):

tomer-stripe:
Closing this one in favor of #944

PR #943, opened 2022-08-13
Fix Checkout related fixtures to not use legacy line items path

remi-stripe:
### Reviewers
r? @tomer-stripe 
cc @stripe/developer-products

 ### Summary
Fix Checkout related fixtures to not use legacy line items path as it doesn't work on the latest API version.

### DIFF
--- a/pkg/fixtures/triggers/checkout.session.async_payment_failed.json
+++ b/pkg/fixtures/triggers/checkout.session.async_payment_failed.json
@@ -4,6 +4,25 @@
   },
   "fixtures": [
     {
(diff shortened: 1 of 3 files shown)


etsai-stripe:
if user specify an older API version when triggering the fixture (https://github.com/stripe/stripe-cli/blob/master/pkg/cmd/trigger.go#L58), will it work?

remi-stripe:
@etsai-stripe yes `line_items` have existed and been the default for years, it works on all API versions.

etsai-stripe (approved):

PR #955, opened 2022-08-25
Payout.* fixtures updates

bensnell-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

Updating payout.created and payout.updated fixtures to include a PaymentIntent creation that bypasses pending balance to ensure there is available balance for these triggers to succeed.

Tested locally for both triggers and works great!


bensnell-stripe:
r?

gracegoo-stripe (approved):
LGTM!

PR #975, opened 2022-09-20
Poll for API keys earlier in the login flow

vcheung-stripe:
### Reviewers
cc @stripe/developer-products

 ### Summary

Before, when running `stripe login`, the CLI would wait for user input to open the browser, then try to retrieve keys. This caused an issue where if you went to the auth page directly without input, the CLI would not retrieve keys. We fix this by polling for keys concurrently while waiting for input.

### DIFF
--- a/pkg/cmd/login.go
+++ b/pkg/cmd/login.go
@@ -1,8 +1,6 @@
 package cmd
 
 import (
-	"os"
-
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


vcheung-stripe:
I initially left tests out because I had problems with mocking stdin, but I found way to do it using an interface

gracegoo-stripe (approved):
LGTM!

PR #976, opened 2022-09-22
Build and release for linux arm64

vcheung-stripe:
### Reviewers
cc @stripe/developer-products

 ### Summary

Update goreleaser config to build and release for linux arm64. Will need to verify if it works with package managers.

### DIFF
--- a/.goreleaser/linux.yml
+++ b/.goreleaser/linux.yml
@@ -17,6 +17,17 @@ builds:
       - linux
     goarch:
       - amd64
+  - id: stripe-linux-arm
+    ldflags:
+      - -s -w -X github.com/stripe/stripe-cli/pkg/version.Version={{.Version}}
+    binary: stripe
+    env:
+      - CGO_ENABLED=0
+    main: ./cmd/stripe/main.go
+    goos:
+      - linux
+    goarch:
+      - arm64
 archives:
   - replacements:
       linux: linux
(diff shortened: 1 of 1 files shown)


.goreleaser/linux.yml
```diff
@@ -17,6 +17,17 @@ builds:
       - linux
     goarch:
       - amd64
+  - id: stripe-linux-arm
+    ldflags:
+      - -s -w -X github.com/stripe/stripe-cli/pkg/version.Version={{.Version}}
+    binary: stripe
+    env:
+      - CGO_ENABLED=0
```
etsai-stripe, line 25: is there a reason why `CGO_ENABLED` is off?

vcheung-stripe: `goreleaser build -f .goreleaser/linux.yml` fails on my local machine unless `CGO_ENABLED` is off. I'm not really sure what the implications are.

etsai-stripe:
wondering if we should release it separately from your other PR in case this doesnt work for jfrog or dockers. but i could be just over-thinking

vcheung-stripe:
> wondering if we should release it separately from your other PR in case this doesnt work for jfrog or dockers. but i could be just over-thinking

That's fair, I'll release the other one first.

etsai-stripe (approved):

PR #992, opened 2022-10-21
feat: Add --edit flag to `stripe trigger` command

charliecruzan-stripe:
### Reviewers
r? @stripe/developer-products
cc @stripe/developer-products

 ### Summary

Adds the `--edit` flag to `stripe trigger`. This allows the user to directly inspect and edit the fixture payload for the trigger directly in their preferred IDE. This avoids the need for people to directly inspect the source files of our fixtures (so that they know what to add/override), and also avoids the mixups people run into with when to use `--add` vs `--override`.

The IDE we open relies on [`GIT_EDITOR`](https://git-scm.com/docs/git-var)

Screen recording shows:
1. triggering with vscode set as default editor & a successful fixture payload
2. triggering with vim set as the default editor & setting an improper payload param

https://user-images.githubusercontent.com/97612659/197289491-419cc6f5-261b-4f80-9e93-0dc0b2fbb6db.mov


pkg/rpcservice/trigger.go
```diff
@@ -32,6 +32,7 @@ func (srv *RPCService) Trigger(ctx context.Context, req *rpc.TriggerRequest) (*r
 		req.Remove,
 		req.Raw,
 		req.ApiVersion,
+		req.Edit,
```
etsai-stripe, line 35: curious why for the fixtures rpcservice `edit` is always false, but for triggers is user defined?

pkg/fixtures/fixtures.go
```diff
@@ -254,6 +262,23 @@ func (fxt *Fixture) Remove(removals []string) error {
...
+		return nil, fixtureRewriteError{operation: "edit", err: err, fixture: fxt}
+	}
+
+	return editor.EditContent()
+}
+
+func getFixtureFilenameWithWildcard(path string) string {
+	pathComponents := strings.Split(path, "/")
+	fixtureName := strings.Split(pathComponents[len(pathComponents)-1], ".")
+	return strings.Join(fixtureName[0:len(fixtureName)-1], ".") + ".*." + fixtureName[len(fixtureName)-1]
```
etsai-stripe, line 279: why do we need to add wildcard to the fixtures path name?

pkg/git/editor.go
```diff
@@ -0,0 +1,144 @@
...
+	editCmd := exec.Command(programArgs[0], programArgs[1:]...)
+	editCmd.Stdout = os.Stdout
+	editCmd.Stdin = os.Stdin
+	editCmd.Args = append(editCmd.Args, e.File)
+	return editCmd, nil
+}
+
+// Get's the OS default git editor via the `git var GIT_EDITOR` command.
+func getDefaultEditor() (string, error) {
+	output, err := exec.Command("git", "var", "GIT_EDITOR").Output()
```
etsai-stripe, line 92: will this work on a new docker container? do we need to check if `git` is installed first?

pkg/fixtures/fixtures.go
```diff
@@ -98,6 +99,13 @@ func NewFixtureFromFile(fs afero.Fs, apiKey, stripeAccount, baseURL, file string
 		}
 	}
 
+	if edit {
+		filedata, err = fxt.Edit(file, filedata)
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
etsai-stripe, line 108: do we still want to allow other edit flags (ie `override` `add` `remove` `skip`) if we are using `edit` flag?

pkg/git/editor.go
vcheung-stripe, line 92: Suggest using the same editor as `stripe config --edit`: https://github.com/stripe/stripe-cli/blob/2b00f2fb7f28ae652f551d2713df6cfdd17c556d/pkg/config/config.go#L173-L202

pkg/fixtures/fixtures.go
charliecruzan-stripe: good question! I couldn't think of a really strong reason to **block** users from doing this (aside from the fact that it makes it slightly easier to mess yourself up, like by editing the fixture first and then trying to override a value that isn't there)

But I think that the scenario where devs use `--edit` to change just one value one-off in their fixture, while also having a long list of `--add`s (more convenient if you're adding something every single time) outweighed that possibility so I decided to keep it.

Maybe having the `edit` step go last makes more sense? Since you're visually seeing the fixture in that step, and might be confused why your `add`ed property isn't there

pkg/rpcservice/trigger.go
charliecruzan-stripe: I was really just following what looked like convention here- for `triggers` we're passing the user definitions for all the params, whereas for `fixtures` we're passing the defaults

pkg/git/editor.go
charliecruzan-stripe: is `EDITOR` a preset env variable? It's not configured on my machine 🤔 

I like relying on an env variable that's _probably_ set by the user already, but this would be good to fallback to in case `git` isn't installed

charliecruzan-stripe: good point, this will error out if git is not installed. Could fallback to [this](https://github.com/stripe/stripe-cli/pull/992#discussion_r1007101780) if it fails?

pkg/fixtures/fixtures.go
etsai-stripe: i think its not that intuitive whether the `add/remove/skip/override` is applied before or after the `edit`. it might make sense to check that if user already has `edit` specified, then we dont want other fixture customization flags.

pkg/fixtures/fixtures.go
etsai-stripe: why cant we have the random string appended to the end? why is in the middle a better?

charliecruzan-stripe: appending the string at the end (after the `.json` file extension) gets rid of syntax highlighting for JSON file in your editor

pkg/fixtures/fixtures.go
charliecruzan-stripe: fixed in https://github.com/stripe/stripe-cli/pull/992/commits/5423ce16a51fc0aa0c94de337c6f0cbe7d8798f2

pkg/git/editor.go
charliecruzan-stripe: Using this logic now for a fallback, see https://github.com/stripe/stripe-cli/pull/992/commits/4932758ac2671f879224c9a3a4a35720c9aaac39

charliecruzan-stripe: fixed in https://github.com/stripe/stripe-cli/pull/992/commits/4932758ac2671f879224c9a3a4a35720c9aaac39

etsai-stripe (approved):
very cool feature! thank you for putting it in!

endless-win:
1

Issue #993, opened 2022-10-24
exec /bin/stripe: exec format error when running v1.13.0 in a docker container


occamssafetyrazor:
### Issue
I'm trying to run the latest version v1.13.0 in a docker container:
`docker run --rm -it --name stripe_old stripe/stripe-cli:v1.13.0`
This returns "exec /bin/stripe: exec format error"

### Expected Behavior
The latest version to return function as it did when running a container with version v1.12.4:
`docker run --rm -it --name stripe_old stripe/stripe-cli:v1.12.4`
This does not return an error

### Steps to reproduce
Just run
`docker run --rm -it --name stripe_old stripe/stripe-cli:v1.13.0`

### Traceback
exec /bin/stripe: exec format error

### Environment
Ubuntu 22.04.1 LTS
Docker version 20.10.17

tomer-stripe:
Hi folks, we're trying to reproduce this but are having some trouble. Could you share:
* Stripe CLI version
* Ubuntu version
* Docker version
* CPU arch (Intel, ARM/M1, etc)

occamssafetyrazor:
Hi @tomer-stripe 

stripe-cli:v1.13.0
Ubuntu 22.04.1 LTS
Docker version 20.10.17
Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz

Moizsohail:
Getting the same error

dmwyatt:
Getting this as well:

* latest stripe CLI version as of the time of this post (aka, not specifying version on `docker run` command)
* Ubuntu 20.04.4
* Docker version 20.10.17, build 100c701
* CPU info follows (this is on Digital Ocean server)

```
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 63
model name      : DO-Regular
stepping        : 2
microcode       : 0x1
cpu MHz         : 2199.996
cache size      : 4096 KB
physical id     : 0
siblings        : 1
core id         : 0
cpu cores       : 1
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 13
wp              : yes
[...]

PR #999, opened 2022-11-11
Updating image template to include arch

gracegoo-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary
Users reported not being able to run the stripe binary that comes in the docker image: https://github.com/stripe/stripe-cli/issues/993.
For some reason, I wasn't able to reproduce this with either version on my WSL2 setup, but I could verify that the 12.4.1 version would always pull the x86_64 and the 1.13.0 would always pull ARM. I think we just end up with whatever version we last pushed to the image since they share the same template. 
This adds the arch to the image template so we can have both instances. I would hope that grabbing just `stripe/stripe-cli` would grab the latest and correct arch automatically, but I'll have to test this after we run the release.

### DIFF
--- a/.goreleaser/linux.yml
+++ b/.goreleaser/linux.yml
@@ -71,8 +71,8 @@ dockers:
       - stripe
       - stripe-linux
     image_templates:
-      - "stripe/stripe-cli:latest"
-      - "stripe/stripe-cli:{{ .Tag }}"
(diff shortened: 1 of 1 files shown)


vcheung-stripe (approved):
TY for fixing this!

PR #1000, opened 2022-11-11
fix: do not overwrite local plugin manifest if fetched manifest is invalid

charliecruzan-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary

Previously, we'd fetch the remote plugin manifest and write it the local config. this caused issues if the remote manifest was invalid (for example, if it was empty), since the install step would fail, but it would break the `stripe plugin upgrade` and `stripe plugin install` commands from then on. 
This PR changes it so that:
- when running `install`, it will fail fast with a useful error
- when running `upgrade`, it will log that we couldn't fetch the remote manifest, so we'll upgrade based off of the local `plugins.toml`

Screen Shot 2022-11-11 at 3 49 04 PM


vcheung-stripe (approved):
LGTM

PR #1007, opened 2022-12-02
Add --api-base flag suport to fixtures cmd

bernerd-stripe:
### Reviewers
r?
cc @stripe/developer-products

 ### Summary
Most CLI commands support the (hidden) `--api-base` flag to override the base URL used for constructing API requests, which is useful during development and debugging.

The fixtures command did not include this, though, so it was possible to send individual requests to an alternate API endpoint but not a batch of fixture requests.


etsai-stripe (approved):

gracegoo-stripe (approved):
LGTM!

PR #1009, opened 2022-12-03
Add idempotency_key support to fixture request spec

bernerd-stripe:
### Reviewers
r?
cc @stripe/developer-products

 ### Summary

* Adds support for setting an idempotency key on fixture requests so they can be retried multiple times without duplicating data
* Fixes (and adds test coverage for) automatic metadata attributing resources to fixtures, which implementing idempotency keys turned up

### DIFF
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -43,6 +43,7 @@ type FixtureRequest struct {
 	Path              string                 `json:"path"`
 	Method            string                 `json:"method"`
 	Params            map[string]interface{} `json:"params"`
+	IdempotencyKey    string                 `json:"idempotency_key"`
 }
 
 // FixtureQuery describes the query in fixture request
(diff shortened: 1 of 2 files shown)


pkg/fixtures/fixtures.go
```diff
@@ -297,19 +298,10 @@ func (fxt *Fixture) UpdateEnv() error {
...
 
 func (fxt *Fixture) makeRequest(ctx context.Context, data FixtureRequest, apiVersion string) ([]byte, error) {
-	var rp requests.RequestParameters
-
-	if data.Method == "post" && !fxt.FixtureData.Meta.ExcludeMetadata {
-		now := time.Now().String()
-		metadata := fmt.Sprintf("metadata[_created_by_fixture]=%s", now)
-		rp.AppendData([]string{metadata})
-	}
-
```
bernerd-stripe, line 307: My first pass at the idempotency keys was to add `rp.SetIdempotency(idempotencyKey)` here like metadata, but found that tests failed. In fact, these request parameters are entirely unused during the fixture execution path!

gracegoo-stripe: Thank you so much for fixing this!

gracegoo-stripe (approved):
LGTM!

PR #1015, opened 2023-01-04
For plugins, only enable color output if command is running in a terminal

vcheung-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

This PR changes the plugin interface so that the host tells the plugin process if the host's stdin/out/err are terminals or not. Plugins can read this this to decide if they should print color output. In addition, this opens up the possibility of adding other terminal-only features, evolving the plugin interface, and separating concerns between host and plugin.

This change is needed because functions like [`term.IsTerminal`](https://pkg.go.dev/golang.org/x/term#IsTerminal) always return false when called inside the plugin due to how the plugin's file descriptors are managed by the host process.

To change the interface, we leverage the [go-plugin](https://github.com/hashicorp/go-plugin) package's versioning ability to implement a "major change" to the plugin interface. The old interface is defined as:
```
type Dispatcher interface {
	RunCommand(args []string) (string, error)
}
```
while the new interface is defined as:
```
type DispatcherGRPC interface {
	RunCommand(additionalInfo *proto.AdditionalInfo, args []string) error
}
```

This new interface uses gRPC instead of net/rpc to handle IPC between the host and plugin. gRPC is needed because the net/rpc implementation can't handle serializing complex data structures, which we've introduced in a struct called `AdditionalInfo`.

This change is backwards compatible because the go-plugin library automatically handles version negotiation. Although the new interface is backwards incompatible, we keep the old interface around so older clients can continue to request the old interface.


pkg/ansi/ansi.go
```diff
@@ -36,6 +36,12 @@ var DisableColors = false
 // `CLICOLOR_FORCE`. Cf. https://bixense.com/clicolors/
 var EnvironmentOverrideColors = true
 
+// HostStdoutIsTerminal enables color on stdout output when this package is used in a plugin.
+var HostStdoutIsTerminal = true
+
+// HostStderrIsTerminal enables color on stderr output when this package is used in a plugin.
+var HostStderrIsTerminal = true
+
```
vcheung-stripe, line 44: These are global variables that the plugin process sets at runtime, after it receives the rpc request from the host and reads the `AdditionalInfo` struct.

tjfontaine-stripe: I don't see where the plugin sets that information, is that handled on a per plugin basis, iow is a plugin expected to call this interface if they care about the information?

vcheung-stripe: These variables will be set by the plugin bootstrap library that all plugins import: https://github.com/stripe-dev/stripe-cli-plugin-bootstrap/pull/3. Individual plugin code will not handle any of this.

tjfontaine-stripe: Ah, excellent, that's what I was looking for -- I had missed that repo

gracegoo-stripe (approved):

PR #1016, opened 2023-01-06
Refactor `samples.go` to be more configurable

richardm-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary
Refactors the `samples` command so that the code can be re-used/configured via plugin.

Specifically
* I renamed the `Samples` struct to `SampleManager` as identifier `samples` was referring both to the package and to an instance of `Samples` which I found confusing.
* I gave `SampleManager` a constructor helper, `NewSampleManager` with a default initialization, rather than having it initialized directly everywhere.
* I changed the `Create` and `GetSampleConfig` functions to be methods with a `SampleManager` receiver, so that they are more configurable.
* I refactored `ConfigureDotEnv` into `SampleManager.ConfigureDotEnv` (which can be given a custom implementation, now that it's on the struct), and `WriteDotEnv`.
* I introduced a `SampleLister` interface to replace `getSamples`, so that this can be customized (e.g., an implementation that returns a hardcoded list instead of fetching and caching from a specific github repository) and refactored the existing implementation into `cachedGithubSampleLister`.

### Testing
As a basic sanity check, I manually ran `stripe samples list` and `stripe samples create` and they appear still to work.

### DIFF
--- a/pkg/cmd/samples/create.go
+++ b/pkg/cmd/samples/create.go
@@ -64,7 +64,13 @@ func (cc *CreateCmd) runCreateCmd(cmd *cobra.Command, args []string) error {
 	color := ansi.Color(os.Stdout)
 	spinner := ansi.StartNewSpinner(fmt.Sprintf("Downloading %s", selectedSample), os.Stdout)
 
-	sampleConfig, err := samples.GetSampleConfig(selectedSample, cc.forceRefresh)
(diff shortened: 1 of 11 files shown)


gracegoo-stripe (approved):
Thanks for refactoring this, LGTM!

PR #1025, opened 2023-02-03
Update test that fails locally

vcheung-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary

This test passes in CI, but fails locally when you run `make test` due to side effects from other tests. Fix this by running the test in its own subprocess as per https://stackoverflow.com/questions/26225513/how-to-test-os-exit-scenarios-in-go


endless-win:
rr

ianjabour-stripe (approved):
LGTM

PR #1067, opened 2023-05-01
Support additional plugin manifests

vcheung-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

Add support for merging multiple plugin manifests into one plugins.toml, so we can conditionally enable plugin visibility for specific users. For example, to enable the `foo` plugin, we send to the client:
```
GET /v1/stripecli/get-plugin-url
{
  "base_url": "https://stripe.jfrog.io/artifactory/stripe-cli-plugins-local",
  "additional_manifests": [
    "plugins-foo.toml"
  ]
}
```
where plugins-foo.toml contains the manifest for `foo`, and the client handles merging the data into plugins.toml locally.


pkg/plugins/utilities.go
```diff
@@ -135,16 +135,68 @@ func RefreshPluginManifest(ctx context.Context, config config.IConfig, fs afero.
...
+
+// Merge additional plugin lists into the main plugin list, in place
+func mergePluginLists(pluginList *PluginList, additionalPluginLists []*PluginList) {
+	for _, list := range additionalPluginLists {
+		for _, pl := range list.Plugins {
+			idx := findPluginIndex(pluginList, pl)
+			if idx == -1 {
+				pluginList.Plugins = append(pluginList.Plugins, pl)
+			} else {
+				pluginList.Plugins[idx].Releases = append(pl.Releases, pluginList.Plugins[idx].Releases...)
```
vcheung-stripe, line 175: Just realized this line violates this assumption: https://github.com/stripe/stripe-cli/blob/fdf82762ad5089ebc4ce0839b9401093a1c58495/pkg/plugins/plugin.go#L148

I will update this to sort the releases after merging.

pkg/plugins/utilities.go
```diff
@@ -135,16 +136,77 @@ func RefreshPluginManifest(ctx context.Context, config config.IConfig, fs afero.
...
+		return nil, err
+	}
+
+	additionalPluginLists := []*PluginList{}
+	for _, filename := range pluginData.AdditionalManifests {
+		additionalPluginList, err := fetchPluginList(pluginData.PluginBaseURL, filename)
+		if err != nil {
+			return nil, err
+		}
+		additionalPluginLists = append(additionalPluginLists, additionalPluginList)
```
etsai-stripe, line 151: do we keep appending every time install command is ran? or do we only merge if not exist

vcheung-stripe: I want to ensure the user's plugins.toml file stays up to date, so I made this fetch and merge each file every time the install command is run.

charliecruzan-stripe (approved):

PR #1072, opened 2023-05-09
Request signing beta

charliecruzan-stripe:
### Reviewers

 ### Summary

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -60,12 +60,19 @@ var KeyRing keyring.Keyring
 
 // CreateProfile creates a profile when logging in
 func (p *Profile) CreateProfile() error {
-	writeErr := p.writeProfile(viper.GetViper())
+	// Remove existing profile first
+	v := p.deleteProfile(viper.GetViper())
+
+	writeErr := p.writeProfile(v)
 	if writeErr != nil {
 		return writeErr
 	}
 
 	return nil
+}
+
+func (p *Profile) deleteProfile(v *viper.Viper) *viper.Viper {
+	return p.safeRemove(v, p.ProfileName)
 }
 
 // GetColor gets the color setting for the user based on the flag or the
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -421,6 +421,34 @@ func getKeyExpiresAt() string {
 	return time.Now().AddDate(0, 0, KeyValidInDays).UTC().Format(DateStringFormat)
 }
 
+// ExperimentalFields are currently only used for request signing
+type ExperimentalFields struct {
+	ContextualName string
+	PrivateKey     string
+	StripeHeaders  string
+}
+
+// GetExperimentalFields returns a struct of the profile's experimental fields. These fields are only ever additive in functionality.
+func (p *Profile) GetExperimentalFields() ExperimentalFields {
+	if err := viper.ReadInConfig(); err == nil {
+		experimentalPrefix := "experimental."
+		name := viper.GetString(p.GetConfigField(experimentalPrefix + "contextual_name"))
+		privKey := viper.GetString(p.GetConfigField(experimentalPrefix + "private_key"))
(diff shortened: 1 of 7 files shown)


pkg/requests/base.go
```diff
@@ -488,3 +499,35 @@ func normalizePath(path string) string {
...
+		if len(header) != 2 {
+			continue
+		}
+		headerName := header[0]
+		headerValue := header[1]
+		if headerName == stripeContextHeaderName {
+			displayMessage := fmt.Sprintf("Operating in %s %s\n", ansi.Bold(experimentalFields.ContextualName), ansi.Color(os.Stdout).Gray(10, "("+headerValue+")..."))
+			fmt.Print(ansi.Color(os.Stdout).Gray(10, displayMessage))
+		} else if headerName == authorizationHeaderName && privKey == "" {
+			creds, err := rb.Profile.GetSessionCredentials()
```
vcheung-stripe, line 518: Is it necessary to write the private key to the config file when we can we read them from the credential manager? I have a couple concerns:
1. Duplicating the field between the config file and credential manager could lead to them going out of sync.
2. I'm not really comfortable showing the private key in cleartext, since this could limit what we can do security-wise. For example, if we wanted to extend the session expiration from ~hours to ~days, this could make it difficult to do in a secure way.

If we can get away with not having writing this in the config file, I think we should do that.

pkg/config/profile.go
```diff
@@ -333,6 +335,9 @@ func (p *Profile) writeProfile(runtimeViper *viper.Viper) error {
 		runtimeViper = p.safeRemove(runtimeViper, "publishable_key")
 	}
 
+	// Remove experimental fields during login
+	runtimeViper = p.removeExperimentalFields(runtimeViper)
+
```
vcheung-stripe, line 340: A bit orthogonal but there's been an unresolved bug in login where the config fields don't get cleared when you `stripe login`, so e.g. if you were to login to account A, then login to account B, some of account A's configs could persist if they're not set on account B (display_name being a common one for me). I wonder if we could just delete the entire profile on login.

pkg/requests/base.go
charliecruzan-stripe: it's not necessary specifically for this case. I wasn't sure if there were edge cases where the private key is only available in the config file, which is why this code is there. But if we can reliably get it for all OSes from `GetSessionCredentials`, then im fine removing private key from the experimental section

pkg/config/profile.go
charliecruzan-stripe: I think I ran into this exact bug you're talking about- removing individual config fields doesn't seem to always work. But removing a higher level umbrella field (in this case, `experimental`) **does** in all my testing (see other PR). That being said, if there's no reason to keep around data from the previous profile, i think removing the the entire thing is our safest bet

pkg/requests/base.go
vcheung-stripe: I'm fairly confident `GetSessionCredentials` will be okay for mac and windows, but I haven't been able to validate this on linux. Given that this change is experimental still, I think this is okay but adding a ticket for validating on all OSes could be useful.

pkg/config/profile.go
charliecruzan-stripe: Tested that [removing all experimental fields](https://github.com/stripe/stripe-cli/pull/1072/commits/6e8e5248d43b7ce1a7d0124d58e16678dc5101fc) works

@vcheung-stripe I also added a commit deleting the existing profile before creating a new one with the same name, let me know if you still want me to keep that in or revert it: https://github.com/stripe/stripe-cli/pull/1072/commits/ceeb41091807d243e7597b322b8fcc49914d6a1b

vcheung-stripe: sounds good, I'm okay with keeping that commit in since presumably it'd be buggy otherwise

vcheung-stripe (approved):

vcheung-stripe (approved):

PR #1075, opened 2023-05-17
Set CGO_ENABLED=1 for arm64 build

charliecruzan-stripe:
### Reviewers

cc @stripe/developer-products

 ### Summary

In https://github.com/stripe/stripe-cli/pull/608, we added a new build for arm64, but had to exclude `CGO_ENABLED=1` because:

> because we can't run GitHub Actions on macos11, I had to remove the CGO build environment variable, otherwise the build step fails saying it's an unsupported environment

And since [keyring requires cgo for keychain](https://github.com/99designs/keyring/blob/929f178361c0be06c50f0a82598cbfa1a965f40f/keychain.go#L1-L2C21), without that flag we had to have 2 different codepaths depending on the architecture (specifically, depending on CGO_ENABLED=0 or =1). Now, in both cases, we have CGO_ENABLED=1, so we don't need those divergent code paths anymore.


vcheung-stripe (approved):
LGTM thanks for fixing this!

vcheung-stripe (approved):
LGTM thanks for fixing this!

PR #1078, opened 2023-05-23
fix: remove all profile keys when logging into a new account

charliecruzan-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary
https://jira.corp.stripe.com/browse/DX-8171

We want to clear the profile before writing to it, otherwise stale profile data can linger around. For example, if you login to a new account that doesn't have a business name set, your display name in the Stripe CLI will be innaccurate.

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -60,7 +60,7 @@ var KeyRing keyring.Keyring
 
 // CreateProfile creates a profile when logging in
 func (p *Profile) CreateProfile() error {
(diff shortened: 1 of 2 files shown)


vcheung-stripe (approved):
LGTM

charliecruzan-stripe:
@vcheung-stripe could i get another stamp? i wanted to add one more test case to make sure we're not deleting other profiles somehow

PR #1087, opened 2023-06-16
Add name and address to payment method billing details for checkout.s…

abirck-stripe:
…ession.completed webhook

 ### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary

Adds a billing name and address to the card payment method created as part of generating a checkout.session.completed webhook. This more realistically models actual payments because Checkout enforces card payments contain these attributes.


vcheung-stripe (approved):
LGTM, thanks for the PR

PR #1099, opened 2023-07-13
Better `payment_intent.requires_action` trigger 

karlr-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

Changes the `payment_intent.requires_action` trigger to be more straightforward/idiomatic and charge a card that [requires 3D Secure authentication](https://stripe.com/docs/testing?testing-method=payment-methods#authentication-and-setup) in order to have the effect of putting the PaymentIntent in `requires_action`. The previous way was added for a specific payment method and only works on Stripe accounts in Japan and just didn't make a lot of sense as a generic public fixture.
Motivation: developer user feedback


vcheung-stripe (approved):
LGTM

PR #1109, opened 2023-07-25
fix: restrict WriteConfigField to atomic writes

charliecruzan-stripe:
### Reviewers
cc @stripe/developer-products

 ### Summary

It's possible to get into a state where the values stored in the config file do not equal the values that viper has in memory. When this happens, if we call `.WriteConfigField`, viper will write _other_ values besides the one we are explicitly passing in, causing unforeseen behavior and very difficult to track down bugs. This change makes it so that viper reads from the current config file, then sets the key we are explicitly trying to override, then writes to the file.

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -276,6 +276,7 @@ func (p *Profile) RegisterAlias(alias, key string) {
 // WriteConfigField updates a configuration field and writes the updated
 // configuration to disk.
 func (p *Profile) WriteConfigField(field, value string) error {
+	viper.ReadInConfig()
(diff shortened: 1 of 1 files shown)


vcheung-stripe (commented):
The change makes sense, one additional thing we could verify is that this works when there's more than one profile in the config

charliecruzan-stripe:
good call! Tested and works as expected when using the `--project-name` flag: switching accounts in two different profiles only affects the appropriate profile

vcheung-stripe (approved):

PR #1114, opened 2023-08-16
Corrected the charge.dispute file name

RyanRussell00:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary
The stripe CLI supports a dispute event called `charge.dispute.created` but the file name was incorrect which led me to run the wrong command and troubleshoot a bit about why the command was not recognized.

This PR fixes the file name to prevent confusion.


etsai-stripe:
we will fix it altogether at all the places where this fixture is being used. thank you for your recommendation!

PR #1120, opened 2023-09-05
stripe samples: show 7 server options, not 5

richardm-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

When you do "stripe samples create" it uses `promptui.Select` which [defaults](https://pkg.go.dev/github.com/manifoldco/promptui#Select) to only showing 5 options. We have 7 official SDKs, so many of the stripe samples have 7 options. Why not just show all of them? It's sometimes hard to spot that little arrow that indicates that you're able to scroll, so at first I didn't think the stripe sample had support for go or .NET, but it does:
image

Tested locally:
image

Some samples have more than 7 it's true (this one has a "node-typescript" option) but let's put at least the basic versions of the 7 official libraries above the fold.

### DIFF
--- a/pkg/cmd/samples/create.go
+++ b/pkg/cmd/samples/create.go
@@ -178,6 +178,7 @@ func selectOptions(template, label string, options []string) (string, error) {
 		Label:     label,
 		Items:     options,
 		Templates: templates,
+		Size: 7,
 	}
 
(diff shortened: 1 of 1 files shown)


vcheung-stripe (approved):
Thanks!

PR #1130, opened 2023-10-17
Logout from the server when experimental config is set

vcheung-stripe:
### Summary

On `stripe logout`, make a request to the logout endpoint on the server if experimental fields in the config are set. This is a no-op if there are no experimental fields. I manually verified this end-to-end.


pkg/cmd/logout.go
```diff
@@ -25,13 +34,43 @@ func newLogoutCmd() *logoutCmd {
...
+func serverLogout(ctx context.Context, profile *config.Profile, baseURL string, apiKey string) {
+	base := &requests.Base{
+		Profile:        profile,
+		Method:         http.MethodPost,
+		SuppressOutput: true,
+		APIBaseURL:     baseURL,
+	}
+
+	// Fail open to allow the logout command to complete
+	resp, err := base.MakeRequest(ctx, apiKey, "/v1/stripecli/logout", &requests.RequestParameters{}, true)
```
etsai-stripe, line 66: does it throw if user logs out more than once?

Issue #1013, opened 2022-12-30
Stripe CLI not using auth from keychain


yassinecc:
### Issue
I can successfully log in with `stripe login`, and credentials are properly written into the Ubuntu keychain + `~/.config/stripe/config.toml` under the `default` project

When running `stripe listen --live` or `stripe events resend --live`, I always get redirected to the login flow. The CLI calls succeed  when I include the restricted API key

### Expected Behavior
After successfully logging in with `stripe login`, I should not have to include an API key to CLI commands for them to work

### Steps to reproduce
- Run `stripe login`
- Call e.g. `stripe listen --live`
- The listened should correctly start

### Traceback
I don't know how to get debug logs, could not see any error logs

### Environment
Ubuntu 22.04.1 LTS
stripe CLI v1.13.6

ksullivan:
Looks like this fixes it...

```diff
	// KeyManagementService is the key management service name
-	KeyManagementService = "Stripe CLI"
+	KeyManagementService = "Stripe"
```
There seems to be an issue with the gnome keyring where it can't handle spaces in service names...

See https://github.com/99designs/keyring/issues/44, https://github.com/99designs/keyring/pull/83 and https://github.com/99designs/keyring/pull/90

weirdvic:
Hello I have the same problem. Is there any workaround available?

nasgir:
Can confirm this bug on Ubuntu 22.10 and Stripe CLI 1.14.7 - I get stuck in an authentication loop when trying to run 'stripe event resend'. I used the Windows application instead and it worked.

PR #1136, opened 2023-12-04
Implement suggested fix from #1013

lol768:
Fixes auth-loop on Linux/Gnome systems.

Thanks @ksullivan

See also
https://github.com/GNOME/gnome-keyring/blob/f24e374/daemon/dbus/gkd-secret-util.c#L111 for root-cause

 ### Reviewers
r? @
cc @stripe/developer-products

 ### Summary
- Previously, users on Linux/Gnome were stuck in an auth loop and unable to use the Stripe CLI _at all_ for anything live mode related. This is because of a bug in the upstream 99designs keyring module (which I think https://github.com/99designs/keyring/pull/90 would've fixed if it was pulled). We work around Gnome's encoding logic and 99designs' handling logic by using a simpler keychain name, without spaces.
- If pulled, this is likely to require some users who already have credentials in the previously-named keychain to need to re-authenticate.

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -52,7 +52,7 @@ const (
 	KeyValidInDays = 90
 
 	// KeyManagementService is the key management service name
-	KeyManagementService = "Stripe CLI"
+	KeyManagementService = "StripeCLI"
 )
 
 // KeyRing ...


vcheung-stripe (approved):
Thanks for the PR! This looks good to me, we'll include this in the next release.

PR #1140, opened 2024-01-22
Fix fixture param index bug

vcheung-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

Fix a bug in fixture param parsing. If a param in your fixture is an array of objects, and one of these objects itself has an object, then we were parsing it incorrectly.

For example, if you have
```
"custom_fields": [
  {
    "label": {
      "custom": "First Name",
      "type": "custom"
    }
  }
]
```

We would expect to get
```
custom_fields[0][label][custom]=First Name
custom_fields[0][label][type]=custom
```

But we were instead getting
```
custom_fields[0][label][0][custom]=First Name
custom_fields[0][label][0][type]=custom
```

The problem was we were passing down the root array's `index` into each recursive call to the parser when we only want to pass it down one level.

### DIFF
--- a/pkg/parsers/parsers.go
+++ b/pkg/parsers/parsers.go
@@ -187,7 +187,7 @@ func ParseMap(params map[string]interface{}, parent string, index int, queryResp
 		case reflect.Map:
 			m := value.(map[string]interface{})
 
-			result, err := ParseMap(m, keyname, index, queryRespMap)
+			result, err := ParseMap(m, keyname, -1, queryRespMap)
 
 			if err != nil {
 				return make([]string, 0), err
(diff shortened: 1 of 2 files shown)


pkg/parsers/parsers.go
```diff
@@ -187,7 +187,7 @@ func ParseMap(params map[string]interface{}, parent string, index int, queryResp
 		case reflect.Map:
 			m := value.(map[string]interface{})
 
-			result, err := ParseMap(m, keyname, index, queryRespMap)
+			result, err := ParseMap(m, keyname, -1, queryRespMap)
```
etsai2-stripe, line 190: what about when we do need to pass the index recursively down?

vcheung-stripe: Do you have any examples? I'm not seeing where we would need to pass the index down more than one level. If I'm parsing a value, would I need to know what index I'm at in the grandparent?

etsai2-stripe: mmm i cannot come up with one either. it seems to be doing what you did in other places too. i think it is fine!

etsai2-stripe (approved):

PR #1157, opened 2024-03-26
[Bug] Fixes #1158 bad invoice items URL

### LINKED ISSUE #1158: Bad invoice items URL
### Issue
The URL for interacting with invoice items should be `/invoiceitems` but right now it's `/invoice_items` which makes `stripe get ii_invoice_item_id` calls fail

### Expected Behavior
Calling `stripe get` with an invoice item ID should show the JSON object for the invoice item. Instead, it's showing an error that the URL is not valid as it's trying to fetch `v1/invoice_items` instead of `v1/invoiceitems`: https://docs.stripe.com/api/invoiceitems

### Steps to reproduce
Run a command to get some invoice item with its ID:
```
stripe get ii_im_an_invoice_item_id
```

The current CLI returns the following error:
```
{
  "error": {
    "message": "Unrecognized request URL (GET: /v1/invoice_items/ii_im_an_invoice_item_id). Please see https://stripe.com/docs or we can help at https://support.stripe.com/.",
  }
}
```

However, getting the full path will correctly return the JSON object, which is the expected behavior from the first command:
```
stripe get v1/invoiceitems/ii_im_an_invoice_item_id
```

### Traceback
```
{
  "error": {
    "message": "Unrecognized request URL (GET: /v1/invoice_items/ii_im_an_invoice_item_id). Please see https://stripe.com/docs or we can help at https://support.stripe.com/.",
  }
}
```

### Environment
macOS

jcbages:
### Reviewers
r? @tomer-stripe 
cc @stripe/developer-products

 ### Summary

Hi Stripe!

When I was using the CLI to retrieve an invoice item by ID, I tried the following:
```
stripe get ii_im_an_invoice_item_id
```

This resulted in the following error:
```
{
  "error": {
    "message": "Unrecognized request URL (GET: /v1/invoice_items/ii_im_an_invoice_item_id). Please see https://stripe.com/docs or we can help at https://support.stripe.com/.",
  }
}
```

Looking at the documentation for invoice items it seems the URL has a typo as it should be `invoiceitems` instead of `invoice_items`: https://docs.stripe.com/api/invoiceitems. I changed the `ids.go` file that maps each ID to the corresponding URL as I effectively found it was mapping to `invoice_items`. Now, running the command shown above effectively renders the invoice item.

Thanks! 🎉

### DIFF
--- a/pkg/requests/ids.go
+++ b/pkg/requests/ids.go
@@ -16,7 +16,7 @@ var idURLMap = map[string]string{
 	"ic":        "/v1/issuing/cards/",
 	"ich":       "/v1/issuing/cardholders/",
 	"idp":       "/v1/issuing/disputes/",
-	"ii":        "/v1/invoice_items/",
+	"ii":        "/v1/invoiceitems/",
 	"in":        "/v1/invoices/",
 	"ipi":       "/v1/issuing/transactions/",
 	"issfr":     "/v1/radar/early_fraud_warnings/",


tomer-stripe (approved):
Thanks!

PR #1173, opened 2024-04-22
add action to test package manager installs

charliecruzan-stripe:
### Reviewers

r? @vcheung-stripe 
or
r? @etsai-stripe 

 ### Summary

Adds a new Github Action that acts as a smoke test for all of our package manager install cases. This will install Stripe and run `stripe version` after:
- installing with homebrew
- installing with apt
- installing with scoop
- installing with yum
- running in the docker image

This will run once an hour, and also whenever a new tag is pushed. If it fails, this is a strong indicator that something broken in the installation flow and this should help us catch it early. To avoid noise, there's retry logic.

### DIFF
--- a/.github/workflows/install-test.yml
+++ b/.github/workflows/install-test.yml
@@ -22,8 +22,8 @@ jobs:
     runs-on: ubuntu-latest
     container: ubuntu:18.04
     steps:
-      - run: sudo apt update
-      - run: sudo apt install yum
+      - run: apt update
+      - run: apt install yum
       - run: mkdir /etc/yum.repos.d
       - run: echo -e "[Stripe]\nname=stripe\nbaseurl=https://packages.stripe.dev/stripe-cli-rpm-local/\nenabled=1\ngpgcheck=0" >> /etc/yum.repos.d/stripe.repo
       - run: yum install stripe
--- a/.github/workflows/install-test.yml
+++ b/.github/workflows/install-test.yml
@@ -25,7 +25,7 @@ jobs:
       - run: apt-get update
       - run: apt-get -y install yum
       - run: mkdir /etc/yum.repos.d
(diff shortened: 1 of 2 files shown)


.github/workflows/install-test.yml
```diff
@@ -0,0 +1,45 @@
+on:
+  schedule:
+    - cron: '0 0 * * *' # Every day at midnight
```
fhats-stripe, line 3: why only at midnight? it seems like there's probably value in running this more frequently

.github/workflows/install-test.yml
```diff
@@ -0,0 +1,45 @@
...
+      - run: echo "deb [signed-by=/usr/share/keyrings/stripe.gpg] https://packages.stripe.dev/stripe-cli-debian-local stable main" | sudo tee -a /etc/apt/sources.list.d/stripe.list
+      - run: sudo apt update
+      - run: sudo apt install stripe
+      - run: stripe --version
+  yum:
+    runs-on: ubuntu-latest
+    container: ubuntu:18.04
+    steps:
+      - run: apt-get update
+      - run: apt-get -y install yum
```
fhats-stripe, line 27: sad we have to do this intead of `runs-on: centos` but i see github actions doesn't support any linux runners besides ubuntu :(

.github/workflows/install-test.yml
charliecruzan-stripe: It seems like _maybe_ artifactory might rate limit requests ([one of these runs resulted in a 403](https://github.com/stripe/stripe-cli/actions/runs/8790803561/job/24123652260)), so that would be something to be aware of. But we could easily do this once an hour if we want

.github/workflows/install-test.yml
charliecruzan-stripe: Not ideal but it should work in perpetuity without us having to fiddle with it

charliecruzan-stripe:
r? @fhats-stripe 

I changed this to run every hour. I'd like to see how this plays in the wild before adding any paging/notification to the workflow

PR #1218, opened 2024-07-15
[actions] use pagerduty CLI to create alert

charliecruzan-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

The github action isn't maintained and doesn't have a lot of features, let's just use Pager Duty CLI instead. This makes it so that the title of the pagerduty alert isn't a very long and confusing string plus url (plus my name, which im not a super big fan of 😄 ). Hopefully this will make this page a lot clearer to the runner

### DIFF
--- a/.github/workflows/install-test.yml
+++ b/.github/workflows/install-test.yml
@@ -71,8 +71,7 @@ jobs:
   pagerduty:
     needs: [docker, scoop, yum, apt, homebrew]
     runs-on: ubuntu-latest
-    if: ${{ always() }}
-    # if: ${{ always() && contains(needs.*.result, 'failure') }}
+    if: ${{ always() && contains(needs.*.result, 'failure') }}
     steps:
       - shell: bash
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/install-test.yml
```diff
@@ -73,8 +73,13 @@ jobs:
...
-      - name: Send PagerDuty alert on failure
-        uses: Entle/action-pagerduty-alert@0.2.0
-        with:
-          pagerduty-integration-key: '${{ secrets.PAGERDUTY_INTEGRATION_KEY }}'
-          pagerduty-dedup-key: package_manager_install_workflow_failed
+      - shell: bash
+        env:
+          PAGERDUTY_INTEGRATION_KEY: ${{ secrets.PAGERDUTY_INTEGRATION_KEY }}
+        run: |
+          sh -c "$(curl -sL https://raw.githubusercontent.com/martindstone/pagerduty-cli/master/install.sh)"
```
fhats-stripe, line 80: thanks i hate it

fhats-stripe (approved):
👍

PR #1232, opened 2024-08-29
Add more logging to websocket client and attempt to fix deadlock in stripe listen

tomas-stripe:
### Reviewers
r? @fhats-stripe 
cc @stripe/developer-products

 ### Summary

Adds logs to some blind spots in the websocket client. The goal is to help debug a somewhat rare deadlock in the listen command when its automatically restarting a connection after the reconnect interval has passed.

I was able to reproduce the deadlock by manually setting a short time to reconnect on the `stripe listen` command (around 3s), and in parallel triggering events in a loop using `stripe trigger`.

After a few minutes, the `stripe listen` command ran into a deadlock, at which point I attached a delve debugger to see that it was locked on the `writePump` goroutine sending a message to the `notifyClose` channel: 

https://github.com/stripe/stripe-cli/blob/5ce3aba721970df00229725e7db344e4f05c3dff/pkg/websocket/client.go#L495

Screenshot 2024-08-29 at 5 42 11 PM

The deadlock happens when the `writePump` and the `readPump` goroutines race to send a message to the `notifyClose` channel. The fix I'm attempting here is to make the `notifyClose` channel buffered with non-blocking reads and writes.

I confirmed (by observing the new log lines) that the same code path that previously used to deadlock is now deadlock free.

Out of an abundance of caution, I'm putting the `notifyClose` fix behind a feature flag that's controlled by an environment variable. This is quite similar to https://github.com/stripe/stripe-cli/pull/798 but hopefully covers all reads and writes to `notifyClose`. I'll let the fix soak in an internal canary to confirm it's working before releasing it in a future version.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -132,6 +132,12 @@ func (c *Config) InitConfig() {
 		}).Debug("Using profiles file")
 	}
 
+	if os.Getenv("STRIPE_CLI_CANARY") == "true" {
+		log.WithFields(log.Fields{
+			"prefix": "config.Config.InitConfig",
+		}).Debug("Running with STRIPE_CLI_CANARY=true")
+	}
+
 	if c.Profile.DeviceName == "" {
 		deviceName, err := os.Hostname()
 		if err != nil {
--- a/pkg/websocket/client.go
+++ b/pkg/websocket/client.go
@@ -339,7 +339,11 @@ func (c *Client) changeConnection(conn *ws.Conn) {
 	c.stopReadPumpMutex.Lock()
 	defer c.stopReadPumpMutex.Unlock()
 	c.conn = conn
-	c.notifyClose = make(chan error, 1)
+	if os.Getenv("STRIPE_CLI_CANARY") == "true" {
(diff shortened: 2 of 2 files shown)


fhats-stripe (approved):

pkg/config/config.go
```diff
@@ -132,6 +132,12 @@ func (c *Config) InitConfig() {
 		}).Debug("Using profiles file")
 	}
 
+	if os.Getenv("STRIPE_CLI_CANARY") == "true" {
```
fhats-stripe, line 135: i'm peaceful with it for now since it's meant as a temporary patch to vet the fix, but i'd feel a lot more comfortable if this was in a helper function since the result typing is loose and we're using a bool-as-a-string from an env var, which is easy to typo

fhats-stripe (approved):

pkg/config/config.go
tomas-stripe: Makes sense! I need to resolve this to be able to merge but will use a helper method the next time we need an env var check.

PR #1241, opened 2024-09-16
fix: link to thin event in local listener view

charliecruzan-stripe:
### Reviewers
r? @tomas-stripe 
cc @stripe/developer-products

 ### Summary

This adds support for linking directly to an event delivery attempt in the local listener view in workbench. We pass back the `DeviceToken` now, and use that to construct the URL. Tested this locally and confirmed the URL works as expected


tomas-stripe (approved):
Would be really nice to have a test for this, but I don't see an existing test setup that'd be easy to extend to support this 🤔

PR #1243, opened 2024-09-16
fix: show pretty error when trying to create meter event stream or session

charliecruzan-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary

These paths aren't supported yet, but we want to throw a helpful error until they are

Screenshot 2024-09-16 at 11 38 17 AM

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -204,6 +204,7 @@ func init() {
 	rootCmd.AddCommand(newCommunityCmd().cmd)
 	rootCmd.AddCommand(newPluginCmd().cmd)
 	addAllResourcesCmds(rootCmd)
+	addV2BillingStubs(rootCmd)
 
 	err := resource.PostProcessResourceCommands(rootCmd, &Config)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/cmd/root.go
```diff
@@ -225,3 +226,15 @@ func init() {
...
+func addV2BillingStubs(rootCmd *cobra.Command) {
+	cmd, _, err := rootCmd.Find([]string{"billing"})
+	if err != nil {
+		// silently fail
+		return
+	}
+	rBillingMeterEventSessionCmd := resource.NewResourceCmd(cmd, "meter_event_session")
+	rBillingMeterEventStreamCmd := resource.NewResourceCmd(cmd, "meter_event_stream")
+	resource.NewUnsupportedV2BillingOperationCmd(rBillingMeterEventSessionCmd.Cmd, "create", "/v2/billing/meter_event_session")
+	resource.NewUnsupportedV2BillingOperationCmd(rBillingMeterEventStreamCmd.Cmd, "create", "/v2/billing/meter_event_stream")
```
tomas-stripe, line 239: I know these are temporary but would be nice to have a very short test for them such as https://github.com/stripe/stripe-cli/blob/8d33a698892e945948c3f82084c3e5d51c3c7f26/pkg/cmd/root_test.go#L52

tomas-stripe (commented):
r? @charliecruzan-stripe

pkg/cmd/root.go
charliecruzan-stripe: yeah agreed, added a small test just to confirm the overrides are present

tomas-stripe (approved):

PR #1247, opened 2024-09-20
supply more URL endpoint details to server

charliecruzan-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary
Added support for these server side, so just need to supply them from the client

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -266,8 +266,10 @@ func (p *Proxy) createSession(ctx context.Context) (*stripeauth.StripeCLISession
 		// transient errors that we just need to retry for.
 		for i := 0; i <= 5; i++ {
 			devURLMap := stripeauth.DeviceURLMap{
-				ForwardURL:        p.cfg.ForwardURL,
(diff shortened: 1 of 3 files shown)


tomas-stripe (approved):

pkg/proxy/proxy.go
```diff
@@ -157,14 +157,14 @@ func (p *Proxy) Run(ctx context.Context) error {
 
 	for nAttempts < maxConnectAttempts {
 		session, err := p.createSession(ctx)
-		*p.cfg.DeviceToken = session.DeviceToken
```
charliecruzan-stripe, line 160: this could cause a segfault if we received an error back from the server. luckily we never deployed this version so no harm done

charliecruzan-stripe:
@tomas-stripe could i get another review? had to fix a pre-existing bug that this feature caused me to notice

pkg/proxy/proxy.go
tomas-stripe: Good catch

tomas-stripe (approved):

PR #1270, opened 2024-11-01
Adds the capability to alias resource names and aliases line_items -> invoice_line_items

fhats-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary
This change adds the ability to alias resources so that they are renamed in the CLI's resource list. This allows resources to effectively be renamed, while preserving backwards compatibility by supporting the "principle" (e.g. OpenAPI spec) name but hiding that name from help text output.

The first consumer of this feature is invoice line items. Today, the CLI supports these under the name `line_items`, which differs from how the same resource is accessed in the Stripe Shell and Workbench (there it's `invoice_line_items`). It's also just not a very descriptive name -- there's the notion of line items in at least one other API (credit notes), and not having the `invoice_` prefix separates this resource from its other invoicing siblings in the help text output.

As of this PR, `line_items` does not show up in the resource list and `invoice_line_items` does:

image

The CLI generates two resources in the case of an alias: one for the principle/original resource (from the openapi spec), and one for the alias name. The principle resource is then hidden using post-processing but still works for compatibility. They hit the same underlying API:

image

I added a couple tests to ensure that aliased names show up and original names don't, and that original and aliased resources hit the same underlying API endpoint.

### DIFF
--- a/pkg/cmd/resource/alias.go
+++ b/pkg/cmd/resource/alias.go
@@ -11,6 +11,7 @@ var aliasedCmds = map[string]string{
 	"line_item": "invoice_line_item",
 }
 
+// GetCmdAlias retrieves the alias for a given resource, if one is present; otherwise returns ""
 func GetCmdAlias(principle string) string {
 	alias, ok := aliasedCmds[principle]
 	if !ok {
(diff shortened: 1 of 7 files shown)


vcheung-stripe (approved):
LGTM

PR #1315, opened 2025-03-14
Add `preview v2` command support

tomer-stripe:
### Reviewers
r? @etsai2-stripe 
cc @stripe/developer-products

 ### Summary
Adds support for `preview` APIs in the CLI:
```
➜ ./stripe preview v2 --help
Usage:
  stripe preview v2 <resource> <operation> [parameters...]

Available Resources:
  ...
  billing
  core
  test_helpers
  ...
  ```

### DIFF
--- a/pkg/gen/gen_resources_cmds.go
+++ b/pkg/gen/gen_resources_cmds.go
@@ -23,8 +23,9 @@ import (
 type SpecVersion = string
 
 const (
-	V1Spec SpecVersion = "v1"
-	V2Spec SpecVersion = "v2"
+	V1Spec        SpecVersion = "v1"
+	V2Spec        SpecVersion = "v2"
+	V2PreviewSpec SpecVersion = "v2-preview"
 )
 
 type TemplateData struct {
@@ -60,6 +61,8 @@ const (
 
 	pathStripeSpecV2 = "../../api/openapi-spec/spec3.v2.sdk.json"
 
+	pathStripeSpecV2Preview = "../../api/openapi-spec/spec3.v2.sdk.preview.json"
+
 	pathTemplate = "../gen/resources_cmds.go.tpl"
 
 	pathName = "resources_cmds.go.tpl"
@@ -90,6 +93,11 @@ func main() {
 		panic(err)
 	}
 
+	v2PreviewSpec, err := spec.LoadSpec(pathStripeSpecV2Preview)
+	if err != nil {
+		panic(err)
+	}
+
 	// Generate the stripe version header
 	v2Version := v2Spec.Info.Version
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/requests/base.go [hunk 1]
```diff
@@ -318,10 +318,6 @@ func compileRequestError(body []byte, statusCode int, isV2 bool) RequestError {
 	json.Unmarshal(body, &errorBody)
 
 	msg := "Request failed"
-	if statusCode == 401 && isV2 {
```
tomer-stripe, line 321: This is an old, manual override for errors we used to provide before v2 had RAK support. Removed this since it's no longer accurate.

pkg/gen/resources_cmds.go.tpl
```diff
@@ -11,9 +11,11 @@ import (
 )
 
 func addAllResourcesCmds(rootCmd *cobra.Command) {
-	{{ range $specVersion, $vData := .Versions }}{{ if eq $specVersion "v1" }}v1root := rootCmd{{ else }}
+	{{ range $specVersion, $vData := .Versions }}{{ if eq $specVersion "v1" }}v1root := rootCmd{{ else if eq $specVersion "v2-preview" }}
+	previewRoot := resource.NewNamespaceCmd(rootCmd, "preview").Cmd
+	previewV2Root := resource.NewNamespaceCmd(previewRoot, "v2").Cmd{{ else }}
```
tomer-stripe, line 16: This is a little hacky but I wanted to ensure that the commands were properly nested under `v2 preview` and this achieves that pretty easily. We may want a more scaleable option in the future if we have v1 previews

pkg/gen/gen_resources_cmds.go
```diff
@@ -60,6 +61,8 @@ const (
 
 	pathStripeSpecV2 = "../../api/openapi-spec/spec3.v2.sdk.json"
 
+	pathStripeSpecV2Preview = "../../api/openapi-spec/spec3.v2.sdk.preview.json"
```
etsai2-stripe, line 64: where is this file?

tomer-stripe: Hasn't been committed yet since I'm waiting for approval to make this public.

tomer-stripe:
I think all of the failures around `not enough arguments in call to resource.NewOperationCmd` should get resolved next week once we commit the preview spec and regenerate all of the commands. The template is updated to pass those in during the gen+build step.

pkg/gen/gen_resources_cmds.go
```diff
@@ -306,33 +320,36 @@ func addToTemplateData(data *TemplateData, specVersion SpecVersion, nsName, resN
...
+				EnumFlags: enums,
 			}
 		}
 	}
 
 	return nil
 }
 
-func getMethodProperties(specVersion SpecVersion, specOp *spec.Operation, op spec.StripeOperation) map[string]string {
+func getMethodProperties(specVersion SpecVersion, specOp *spec.Operation, op spec.StripeOperation) (map[string]string, map[string][]spec.StripeEnumValue) {
```
charliecruzan-stripe, line 345: non blocking nit: could probably rename this method now to make it clearer

charliecruzan-stripe (approved):

pkg/gen/gen_resources_cmds.go
tomer-stripe: Ah yea. I couldn't think of a better name but open to ideas 😅

PR #1321, opened 2025-03-29
fix: duplicate config initialization

omgupta1608:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary
The config was being initialized twice which was unnecessary (see below redundant calls to `Config.InitConfig()`)
First in Cobra OnInitialize
```go
cobra.OnInitialize(Config.InitConfig, ReBindKeys)
```

then again, later in the `init()` function in `cmd/pkg/root.go` - which is redundant as cobra has already initialized it
```go
// config is not initialized by cobra at this point, so we need to temporarily initialize it
Config.InitConfig()
```

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -219,9 +219,6 @@ func init() {
 		log.Fatal(err)
 	}
 
-	// config is not initialized by cobra at this point, so we need to temporarily initialize it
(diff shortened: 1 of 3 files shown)


omgupta1608:
@tomasf-stripe ^

PR #1347, opened 2025-07-01
Bump API version to 2025-06-30.basil

tomas-stripe:
### Reviewers
r? @tomer-stripe @ianjabour-stripe 
cc @stripe/developer-products

 ### Summary

Updates the API version to 2025-06-30.basil.

I ran into an issue with the `stripe v2 billing meter_events create` command. In a nutshell, the CLI doesn't know how to handle command generation for a resource that has two operations with the same name.

The OpenAPI spec now includes two create operations for the same v2.billing.meter_event resource [0]. You can create a meter event in two different ways:
- `POST api.stripe.com/v2/billing/meter_events`
- `POST meter-events.stripe.com/v2/billing/meter_event_stream`

The issue is that the CLI command generation only generates one `stripe v2 billing meter_events create` command that favors the latter endpoint, but the CLI doesn’t support requests to different subdomains. It should still support the first endpoint but that's not possible either because it doesn't know how to favor one endpoint over the other when the two share the same command name.

I've temporarily edited the API spec to workaround this, and we'll work on a longer term solution for the next release.


tomer-stripe (approved):
Looks good! Was the majority of changes just the billing meter things we removed?

PR #1380, opened 2025-10-27
support logged in account id in the listening event URL

etsai2-stripe:
### Reviewers
r? 
cc @stripe/developer-products

 ### Summary

https://github.com/user-attachments/assets/4235857f-8c67-47e0-b1b8-f4ecf24e5d2c

### DIFF
--- a/pkg/proxy/stripeevent.go
+++ b/pkg/proxy/stripeevent.go
@@ -47,9 +47,11 @@ func BaseDashboardURL(livemode bool, connectedAccountID, loggedInAccountID strin
 		maybeTest = "/test"
 	}
 
-	account := fmt.Sprintf("/%s", loggedInAccountID)
+	account := ""
 	if connectedAccountID != "" {
 		account = fmt.Sprintf("/%s", connectedAccountID)
+	} else if loggedInAccountID != "" {
+		account = fmt.Sprintf("/%s", loggedInAccountID)
 	}
 
 	return fmt.Sprintf("https://dashboard.stripe.com%s%s", account, maybeTest)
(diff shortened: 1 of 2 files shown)


pkg/proxy/stripeevent_test.go
```diff
@@ -16,28 +16,52 @@ func TestIsConnect(t *testing.T) {
 
 func TestUrlForEventID(t *testing.T) {
 	evt := &StripeEvent{ID: "evt_123", Livemode: false, Type: "customer.created"}
-	require.Equal(t, "https://dashboard.stripe.com/test/events/evt_123", evt.URLForEventID())
+	require.Equal(t, "https://dashboard.stripe.com//test/events/evt_123", evt.URLForEventID())
```
vcheung-stripe: It looks like there's an extra `/` in these URLs

pkg/proxy/stripeevent.go
```diff
@@ -31,25 +32,25 @@ func (e *StripeEvent) IsConnect() bool {
...
+func BaseDashboardURL(livemode bool, connectedAccountID, loggedInAccountID string) string {
 	maybeTest := ""
 	if !livemode {
 		maybeTest = "/test"
 	}
 
-	maybeAccount := ""
-	if account != "" {
-		maybeAccount = fmt.Sprintf("/%s", account)
+	account := fmt.Sprintf("/%s", loggedInAccountID)
```
vcheung-stripe: To be defensive, could we check if `loggedInAccountID` is not empty? So that the URL is still valid.

pkg/proxy/stripeevent_test.go
etsai2-stripe: good catch

PR #1393, opened 2025-11-26
Support unified OpenAPI files for codegen

mello-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

- Adds unified OAS spec files containing paths/definitions of both v1 and v2 namespaces.
- Support codegen from unified file, and keep existing codegen logic for easy switchover if needed for next month's release.


pkg/cmd/resources_cmds.go
```diff
@@ -13,11 +13,12 @@ import (
 
 func addAllResourcesCmds(rootCmd *cobra.Command) {
```
mello-stripe, line 14: `resources_cmds.go` has been regenerated from the unified specs, and there were no changes to the commands.

Working as expected, since the unified OAS is supposed to represent the exact same as the separate OAS files.

pkg/gen/events_list.go.tpl
```diff
@@ -10,6 +10,12 @@ var validThinEvents = map[string]bool{ {{ range $_, $nsName := .ThinEvents }}
...
 
+{{ if and .PreviewEvents (gt (len .PreviewEvents) 0) -}}
 var validPreviewEvents = map[string]bool{ {{ range $_, $nsName := .PreviewEvents }}
 "{{ $nsName }}": true, {{end}}
 }
+
+{{ end -}}
+var validPreviewThinEvents = map[string]bool{ {{ range $_, $nsName := .PreviewThinEvents }}
+"{{ $nsName }}": true, {{end}}
+}
```
mello-stripe, line 21: If we decide to add v1 preview APIs, the logic is already present, and the `validPreviewEvents` variable is added conditionally on these previewEvents existing.

pkg/gen/gen_events_list.go
```diff
@@ -11,22 +11,18 @@ import (
...
 )
 
 type TemplateData struct {
-	Events        []string
-	ThinEvents    []string
-	PreviewEvents []string
+	Events            []string
+	ThinEvents        []string
+	PreviewEvents     []string
+	PreviewThinEvents []string
```
mello-stripe, line 22: Renamed `PreviewEvents` to `PreviewThinEvents`, as only v2 thin events are in this array.

pkg/gen/gen_events_list.go
```diff
@@ -37,7 +33,7 @@ const (
 func main() {
 	// generate `events_list.go` from OpenAPI spec file
 	// code for this func from gen_resources_cmds.go
-	templateData, err := getTemplateData()
+	templateData, err := getTemplateDataFromUnifiedSpec()
```
mello-stripe, line 36: `getTemplateData()` was left mostly untouched, so if we decide/need to move codegen to the separate specs, we can do so with a one line change

pkg/gen/gen_events_list.go
```diff
@@ -69,91 +65,118 @@ func main() {
...
+	previewEventsV2Set, err := getThinEvents(gen.PathStripeSpecV2Preview)
 	if err != nil {
 		return nil, err
 	}
 
 	data := &TemplateData{
-		Events:        eventsV1,
-		ThinEvents:    eventsV2,
-		PreviewEvents: previewEvents,
+		Events:            setToSortedSlice(eventsV1Set),
```
mello-stripe, line 85: the `getEvents()` functions were changed to return a set (well, a map) so that we could, in the new codegen logic, exclude GA events from the preview list

pkg/gen/gen_events_list.go
```diff
@@ -69,91 +65,118 @@ func main() {
...
 }
 
-func getPreviewEventList() ([]string, error) {
-	api, err := spec.LoadSpec(pathStripeSpecV2Preview)
+// getThinEvents extracts V2 thin event types from an OpenAPI spec file by iterating
+// through all schemas and finding those marked with x-stripeEvent extension where
+// eventKind is "thin".
+// Returns a set (map[string]struct{}) of unique event type strings.
+func getThinEvents(pathSpec string) (map[string]struct{}, error) {
+	api, err := spec.LoadSpec(pathSpec)
```
mello-stripe, line 153: By adding a filter for `x-stripeEvent.kind == "thin"`, the same logic can be used to get thin/v2 events from the v2-only or the unified spec

pkg/gen/gen_resources_cmds.go
```diff
@@ -20,19 +20,20 @@ import (
 	"github.com/stripe/stripe-cli/pkg/spec"
 )
 
-type SpecVersion = string
+type ApiNamespace = string
```
mello-stripe, line 23: Renamed from SpecVersion to ApiNamespace because we now only have one spec, but we still want to namespace our commands

pkg/gen/gen_resources_cmds.go
```diff
@@ -59,17 +60,9 @@ type StripeVersionTemplateData struct {
 }
 
 const (
-	pathStripeSpec = "../../api/openapi-spec/spec3.sdk.json"
-
-	pathStripeSpecV2 = "../../api/openapi-spec/spec3.v2.sdk.json"
-
-	pathStripeSpecV2Preview = "../../api/openapi-spec/spec3.v2.sdk.preview.json"
```
mello-stripe, line 66: paths moved to common `openapi_paths.go` file

scripts/validate_unified_specs.py
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env python3
```
mello-stripe: Claude and I wrote this script to do a sanity check that there's a perfect 1:1 pairing of paths in the separate vs unified specs

mello-stripe: will delete before we merge

pkg/gen/gen_events_list.go
etsai2-stripe: since we still have `previewEvents`. what will be in this array?

mello-stripe: v1 events in preview

pkg/gen/gen_events_list.go
```diff
@@ -69,91 +65,118 @@ func main() {
...
+			filteredPreviewThinEventsSet[e] = struct{}{}
 		}
+	}
 
-		eventType := schema.XStripeEvent.EventType
-		eventList = append(eventList, eventType)
+	data := &TemplateData{
+		Events:            setToSortedSlice(eventsV1Set),
+		ThinEvents:        setToSortedSlice(thinEventsSet),
+		PreviewEvents:     nil, // V1 preview events excluded for now
```
etsai2-stripe, line 123: oh i see the answer to my question here

scripts/validate_unified_specs.py
etsai2-stripe: would it be useful to make the validation as part of the build

kwu-stripe (approved):

scripts/validate_unified_specs.py
mello-stripe: I don't think so as we'll be removing the separate specs after this release

PR #1421, opened 2026-01-22
fix: suppress non-JSON output in JSON mode to enable piping to jq

### LINKED ISSUE #1353: Cannot use jq with `--format JSON`: `jq: parse error: Invalid numeric literal at line 1, column 9`
According to [this (outdated, apparently) documentation](https://docs.stripe.com/cli/listen), I should be able to use `--print-json` to print JSON objects to stdout. However, when I run `stripe listen --print-json` I get this warning message:

```
Flag --print-json has been deprecated, Please use `--format JSON` instead and use `jq` if you need to process the JSON in the terminal.
```

So I go and run `stripe listen --format JSON | jq '.'` to use `jq` to proces the JSON in the terminal as suggested by the previous warning message, but when I do I get this error:

```
jq: parse error: Invalid numeric literal at line 1, column 9
```

### Issue
I was trying to parse the json output with jq. According to the warning message, I should be able to do just that. but I can't because of the error.

### Expected Behavior
jq should not throw an error.

### Steps to reproduce
Run this command:
```
stripe listen --format JSON | jq '.'
```

### Environment
Linux

### Other useful information
When I run `stripe listen --format JSON 2>/dev/null` I see this in the terminal console for a split second, I think this might be the problem:
```
⣾ Checking for new versions...
```

nickita-khylkouski:
When using `--format JSON` or `--print-json`, status messages and version
check output were written to stdout, breaking JSON parsers like `jq`.

## Changes

- **listen.go**: Skip version check when JSON mode is active
- **listen.go**: Suppress status spinner messages in JSON mode  
- **listen.go**: Redirect error messages to stderr in JSON mode
- **tail.go**: Skip version check when JSON mode is active
- **tail.go**: Suppress status spinner messages in JSON mode
- **tail.go**: Redirect warning messages to stderr in JSON mode
- **tail.go**: Use raw JSON output instead of `ColorizeJSON` in JSON mode

## Testing

After this fix:
```bash
# This now works cleanly:
stripe listen --format JSON --skip-update 2>/dev/null | jq '.'
stripe logs tail --format JSON 2>/dev/null | jq '.'
```

Status, version, and diagnostic messages are redirected to stderr in JSON mode,
keeping stdout clean for JSON parsing.

Fixes #1353

### DIFF
--- a/pkg/cmd/listen.go
+++ b/pkg/cmd/listen.go
@@ -122,7 +122,8 @@ func (lc *listenCmd) runListenCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if !lc.printJSON && !lc.onlyPrintSecret && !lc.skipUpdate {
+	isJSONMode := strings.ToUpper(lc.format) == outputFormatJSON || lc.printJSON
+	if !lc.onlyPrintSecret && !lc.skipUpdate && !isJSONMode {
 		version.CheckLatestVersion()
 	}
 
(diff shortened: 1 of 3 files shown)


tomer-stripe:
I think this is a dupe of #1426?

PR #1424, opened 2026-01-22
fix: preserve trailing path segments in ParsePath for multiple dynamic IDs

### LINKED ISSUE #1340: Bank account verify using a fixture removes /verify from the path when using dynamic IDs

### Issue
_Describe what happened and what you were trying to do_

- Using a stripe fixture, I want to create a new customer, add a bank account to the customer, an verify the bank account.
- Ultimate goal is to be able to trigger a `charge.pending` web hook event.
- Perceived problem: when using dynamic IDs in a fixture (ex: ${bank_account:id}) for the `path`, the `/verify` is somehow lost in the value `path: /v1/customers/:id/sources/:id/verify`, resulting in a failure.

### Expected Behavior
_Tell us what you expected to happen_
- A customer is created with a verified bank account as the default payment method

### Steps to reproduce
_What are the steps we can take to reproduce this and verify it's fixed?_
- [bank.verify.fails.json](https://github.com/user-attachments/files/20511257/bank.verify.fails.json)
- [bank.verify.json](https://github.com/user-attachments/files/20511258/bank.verify.json)

- Using attached file `bank.verify.fails.json`, run `stripe fixtures bank.verify.fails.json`
  - Expected: A new customer is created with a verified bank account
  - Actual: 
```bash
Running fixture: bank.verify.fails.json
Setting up fixture for: customer
Running fixture for: customer
Setting up fixture for: bank_account
Running fixture for: bank_account
Setting up fixture for: bank_account_verify
Running fixture for: bank_account_verify
Request failed, status=400, body={
  "error": {
    "code": "parameter_unknown",
    "doc_url": "https://stripe.com/docs/error-codes/parameter-unknown",
    "message": "Received unknown parameter: amounts",
    "param": "amounts",
    "request_log_url": "https://dashboard.stripe.com/test/logs/req_vGGBD3SVjZbPkK?t=1748551143",
  }
}
```
- Retrieve and note the customer id and bank account id from the Stripe dashboard
- Using the attached file `bank.verify.json`, manually fill in the customer id and the bank account id
- Run `stripe fixtures bank.verify.json`
  - Expected: Customer bank is verified

### Traceback
_Share any debug output that was given by the CLI_
Image

### Environment
_Select one of: macOS, Linux, Windows, Other_
- macOS

nickita-khylkouski:
Fixes #1340

When parsing paths with multiple dynamic ID placeholders, the trailing segment was being dropped due to an incorrect modulo condition. Changed to check if the final pathPart is non-empty instead.

### DIFF
--- a/pkg/parsers/parsers.go
+++ b/pkg/parsers/parsers.go
@@ -78,7 +78,7 @@ func ParsePath(httpPath string, queryRespMap map[string]gjson.Result) (string, e
 			newPath = append(newPath, value)
 		}
 
-		if len(pathParts)%2 == 0 {
+		if pathParts[len(pathParts)-1] != "" {
 			newPath = append(newPath, pathParts[len(pathParts)-1])
 		}
 


tomer-stripe:
I think #1549 handles this and had a test so will use that one

PR #1429, opened 2026-01-23
fix: use configured HTTPClient for PostV2 instead of DefaultClient

nickita-khylkouski:
## Summary

Fix PostV2 using `http.DefaultClient` which has no timeout. This caused V2
event forwards to hang indefinitely on slow endpoints and hold semaphore
slots forever.

## Changes

Changed `PostV2` to use `c.cfg.HTTPClient` instead of `http.DefaultClient`,
making it consistent with the regular `Post` method which already uses the
configured client.

## Impact

Before: V2 event forwards had no timeout and could hang forever
After: V2 event forwards respect the configured timeout (default 30 seconds)

## Technical Details

`http.DefaultClient` has `Timeout: 0` meaning it waits indefinitely for a
response. When forwarding V2 events to slow or unresponsive endpoints, this
caused:

1. Goroutines to hang forever
2. Semaphore slots to be held indefinitely
3. Eventually no new forwards could happen

The configured `c.cfg.HTTPClient` has:
- Configurable timeout (default 30 seconds)
- Custom `CheckRedirect` behavior
- Custom TLS config

## Files Changed

- `pkg/proxy/endpoint.go` - One line change in PostV2 method

### DIFF
--- a/pkg/cmd/listen.go
+++ b/pkg/cmd/listen.go
@@ -122,7 +122,8 @@ func (lc *listenCmd) runListenCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if !lc.printJSON && !lc.onlyPrintSecret && !lc.skipUpdate {
+	isJSONMode := strings.ToUpper(lc.format) == outputFormatJSON || lc.printJSON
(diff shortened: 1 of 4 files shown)


tomer-stripe:
Going to close for now since this pr looks like it's including unrelated changes

PR #1441, opened 2026-02-11
Add canary testing system for end-to-end testing

tomer-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary
Note: I worked on this with Claude Code. This is something I've long wanted + was curious if it'd be useful to us. I basically want a version of compiling every PR against each target OS we support and running through some of the core capabilities to make sure it works. Curious if folks think this is useful. Claude's summary below 👇 

Adds a comprehensive end-to-end testing system that builds CLI binaries and physically invokes commands to verify real world behavior across Linux, macOS, and Windows.

Key Gap Addressed: Current tests use Cobra's ExecuteC() for in-process testing. No tests actually invoke the compiled binary, meaning packaging/build issues would go undetected.

**What's Included**

- Test Framework (canary/testutil/runner.go): Cross-platform binary execution helper with timeout, environment isolation, and config directory management
- 32 Tests (canary/runner_test.go):
  - 19 offline tests (no API key required)
  - 13 API tests (require STRIPE_TEST_API_KEY secret)
- CI Workflow (.github/workflows/canary-test.yml): Runs on all 3 platforms for every PR
- Makefile targets: make canary and make canary-offline

**Production-Like Build**

Canary builds match production settings:
- CGO_ENABLED=0
- -ldflags="-s -w"
- go generate ./...

**Test Plan**

- All 32 tests pass locally on macOS
- Verified with real Stripe test API key
- CI runs on ubuntu-latest, macos-latest, windows-latest
- Configure STRIPE_TEST_API_KEY secret for API tests

### DIFF
--- a/.github/workflows/canary-test.yml
+++ b/.github/workflows/canary-test.yml
@@ -39,7 +39,7 @@ jobs:
           - os: windows-latest
             binary: stripe.exe
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 20
+    timeout-minutes: 25
 
     steps:
       - name: Set git to use LF
--- a/canary/listen_test.go
+++ b/canary/listen_test.go
@@ -121,7 +121,7 @@ func TestAPIListenForwardTo(t *testing.T) {
 	defer listen.Stop()
 
 	// Wait for listen to be ready
-	err = listen.WaitForOutput("Ready!", 30*time.Second)
+	err = listen.WaitForOutput("Ready!", 60*time.Second)
 	if err != nil {
 		stdout, stderr := listen.GetOutput()
 		fatalf(t, "Listen failed to become ready: %v\nStdout: %s\nStderr: %s", err, stdout, stderr)
@@ -201,7 +201,7 @@ func TestAPIListenOutputFormat(t *testing.T) {
 	defer listen.Stop()
 
 	// Wait for listen to be ready
-	err = listen.WaitForOutput("Ready!", 30*time.Second)
+	err = listen.WaitForOutput("Ready!", 60*time.Second)
 	if err != nil {
 		stdout, stderr := listen.GetOutput()
 		fatalf(t, "Listen failed to become ready: %v\nStdout: %s\nStderr: %s", err, stdout, stderr)
--- a/canary/logs_test.go
+++ b/canary/logs_test.go
@@ -112,7 +112,7 @@ func TestAPILogsTailCapture(t *testing.T) {
 	defer logsTail.Stop()
 
 	// Wait for logs tail to be ready
-	err = logsTail.WaitForOutput("Ready!", 30*time.Second)
+	err = logsTail.WaitForOutput("Ready!", 60*time.Second)
 	if err != nil {
 		stdout, stderr := logsTail.GetOutput()
 		fatalf(t, "Logs tail failed to become ready: %v\nStdout: %s\nStderr: %s", err, stdout, stderr)


.github/workflows/canary-test.yml
```diff
@@ -0,0 +1,84 @@
...
+  workflow_call:
+    inputs:
+      run_api_tests:
+        description: 'Run API tests (requires secrets)'
+        required: false
+        default: true
+        type: boolean
+    secrets:
+      STRIPE_TEST_API_KEY:
+        required: false
```
tomer-stripe: hmm i think this should be required

.github/workflows/canary-test.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Run API canary tests
+        # Run API tests when:
+        # 1. On push to master (secrets available)
+        # 2. On manual workflow_dispatch with run_api_tests=true
+        # 3. When called from another workflow (e.g. release)
+        if: |
+          secrets.STRIPE_TEST_API_KEY != '' &&
+          (github.event_name == 'push' ||
+           github.event_name == 'workflow_dispatch' ||
+           github.event_name == 'workflow_call')
```
tomer-stripe: ah i see, required false above but we manually check here. wonder if this will appropriately check if we have keys and not fail silently/open

canary/testutil/runner.go
```diff
@@ -0,0 +1,338 @@
...
+// WithTimeout creates a copy of the runner with a different timeout.
+func (r *Runner) WithTimeout(timeout time.Duration) *Runner {
+	newRunner := *r
+	newRunner.Env = make(map[string]string)
+	for k, v := range r.Env {
+		newRunner.Env[k] = v
+	}
+	newRunner.Timeout = timeout
+	return &newRunner
+}
```
tomer-stripe, line 134: wonder if we want this as part of the `NewRunner` call above. iirc that's a bit more inline with the rest of the design in the cli

canary/login_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"strings"
+	"testing"
+)
+
+// =============================================================================
+// Login Command Tests (Offline only - login requires browser)
+// =============================================================================
+
+func TestOfflineLoginHelp(t *testing.T) {
```
tomer-stripe, line 16: should test the non-interactive login

.github/workflows/canary-test.yml
```diff
@@ -0,0 +1,94 @@
...
+      matrix:
+        include:
+          - os: ubuntu-latest
+            binary: stripe
+          - os: macos-latest
+            binary: stripe
+          - os: windows-latest
+            binary: stripe.exe
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
```
vzhang-stripe: claude recommands to increase 20 to 25m to buffer for slow runners

canary/logs_test.go
```diff
@@ -0,0 +1,162 @@
...
+	}))
+
+	logsTail, err := runner.RunBackground("logs", "tail", "--format", "JSON")
+	if err != nil {
+		fatalf(t, "Failed to start logs tail: %v", err)
+	}
+	defer logsTail.Stop()
+
+	// Wait for logs tail to be ready
+	err = logsTail.WaitForOutput("Ready!", 30*time.Second)
```
vzhang-stripe: claude recommends changing it to 60 because Webhook endpoint initialization and API connection can take longer on slow runners.

canary/listen_test.go
```diff
@@ -0,0 +1,241 @@
...
+	}))
+
+	listen, err := runner.RunBackground("listen", "--format", "JSON")
+	if err != nil {
+		fatalf(t, "Failed to start listen: %v", err)
+	}
+	defer listen.Stop()
+
+	// Wait for listen to be ready
+	err = listen.WaitForOutput("Ready!", 30*time.Second)
```
vzhang-stripe: claude recommends changing it to 60 because Webhook endpoint initialization and API connection can take longer on slow runners.

vcheung-stripe (approved):
LGTM but wondering if there's a way to check that it's passing first? I saw https://github.com/stripe/stripe-cli/actions/workflows/canary-test.yml but I'm not sure if these runs are related to the PR.

PR #1444, opened 2026-02-13
Add AI agent detection to telemetry

matv-stripe:
Adds detection for AI coding agents that invoke the CLI, including Claude Code, Cursor, Cline, and others. The agent name is captured in telemetry events via a new ai_agent field.

Detection is based on well-known environment variables set by each agent. The check happens once per CLI invocation when telemetry metadata is initialized, with no performance impact.

Uses omitempty tag so the field is only sent when an AI agent is detected, keeping telemetry data clean.

Committed-By-Agent: claude

 ### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

### DIFF
--- a/pkg/stripe/analytics_telemetry_test.go
+++ b/pkg/stripe/analytics_telemetry_test.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"os"
 	"testing"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


pkg/stripe/analytics_telemetry_test.go
```diff
@@ -191,3 +255,112 @@ func TestTelemetryOptedOut(t *testing.T) {
 	require.True(t, stripe.TelemetryOptedOut("True"))
 	require.True(t, stripe.TelemetryOptedOut("TRUE"))
 }
+
+// AI Agent Detection Tests
+func TestNewEventMetadata_DetectsAIAgent(t *testing.T) {
+	// Save original env state
+	originalEnv := os.Getenv("CLAUDECODE")
```
vcheung-stripe: Should we mock this module? I wanna make sure the tests aren't actually setting env vars on our system

Dustin4444 (commented):
#1444

PR #1453, opened 2026-03-04
Update the stripe plugin install <plugin> command to detect for and install a JS runtime

vzhang-stripe:
### Reviewers                                                                                                                                                                                           
  r? @                                                      
  cc @stripe/developer-products                                                                                                                                                                           
                                                                                                                                                                                                          
  ### Summary
  Restores execution of TypeScript-based plugins via managed Node.js runtimes. When a plugin specifies a `Runtime` field in its manifest (e.g., `Runtime = { node = "24" }`), the CLI now executes the
  plugin using the managed Node.js installation from `~/.config/stripe/runtimes/node/` instead of relying on system Node.js or shebang resolution.

  This completes the Node.js runtime support feature that was partially implemented in earlier PRs.

  ### Motivation
  The Runtime field was added to support TypeScript-based CLI plugins, allowing the CLI to automatically download and manage Node.js runtimes for plugins that need them. However, the
  execution logic was incomplete - the CLI would download Node.js runtimes during plugin installation but wouldn't actually use them when running plugins.

  Without this change:
  - Plugins with Runtime fields rely on shebang (`#!/usr/bin/env node`) and system Node.js
  - Downloaded managed runtimes go unused
  - Users need compatible Node.js installed in their PATH
  - Version consistency is not guaranteed across environments

  ### Changes
  1. **Added `GetNodeBinaryPath()` to `runtime.go`**: Returns the path to the node executable for a specific major version (e.g., `/Users/user/.config/stripe/runtimes/node/24.0.0/bin/node`)

  2. **Restored runtime-aware execution in `plugin.go`**:
     - Checks if plugin release has a `Runtime` field
     - If Node.js runtime is required, executes via managed Node: `node /path/to/plugin`
     - Falls back to direct execution for native binaries

  3. **Disabled checksum validation for runtime-executed plugins**: When executing via `node plugin_path`, the command path points to the node binary rather than the plugin. Checksum validation would
  fail, so we skip it for plugins that specify a runtime. The plugin bundle itself is still validated during installation.

  ### Testing

Tested End to End by running Node on `stripe-cli-generate-plugin` js bundle (building it using 
  Tested with `stripe-cli-generate-plugin` by running `install:plugin:bundle` [here](https://github.com/stripe/stripe-cli-generate-plugin/blob/master/package.json#L31C5-L31C26) locally on my macbook

step1: clone the generate-plugin [repo](https://github.com/stripe/stripe-cli-generate-plugin/tree/master). 
step2: build it by running `pnpm install:plugin:bundle`
step3: run the Stripe CLI's node on the JS bundle `Stripe-dev generate --help`

TODO: test it on windows



  Verified:
  - ✅ Plugins with Runtime field execute via managed Node.js
  - ✅ Plugins without Runtime field execute directly (native binaries)
  - ✅ Commands and subcommands work correctly
  - ✅ Works with both installed and local.build.dev plugins
  - ✅ Falls back gracefully when release info unavailable

  Security Considerations

  Checksum validation is skipped for runtime-executed plugins because:
  - The command path points to the managed Node binary, not the plugin
  - Plugin bundles are still validated during installation via getChecksum()
  - Managed runtimes are validated via hardcoded checksums during download (see runtime.go)
  - This is consistent with how native binaries with embedded interpreters work

### DIFF
--- a/scripts/update-node-checksums.sh
+++ b/scripts/update-node-checksums.sh
@@ -17,6 +17,17 @@ fi
 # Remove 'v' prefix if present
 VERSION="${VERSION#v}"
 
+# Store the original working directory and runtime file path
+ORIGINAL_DIR="$(pwd)"
+RUNTIME_FILE="${ORIGINAL_DIR}/pkg/plugins/runtime.go"
+
+# Check if runtime.go exists
+if [ ! -f "$RUNTIME_FILE" ]; then
+    echo "Error: Could not find runtime.go at $RUNTIME_FILE"
+    echo "Make sure you run this from the project root: make update-node-checksums VERSION=X.Y.Z"
+    exit 1
+fi
+
 # Create temp directory
 TMPDIR=$(mktemp -d)
 trap "rm -rf $TMPDIR" EXIT
@@ -66,13 +77,11 @@ MAJOR_VERSION=$(echo "$VERSION" | cut -d. -f1)
 # Get current date
 CURRENT_DATE=$(date +%Y-%m-%d)
 
-# Generate Go code
-echo ""
-echo "================================================"
-echo "Add the following to pkg/plugins/runtime.go:"
-echo "================================================"
-echo ""
-cat <<EOF
+# Write the new Go code entry to a temporary file
+NEW_ENTRY_FILE=$(mktemp)
+trap "rm -f $NEW_ENTRY_FILE" EXIT
+
+cat > "$NEW_ENTRY_FILE" <<EOF
 	"${MAJOR_VERSION}": {
 		Version: "${VERSION}",
 		// Checksums verified from https://nodejs.org/dist/v${VERSION}/SHASUMS256.txt
@@ -92,8 +101,61 @@ cat <<EOF
 		},
 	},
 EOF
+
 echo ""
-echo "================================================"
+echo "Updating ${RUNTIME_FILE}..."
+
+# Check if the major version already exists
+if grep -q "\"${MAJOR_VERSION}\":" "$RUNTIME_FILE"; then
+    echo "Version ${MAJOR_VERSION} already exists in runtime.go"
+    echo "Replacing existing entry..."
+
+    # Use awk to replace the existing entry
+    awk -v major="$MAJOR_VERSION" -v newfile="$NEW_ENTRY_FILE" '
+    BEGIN { in_block=0; skip=0 }
+    {
+        if ($0 ~ "\"" major "\":") {
+            in_block=1
+            skip=1
+            # Read and print the new entry from file
+            while ((getline line < newfile) > 0) {
+                print line
+            }
+            close(newfile)
+            next
+        }
+        if (in_block && $0 ~ /^\t},/) {
+            in_block=0
+            skip=0
+            next
+        }
+        if (!skip) {
+            print $0
+        }
+    }
+    ' "$RUNTIME_FILE" > "${RUNTIME_FILE}.tmp"
+
+    mv "${RUNTIME_FILE}.tmp" "$RUNTIME_FILE"
+else
+    echo "Adding new version ${MAJOR_VERSION} to runtime.go..."
+
+    # Insert the new entry before the closing brace of the map
+    # Find the line with just "}" that closes the nodeRuntimeConfigs map
+    awk -v newfile="$NEW_ENTRY_FILE" '
+    /^}$/ && prev_line ~ /^\t},/ {
+        # Read and print the new entry from file
+        while ((getline line < newfile) > 0) {
+            print line
+        }
+        close(newfile)
+    }
+    { print; prev_line=$0 }
+    ' "$RUNTIME_FILE" > "${RUNTIME_FILE}.tmp"
+
+    mv "${RUNTIME_FILE}.tmp" "$RUNTIME_FILE"
+fi
+
+echo "✓ Successfully updated ${RUNTIME_FILE}"
 echo ""
 echo "Summary:"
 echo "  Version: ${VERSION}"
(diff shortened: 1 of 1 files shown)


pkg/plugins/runtime.go
```diff
@@ -0,0 +1,351 @@
...
+	Checksums map[string]map[string]string // OS -> Arch -> Checksum
+}
+
+// Hardcoded Node.js LTS runtime configurations
+// Using Node.js 20.x LTS (Iron) as the default runtime
+//
+// Checksums are verified from official Node.js distribution over HTTPS.
+// For maximum security, checksums can also be verified against GPG signatures.
+//
+// To update checksums for a new Node.js version:
```
vcheung-stripe: Nit: can you make this a script that we can run? Maybe add a target to the makefile?

pkg/plugins/runtime.go
```diff
@@ -0,0 +1,351 @@
...
+// InstallNodeRuntime downloads and installs a specific Node.js runtime version
+func InstallNodeRuntime(ctx context.Context, cfg config.IConfig, fs afero.Fs, majorVersion string) error {
+	runtimeConfig, exists := nodeRuntimeConfigs[majorVersion]
+	if !exists {
+		return fmt.Errorf("unsupported Node.js version: %s", majorVersion)
+	}
+
+	// Check if runtime is already installed
+	if IsRuntimeInstalled(cfg, fs, majorVersion) {
+		return nil
```
vcheung-stripe, line 196: Can you add some debug log lines for whether we found one or are about to install one? This will help with debugging.

pkg/plugins/runtime.go
```diff
@@ -0,0 +1,351 @@
...
+	baseURL := "https://nodejs.org/dist"
+
+	// Map Go arch names to Node.js arch names
+	nodeArch := arch
+	if arch == "amd64" {
+		nodeArch = "x64"
+	}
+
+	var filename string
+	switch opsys {
```
vcheung-stripe, line 262: Can you return an error if the os or arch are unsupported?

pkg/plugins/runtime.go
```diff
@@ -0,0 +1,351 @@
...
+	// Create destination directory
+	if err := fs.MkdirAll(destPath, 0755); err != nil {
+		return fmt.Errorf("failed to create runtime directory: %w", err)
+	}
+
+	switch opsys {
+	case "darwin", "linux":
+		return extractTarGz(fs, data, destPath)
+	case "windows":
+		return fmt.Errorf("Windows runtime extraction not yet implemented")
```
vcheung-stripe: What should we do on windows?

pkg/plugins/runtime.go
```diff
@@ -0,0 +1,287 @@
...
+	gzr, err := gzip.NewReader(strings.NewReader(string(data)))
+	if err != nil {
+		return fmt.Errorf("failed to create gzip reader: %w", err)
+	}
+	defer gzr.Close()
+
+	tr := tar.NewReader(gzr)
+
+	for {
+		header, err := tr.Next()
```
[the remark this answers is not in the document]
vcheung-stripe: Can you address this?

pkg/plugins/utilities.go
```diff
@@ -167,6 +167,57 @@ func fetchPluginList(baseURL, manifestFilename string) (*PluginList, error) {
 	return validatePluginManifest(body)
 }
 
+// validateRuntimeVersions validates that Runtime specifications only contain valid LTS Node.js versions
+func validateRuntimeVersions(pluginList *PluginList) error {
```
vcheung-stripe, line 171: Nit: can we make this less nested for readability?

pkg/plugins/utilities.go
```diff
@@ -167,6 +167,57 @@ func fetchPluginList(baseURL, manifestFilename string) (*PluginList, error) {
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// isValidNodeLTSVersion checks if a Node.js version string is a valid LTS major version
+// Valid LTS versions are even-numbered major versions starting from 18
+func isValidNodeLTSVersion(version string) bool {
```
vcheung-stripe, line 212: If we detect that the version is "invalid," can we tell the user what they should do to resolve it? For example:
- If the core CLI supports 24, 26, 28 but the plugin requires 30, then the core CLI is too old -> tell user to update the Stripe CLI
- If the core CLI supports 24, 26, 28 but the plugin requires 22, then the plugin is too old -> tell user to update the plugin

vcheung-stripe: This can be in a followup

vcheung-stripe (changes requested):

pkg/plugins/runtime.go
vzhang-stripe: good catch! Just added the windows extract function

.github/workflows/test.yml
```diff
@@ -29,6 +29,7 @@ jobs:
     - name: Install protoc
       uses: arduino/setup-protoc@v3
       with:
+        version: '28.3'
```
vcheung-stripe, line 32: Is this needed?

vzhang-stripe: it was failing the windows test without

vzhang-stripe: This version is needed because:                                                                                                                                                                         
                  
  1. Reproducible builds - Pins the exact protoc version so all CI runs and developers use the same version, preventing "works on my machine" issues                                                      
  2. Prevents breaking changes - Newer protoc versions can generate different code or have incompatible changes. Pinning the version ensures the generated .pb.go files remain consistent                 
  3. Matches the codebase - The stripe-cli uses protobuf for its RPC layer (the rpc/*.pb.go files). This version likely matches what was used to generate those files initially
  4. CI stability - Without a version pin, the workflow would use whatever the latest version is, which could cause unexpected CI failures when protoc releases new versions

PR #1458, opened 2026-03-05
allow installing some plugins without being logged in

vcheung-stripe:
### Reviewers
cc @stripe/developer-products

 ### Summary

This allows plugins to be installed without being logged in. The reason we were requiring auth before was to support incrementally roll out plugins to subsets of users, but this is (1) not necessary for plugins today and (2) not necessary for some new plugins in the future. So, let's skip a step to make the process more streamlined.

For plugins that we do want to roll out incrementally, we keep the original path that requires that the user be logged in. In summary:
- If logged in, call old path with auth
- If not logged in, call new path without auth

#### Not logged in

I ran this series of commands and confirmed it works e2e and the expected debug line I added was printed:
```
# Simulate brand new user by deleting ~/.config/stripe/plugins and ~/.config/stripe/plugins.toml and running stripe logout

# First time install
go run cmd/stripe/main.go plugin install apps --log-level debug

# Second time install
go run cmd/stripe/main.go plugin install apps --log-level debug

# Reinstall at a later time
stripe plugin uninstall apps
go run cmd/stripe/main.go plugin install apps --log-level debug
```

#### Logged in

Similarly, I reran that series of commands after logging in and confirmed it works and the expected debug line was printed:
```
# Simulate brand new *authenticated* user by deleting ~/.config/stripe/plugins and ~/.config/stripe/plugins.toml and running *stripe login*

# First time install
go run cmd/stripe/main.go plugin install apps --log-level debug

# Second time install
go run cmd/stripe/main.go plugin install apps --log-level debug

# Reinstall at a later time
stripe plugin uninstall apps
go run cmd/stripe/main.go plugin install apps --log-level debug
```

### DIFF
--- a/pkg/plugins/utilities.go
+++ b/pkg/plugins/utilities.go
@@ -24,6 +24,7 @@ import (
 	"github.com/stripe/stripe-cli/pkg/config"
 	"github.com/stripe/stripe-cli/pkg/requests"
 	"github.com/stripe/stripe-cli/pkg/stripe"
+	"github.com/stripe/stripe-cli/pkg/validators"
 )
 
 // GetBinaryExtension returns the appropriate file extension for plugin binary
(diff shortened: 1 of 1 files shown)


vzhang-stripe (commented):
Error Handling Concern

  The error suppression pattern used in this PR could hide legitimate errors:

```
  // pkg/plugins/plugin.go and pkg/plugins/utilities.go
  - apiKey, err := cfg.GetProfile().GetAPIKey(false)
  - if err != nil { return err }
  + apiKey, _ := cfg.GetProfile().GetAPIKey(false)

```
  Issue: Using _ treats all errors the same - whether it's "user not logged in" (expected) or "config file corrupted" (critical). This makes debugging harder and could lead to silent failures.

  Scenario: If a user's config file is corrupted, the install will silently succeed using fallback URLs, but the user won't know their config needs fixing.

  Suggested fix:

```
  apiKey, err := cfg.GetProfile().GetAPIKey(false)
  if err != nil {
      // Only allow unauthenticated installs when user is explicitly not logged in
      // Return other errors (config corruption, permission issues, etc.)
      if !IsNotAuthenticatedError(err) {
          return fmt.Errorf("failed to read configuration: %w", err)
      }
      apiKey = "" // User not logged in - use fallback
  }
```

  Additional testing suggestion: Consider adding a test case for corrupted config files to ensure real errors aren't silently ignored.

  Otherwise the approach looks good - backward compatible and achieves the stated goal!

vzhang-stripe (approved):
niceee

PR #1481, opened 2026-03-12
Add --map global flag to print recursive command tree

tomer-stripe:
## Summary

- Adds a `--map` PersistentFlag that prints a tree-view sitemap of all available subcommands from any point in the hierarchy using box-drawing characters (e.g. `stripe --map`, `stripe issuing --map`, `stripe customers --map`)
- Adds plugin subcommand introspection via manifest metadata: plugins can declare a `[[Plugin.Command]]` tree in `plugins.toml`, and the CLI creates cobra.Command stubs so plugin subcommands appear in `--map` output without launching the plugin binary
- Handles `--map` interception before Cobra's `ExecuteContext` so it works for non-runnable commands (root, namespaces) where `PersistentPreRun` never fires
- Properly handles `--map=true`/`--map=false` forms and rejects substring matches like `--mapfile`

### Example output

```
stripe issuing
├── authorizations
│   ├── list
│   ├── retrieve
│   └── update
├── cards
│   ├── create
│   ├── list
│   ├── retrieve
│   └── update
└── disputes
    ├── list
    └── retrieve
```

### Related PRs
- Go plugin bootstrap: stripe/stripe-cli-plugin-bootstrap (branch: tomer/add-map-command-tree)
- TS plugin bootstrap: stripe/stripe-cli-ts-plugin-bootstrap (branch: tomer/add-map-command-tree)

## Test plan

- [x] `go test ./pkg/cmd/... ./pkg/plugins/...` — all tests pass
- [x] `go build ./cmd/stripe && ./stripe --map` — full tree from root
- [x] `./stripe issuing --map` — scoped subtree
- [x] `./stripe customers --map` — resource with operations
- [x] `./stripe version --map` — leaf command, just prints command path
- [x] `./stripe --map | grep payment` — piping works, ANSI disabled


pkg/plugins/test_artifacts/plugins.toml
```diff
@@ -108,3 +108,46 @@
...
     Sum = "2f05d4b689d270cafb02285f35f44866f7dc8a2d368a3f9d1124373eeab31fb1"
 
+# appC has subcommand metadata for --map testing
+[[Plugin]]
+  Shortname = "appC"
+  Shortdesc = "A plugin with subcommands"
+  Binary = "stripe-cli-app-c"
+  MagicCookieValue = "A1B2C3D4-E5F6-7890-ABCD-EF1234567890"
+
+  [[Plugin.Command]]
```
vcheung-stripe, line 118: Looks good - since the command tree is per _plugin_ instead of per _plugin release_ then this should be fine size-wise.

PR #1507, opened 2026-03-23
Add logic to prevent symlink usage in samples

ianjabour-stripe:
### Reviewers
cc @stripe/developer-products

 ### Summary
This PR adds logic to prevent symlinks from be used within the `stripe samples create` command. This avoids situations where a compromised sample might attempt to link outside the bounds of the destination directory.

### DIFF
--- a/.goreleaser/linux.yml
+++ b/.goreleaser/linux.yml
@@ -57,11 +57,6 @@ nfpms:
     license: Apache 2.0
     formats:
       - deb
-    dependencies:
-      - libsecret-1-0
-      - gnome-keyring
-      - libsecret-tools
-      - dbus-x11
   - id: rpm
     package_name: stripe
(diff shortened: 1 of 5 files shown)


pkg/samples/samples.go
```diff
@@ -360,6 +369,7 @@ func (s *SampleManager) WriteDotEnv(ctx context.Context, sampleLocation string)
 		if err != nil {
 			return err
 		}
+		defer file.Close()
```
ianjabour-stripe, line 372: Because Windows locks open files, we need to explicitly close this. Otherwise we can run into `The process cannot access the file because it is being used by another process.`, which is why the windows CI tests seemed to fail here:
https://github.com/stripe/stripe-cli/actions/runs/23458831498/job/68279751448

PR #1517, opened 2026-03-26
feat: add auth_available and requires_confirmation to --dry-run output

bernerd-stripe:
### Reviewers
r?
cc @stripe/developer-products

### Summary
Adds two preflight fields to `DryRunDetails` so orchestrating agents can determine whether credentials are resolved and whether the operation requires confirmation before executing.

- `auth_available`: `true` when a non-empty API key was resolved
- `requires_confirmation`: `true` for DELETE requests (reuses the existing `confirmationCommands` map)

Also updates the `--dry-run` flag description to mention the new fields, and adds `--dry-run` usage examples to `get`, `post`, and `delete` commands.

### DIFF
--- a/pkg/requests/base_test.go
+++ b/pkg/requests/base_test.go
@@ -565,6 +565,24 @@ func TestBuildDryRunOutput_V2Post(t *testing.T) {
 	}}, *output)
 }
 
+func TestBuildDryRunOutput_Delete(t *testing.T) {
+	rb := Base{Method: http.MethodDelete}
+
+	output, err := rb.BuildDryRunOutput("sk_test_abcd", "https://api.stripe.com", "/v1/customers/cus_abc123", &RequestParameters{}, map[string]interface{}{})
+	require.NoError(t, err)
+	require.Equal(t, DryRunOutput{DryRun: DryRunDetails{
+		Method: "DELETE",
+		URL:    "https://api.stripe.com/v1/customers/cus_abc123",
+		Params: map[string]interface{}{},
+		Headers: map[string]string{
+			"Authorization": "Bearer sk_test_abcd",
+			"Content-Type":  "application/x-www-form-urlencoded",
+		},
+		AuthAvailable:        true,
+		RequiresConfirmation: true,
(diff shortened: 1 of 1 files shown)


pkg/requests/base.go
```diff
@@ -124,10 +124,12 @@ type Base struct {
...
-	Method  string                 `json:"method"`
-	URL     string                 `json:"url"`
-	Params  map[string]interface{} `json:"params"`
-	Headers map[string]string      `json:"headers"`
+	Method               string                 `json:"method"`
+	URL                  string                 `json:"url"`
+	Params               map[string]interface{} `json:"params"`
+	Headers              map[string]string      `json:"headers"`
+	AuthAvailable        bool                   `json:"auth_available"`
+	RequiresConfirmation bool                   `json:"requires_confirmation"`
```
tomer-stripe, line 132: How are you thinking about the dry run feature generally? In my head I would've assumed it reflected the API request itself and this feels like more request meta

bernerd-stripe: @tomer-stripe Yeah, that's fair. I was trying to improve the loop of:

* `stripe customers delete cus_123 --dry-run` => review that it looks right
* `stripe customers delete cus_123` => fails with error because you need an API key/login
* `stripe customers delete cus_123 --api-key=$API_KEY` => fails because you also need to confirm deletion
* `stripe customers delete cus_123 --api-key=$API_KEY --confirm` => actually works

One alternative would be to print info about auth and whether confirmation is required as notices on stderr, which would create a clearer separation between CLI meta info and the request data. Do you have a preference?

bernerd-stripe: Claude also suggested adding a "preflight" key for surfacing checks not related to the request itself:

```code
 {
   "dry_run": {
     "method": "DELETE",
     "url": "https://api.stripe.com/v1/customers/cus_abc123",
     "params": {},
     "headers": { "Authorization": "..." }
   },
   "preflight": {
     "auth_available": true,
     "requires_confirmation": true
   }
 }
```

PR #1518, opened 2026-03-26
Implement `stripe plugin auto-update` command

vcheung-stripe:
### Summary

Implement the `stripe plugin auto-update` command to set configs for plugins. It supports configs for all plugins or just specific ones. I set this to Hidden because this config is currently unused, but we will use to configure enabling/disabling plugin auto-updates, which will be implemented in https://github.com/stripe/stripe-cli/pull/1519.

Help:
```
Enable or disable automatic background updates for a plugin.

When disabled, the CLI will not check for or download newer versions automatically.
Omit the plugin name to apply the setting globally to all plugins.

Usage:
  stripe plugin auto-update [plugin] [flags]

Examples:
  stripe plugin auto-update --enable
  stripe plugin auto-update --disable
  stripe plugin auto-update apps --enable
  stripe plugin auto-update apps --disable

Flags:
      --disable   Disable automatic updates
      --enable    Enable automatic updates
  -h, --help      help for auto-update

Global flags:
      --config string         config file (default is $HOME/.config/stripe/config.toml)
      --log-level string      log level (debug, info, trace, warn, error) (default "info")
  -p, --project-name string   the project name to read from for config (default "default")
```

For example, if you want auto updates for every plugin except apps, you'd do:
```
stripe plugin auto-update --enable
stripe plugin auto-update apps --disable
```

Which will write this to the config.toml:
```
[plugin_configs]
[plugin_configs.__global]
updates = 'on'

[plugin_configs.apps]
updates = 'off'
```

### DIFF
--- a/pkg/proxy/endpoint.go
+++ b/pkg/proxy/endpoint.go
@@ -156,7 +156,7 @@ func (c *EndpointClient) PostV2(evtCtx eventContext) error {
 		}
 	}
 
-	resp, err := http.DefaultClient.Do(req)
+	resp, err := c.cfg.HTTPClient.Do(req)
 	if err != nil {
 		c.cfg.OutCh <- websocket.ErrorElement{
 			Error: FailedToPostError{Err: err},
--- a/pkg/proxy/endpoint_test.go
+++ b/pkg/proxy/endpoint_test.go
@@ -211,3 +211,42 @@ func TestClientHandler_EventDestination(t *testing.T) {
 	require.Equal(t, "", rcvCtx.webhookConversationID)
 	require.Equal(t, "evt_123", rcvCtx.v2Event.ID)
 }
+
+func TestPostV2_UsesConfiguredHTTPClient(t *testing.T) {
+	wg := &sync.WaitGroup{}
+	wg.Add(1)
+
+	ts := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("OK!"))
+	}))
+	defer ts.Close()
+
+	httpClient := ts.Client()
+
+	client := NewEndpointClient(
+		ts.URL,
+		[]string{},
+		false,
+		[]string{"*"},
+		true,
+		&EndpointConfig{
+			HTTPClient: httpClient,
+			ResponseHandler: EndpointResponseHandlerFunc(func(evtCtx eventContext, forwardURL string, resp *http.Response) {
+				wg.Done()
+			}),
+		},
+	)
+
+	evtCtx := eventContext{
+		v2Event:        &V2EventPayload{ID: "evt_456"},
+		requestBody:    "{}",
+		requestHeaders: map[string]string{},
+	}
+
+	// this would fail with x509 error if PostV2 used http.DefaultClient
+	err := client.PostV2(evtCtx)
+	require.NoError(t, err)
+
+	wg.Wait()
+}


pkg/cmd/plugin/auto_update.go
```diff
@@ -0,0 +1,68 @@
...
+		Long: `Enable or disable automatic background updates for a plugin.
+
+By default, automatic updates are disabled. When disabled, the CLI will not check
+for or download newer versions automatically.
+Omit the plugin name to apply the setting globally to all plugins.`,
+		Example: `stripe plugin auto-update --enable
+  stripe plugin auto-update --disable
+  stripe plugin auto-update apps --enable
+  stripe plugin auto-update apps --disable`,
+		RunE:   ac.run,
```
bernerd-stripe, line 39: Suggest adding the max args validator (otherwise `stripe plugin auto-update apps bar --enable` works but the extra args are ignored.

```suggestion
		Args:   validators.MaximumNArgs(1),
		RunE:   ac.run,
```

pkg/config/config.go
bernerd-stripe: I noticed that this doesn't use `isProfile` but is also stricter and only considers a profile to be a profile if it has a display name? Though I guess I don't really know what happens if display_name is blank...

Anyway, this all comes up because I think both by this definition and the one in `isProfile`, the new plugin configs will be considered "profiles" because they'll match `value.(map[string]interface{})`. What do you think about updating `isProfile` (and using it here and anywhere else) so that:
1. (preference) a config entry is only a profile if it's a map with a non-nil `display_name` field
2. we maintain an exclusion list of reserved non-profile keys which isProfile then never considers to be a profile

vcheung-stripe: > What do you think about updating isProfile (and using it here and anywhere else) so that ... a config entry is only a profile if it's a map with a non-nil display_name field

Love it!

bernerd-stripe (approved):

pkg/config/config_test.go
```diff
@@ -78,6 +78,32 @@ func TestCopyProfile(t *testing.T) {
 	require.Equal(t, "backup", v.GetString("backup.profile_name"))
 }
 
+func TestCopyProfile_SkipsPluginConfigsField(t *testing.T) {
```
vcheung-stripe, line 109: TODO: this test is not exercising the right thing, need to fix a bug in the copying logic.

pkg/config/config.go
```diff
@@ -469,7 +471,7 @@ func (c *Config) WriteConfigField(field string, value interface{}) error {
 	runtimeViper := viper.GetViper()
 	runtimeViper.Set(field, value)
 
-	return runtimeViper.WriteConfig()
+	return writeConfig(runtimeViper)
```
vcheung-stripe, line 474: This was needed to get the test to pass. Surprisingly, the only caller of this function is the plugins package, so I tested that out to make sure it still works.

vcheung-stripe:
@bernerd-stripe ptal. I had to fix a bug where if you log in with `-p plugin_configs` then log in again, we incorrectly copy the plugin configs to the backup profile. Fixed in this commit: https://github.com/stripe/stripe-cli/pull/1518/commits/dcb446e705c51f67517b24225b58ee50bbe8d53a (The test wasn't asserting the right thing originally, thanks claude.)

Tested logging in with `-p plugin_configs` and logging in again and saw the `plugin_configs.apps.updates` did not get copied to the backup profile.

```
go run cmd/stripe/main.go login -p plugin_configs
```
```
cat ~/.config/stripe/config.toml
color = ''
installed_plugins = []
project-name = 'plugin_configs'

[plugin_configs]
account_id = 'acct_1NRKwLLJDmqA11cn'
device_name = 'st-vcheung4'
display_name = "Tomer's Chicken n' Waffles (v1 Workbench)"
...

[plugin_configs.apps]
updates = 'on'
```
```
go run cmd/stripe/main.go login
```
```
cat ~/.config/stripe/config.toml
color = ''
installed_plugins = []
project-name = 'plugin_configs'

[plugin_configs]
account_id = 'acct_1Plz52CjOw7iF6x1'
device_name = 'st-vcheung4'
display_name = 'kliu'
...
[plugin_configs.apps]
updates = 'on'

["tomer's chicken n' waffles (v1 workbench)"]
account_id = 'acct_1NRKwLLJDmqA11cn'
device_name = 'st-vcheung4'
display_name = "Tomer's Chicken n' Waffles (v1 Workbench)"
...
```

bernerd-stripe (approved):

PR #1519, opened 2026-03-27
Support automatically updating plugins

vcheung-stripe:
### Summary

When running a plugin command, check if there's a new version and install it in the background while the current command is running. We read the config added in https://github.com/stripe/stripe-cli/pull/1518 to decide if we should do this or not (default off).

How it works:
- The user runs a plugin command, e.g. `stripe projects`.
- We run a goroutine to refresh the plugin manifest, check for a new version within the same major, and install it if there is one.
  - At the same time, we write progress messages to a buffer.
- The command proceeds as normal in the main routine.
- When the command finishes, we'll print the progress messages that we saved in the buffer.

To test this, try installing an old version of the projects plugin and then invoking it:
```
go run cmd/stripe/main.go plugin install projects@0.0.58
go run cmd/stripe/main.go projects --version
```

TODO:
- [ ] Check for updates on an interval instead of every invocation (once per day)?
- [ ] Don't check again if the user requested a specific version (maybe wait 1-7 days)?
- [ ] How should we handle compatibility? (maybe plugin defines a min CLI version?)

### DIFF
--- a/pkg/plugins/updater.go
+++ b/pkg/plugins/updater.go
@@ -1,0 +1,152 @@
+package plugins
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"path/filepath"
+
+	"github.com/hashicorp/go-version"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"github.com/spf13/viper"
+
+	"github.com/stripe/stripe-cli/pkg/ansi"
+	"github.com/stripe/stripe-cli/pkg/config"
+	"github.com/stripe/stripe-cli/pkg/gatedwriter"
+)
+
+// WithBackgroundUpdate runs fn (typically plugin.Run) while concurrently
+// checking for and applying a plugin update. Update output is buffered until
+// fn returns, then flushed to stderr so it never interleaves with plugin output.
+func WithBackgroundUpdate(ctx context.Context, cfg config.IConfig, fs afero.Fs, baseURL string, plugin *Plugin, out io.Writer, fn func() error) error {
+	w := gatedwriter.NewGatedWriter(out, 0)
+	cleanupReady := make(chan struct{})
+	updateDone := make(chan struct{})
+
+	go func() {
+		checkAndUpdate(ctx, cfg, fs, plugin, w, baseURL, cleanupReady)
+		close(updateDone)
+	}()
+
+	err := fn()
+
+	w.Open()
+	CleanupAllClients()
+	close(cleanupReady)
+	<-updateDone
+
+	return err
+}
+
+// checkAndUpdate refreshes the plugin manifest and, if a newer version is
+// available, downloads and installs it.
+func checkAndUpdate(ctx context.Context, cfg config.IConfig, fs afero.Fs, plugin *Plugin, out io.Writer, baseURL string, cleanupReady chan struct{}) {
+	logger := log.WithFields(log.Fields{"prefix": "plugins.updater"})
+
+	if !updatesEnabled(plugin.Shortname) {
+		logger.Debugf("Automatic updates disabled")
+		return
+	}
+	logger.Debugf("Automatic updates enabled")
+
+	currentVersion, err := installedPluginVersion(cfg, fs, plugin)
+	if err != nil {
+		logger.Debugf("Error getting installed plugin version: %s", err)
+		return
+	}
+	if currentVersion == "" {
+		logger.Debugf("No installed plugin version found")
+		return
+	}
+	if currentVersion == "local.build.dev" {
+		logger.Debugf("Local build dev version found, skipping update check")
+		return
+	}
+
+	spinner := ansi.StartNewSpinner(ansi.Faint(fmt.Sprintf("checking for '%s' updates...", plugin.Shortname)), out)
+
+	if err := RefreshPluginManifest(ctx, cfg, fs, baseURL); err != nil {
+		ansi.StopSpinner(spinner, ansi.Faint(fmt.Sprintf("'%s' update failed: couldn't find the plugins list", plugin.Shortname)), out)
+		return
+	}
+
+	// Re-look up the plugin so we have the freshest release list.
+	fresh, err := LookUpPlugin(ctx, cfg, fs, plugin.Shortname)
+	if err != nil {
+		ansi.StopSpinner(spinner, ansi.Faint(fmt.Sprintf("'%s' update failed: couldn't find the plugin", plugin.Shortname)), out)
+		return
+	}
+
+	current, err := version.NewVersion(currentVersion)
+	if err != nil {
+		ansi.StopSpinner(spinner, ansi.Faint(fmt.Sprintf("'%s' update failed: couldn't parse current version", plugin.Shortname)), out)
+		return
+	}
+
+	latestVersion := fresh.LookupLatestVersionForMajor(current.Segments()[0])
+	if latestVersion == "" {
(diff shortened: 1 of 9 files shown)


pkg/plugins/updater.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/stripe/stripe-cli/pkg/ansi"
+	"github.com/stripe/stripe-cli/pkg/config"
+	"github.com/stripe/stripe-cli/pkg/gatedwriter"
+)
+
+// WithBackgroundUpdate runs fn (typically plugin.Run) while concurrently
+// checking for and applying a plugin update. Update output is buffered until
+// fn returns, then flushed to stderr so it never interleaves with plugin output.
+func WithBackgroundUpdate(ctx context.Context, cfg config.IConfig, fs afero.Fs, baseURL string, plugin *Plugin, out io.Writer, fn func() error) error {
+	w := gatedwriter.NewGatedWriter(out, 0)
```
bernerd-stripe, line 23: Based on how it's used, I think the gated writer could be greatly simplified to something like:

```go
type terminalBuffer struct {
	bytes.Buffer
	out io.Writer
}

func (b *terminalBuffer) Fd() uintptr {
	if f, ok := b.out.(*os.File); ok {
		return f.Fd()
	}
	return ^uintptr(0)
}
```

and then here we'd do:

```go
buf := &terminalBuffer{out: out}
// ... run background update writing to buf ...
io.Copy(out, buf)
```

pkg/plugins/updater.go
```diff
@@ -0,0 +1,152 @@
...
+	close(cleanupReady)
+	<-updateDone
+
+	return err
+}
+
+// checkAndUpdate refreshes the plugin manifest and, if a newer version is
+// available, downloads and installs it.
+func checkAndUpdate(ctx context.Context, cfg config.IConfig, fs afero.Fs, plugin *Plugin, out io.Writer, baseURL string, cleanupReady chan struct{}) {
+	logger := log.WithFields(log.Fields{"prefix": "plugins.updater"})
```
bernerd-stripe, line 45: Should this also be synced/buffered (I guess we'd need separate err/out writers)? logrus is already safe to operate concurrently—but only w/ other logrus logs. You'd likely see badly interleaved output running w/ verbose logs on without it.

bernerd-stripe: nit: probably want `shortname` as a log field?

pkg/plugins/updater.go
```diff
@@ -0,0 +1,152 @@
...
+		return
+	}
+	if currentVersion == "local.build.dev" {
+		logger.Debugf("Local build dev version found, skipping update check")
+		return
+	}
+
+	spinner := ansi.StartNewSpinner(ansi.Faint(fmt.Sprintf("checking for '%s' updates...", plugin.Shortname)), out)
+
+	if err := RefreshPluginManifest(ctx, cfg, fs, baseURL); err != nil {
```
bernerd-stripe, line 69: It seems high-priority to apply some sort of caching here so that every plugin command invocation doesn't introduce an extra network call. I'd probably consider two things:
1. Only call checkAndUpdate / refresh if it's been at least 15 minutes (or whatever threshold we want) since last checked.
2. Explore using `If-Modified-Since` and/or `If-None-Match` HTTP caching headers so that if the manifest hasn't changed (which will be most of the time) we just get back an empty 304 response rather than having to rewrite the manifest.

(1) is probably blocking, (2) is more nice-to-have

pkg/plugins/updater.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if latest.Segments()[0] != current.Segments()[0] {
+		ansi.StopSpinner(spinner, ansi.Faint(fmt.Sprintf("Skipping update for '%s' because you're on the latest major version", plugin.Shortname)), out)
+		return
+	}
+
+	if !latest.GreaterThan(current) {
+		ansi.StopSpinner(spinner, ansi.Faint(fmt.Sprintf("'%s' is already up to date", plugin.Shortname)), out)
+		return
+	}
```
bernerd-stripe, line 107: These seem like debug-level notices, rather than things we'd want on every invocation?

pkg/plugins/updater.go
```diff
@@ -0,0 +1,152 @@
...
+	ansi.StartSpinner(spinner, ansi.Faint(fmt.Sprintf("updating '%s' to v%s...", plugin.Shortname, latestVersion)), out)
+
+	if err := fresh.install(ctx, cfg, fs, latestVersion, baseURL, io.Discard, false); err != nil {
+		ansi.StopSpinner(spinner, ansi.Faint(fmt.Sprintf("'%s' update to v%s failed", plugin.Shortname, latestVersion)), out)
+		return
+	}
+
+	// Wait for cleanup to be ready because you can't delete a running executable on Windows.
+	<-cleanupReady
+	fresh.cleanUpPluginPath(cfg, fs, latestVersion)
```
bernerd-stripe, line 118: 👍 this is neat

pkg/plugins/updater.go
```diff
@@ -0,0 +1,152 @@
...
+	if currentVersion == "" {
+		logger.Debugf("No installed plugin version found")
+		return
+	}
+	if currentVersion == "local.build.dev" {
+		logger.Debugf("Local build dev version found, skipping update check")
+		return
+	}
+
+	spinner := ansi.StartNewSpinner(ansi.Faint(fmt.Sprintf("checking for '%s' updates...", plugin.Shortname)), out)
```
bernerd-stripe, line 67: I feel like this would be super annoying if it's any more frequent than, like, daily? Even then..

A slightly different way to think about the goals of the communication: I feel like it matters if we've installed/failed to install a new version or if for whatever reason it's taking significantly longer than the main command.

I wonder if we could leverage that in some way? Like, if the other command is still running then we don't really need to print anything. So you could imagine something like:

```go
refreshDone := make(chan err, 1)
go(func() { refreshDone <- RefreshPluginManifest(ctx, cfg, fs, baseURL) })

select {
case refreshErr = <-refreshDone:
  // refresh completed before command so no need to print anything unless we end up actually installing a new version
case <-cleanupReady:
  // hm, main command finished first
  select {
    case refreshErr = <-refreshDone:
      // refresh still finished pretty quickly so just keep moving
    case <-time.After(time.Second):
      // user is now definitely waiting for the refresh to complete, now, so start a spinner and wait
      spinner = ansi.StartNewSpinner(...)
      refreshErr = <-refreshDone
  }
}

if refreshErr != nil {
  // handle err
}
```

pkg/plugins/updater.go
```diff
@@ -0,0 +1,152 @@
...
+	CleanupAllClients()
+	close(cleanupReady)
+	<-updateDone
+
+	return err
+}
+
+// checkAndUpdate refreshes the plugin manifest and, if a newer version is
+// available, downloads and installs it.
+func checkAndUpdate(ctx context.Context, cfg config.IConfig, fs afero.Fs, plugin *Plugin, out io.Writer, baseURL string, cleanupReady chan struct{}) {
```
bernerd-stripe, line 44: One thing to consider is how `ctrl-c` is intended to interact with this process. Especially w/ something like `stripe apps serve`. I don't know how we have that all wired up, but often the setup (or desired setup) is for ctrl-c to trigger cancelation of the ctx. I _think_ it would do the right thing here given that the blocking commands all take the ctx, but I'm not sure whether that's actually how the CLI wires up exit handling, especially w/ plugins.

pkg/plugins/updater.go
bernerd-stripe: Ah, I guess that doesn't quite capture the subtlety of the spinners continuing to push incremental updates after we've finished the buffering phase?

PR #1542, opened 2026-04-10
feat: add approval_requests public API endpoints to V2 CLI commands

aditigoyal-stripe:
Adds CLI commands for the 5 publicly documented endpoints from approval_requests_service.proto: list, retrieve, submit, cancel, execute. Skips approve/reject which are documented: INTERNAL.

Committed-By-Agent: claude

 ### Reviewers
r? @
cc @stripe/developer-products

 ### Summary


vcheung-stripe (changes requested):
Nice! I think this is failing CI because these two files (resources_gen.go and specs_gen.go) are expected to be autogenerated from the OpenAPI spec. I'd recommend moving these lines into a different file (somewhere in pkg/cmd/resource/ is probably good), and then updating [PostProcessResourceCommands](https://github.com/stripe/stripe-cli/blob/81355d2c64f5bdf5c5696caccfde014454bbc500/pkg/cmd/resource/resource.go#L125) to "patch in" these new commands into the command tree.

aditigoyal-stripe:
> Nice! I think this is failing CI because these two files (resources_gen.go and specs_gen.go) are expected to be autogenerated from the OpenAPI spec. I'd recommend moving these lines into a different file (somewhere in pkg/cmd/resource/ is probably good), and then updating [PostProcessResourceCommands](https://github.com/stripe/stripe-cli/blob/81355d2c64f5bdf5c5696caccfde014454bbc500/pkg/cmd/resource/resource.go#L125) to "patch in" these new commands into the command tree.

Okay fixed it.

pkg/cmd/resource/approval_requests.go
```diff
@@ -0,0 +1,70 @@
...
+			},
+		},
+	},
+	{
+		Name:   "retrieve",
+		Path:   "/v2/core/approval_requests/{id}",
+		Method: "GET",
+	},
+	{
+		Name:   "submit",
```
aditigoyal-stripe, line 31: How does it know what inputs the API needs ?

aditigoyal-stripe: Like the different fields the user needs to supply for each of these apis.

aditigoyal-stripe: no worries it does, my proto was outdated.

vcheung-stripe (approved):

pkg/cmd/resource/approval_requests.go
```diff
@@ -0,0 +1,76 @@
...
+			coreCmd = cmd
+			break
+		}
+	}
+
+	if coreCmd == nil {
+		return nil
+	}
+
+	rApprovalRequestsCmd := NewResourceCmd(coreCmd, "approval_requests")
```
vcheung-stripe, line 69: Would you add this line to make it hidden:
```suggestion
	rApprovalRequestsCmd := NewResourceCmd(coreCmd, "approval_requests")
	rApprovalRequestsCmd.Cmd.Hidden = true
```

vcheung-stripe (changes requested):

vcheung-stripe (approved):

PR #1548, opened 2026-04-15
Fix approval_requests commands not available

vcheung-stripe:
### Summary

Fixes an issue where the `approval_requests` commands were not available. The problem was we were look for the "stripe core" command and adding them under that command, but "stripe core" doesn't exist, it needs to be "stripe v2 core".

Before: `stripe v2 core approval_requests` isn't actually a command so we just get the help output for `stripe v2 core`
```
stripe v2 core approval_requests
Usage:
  stripe v2 core <resource> <operation> [parameters...]

Available Resources:
  account_links         
  account_person_tokens 
  account_persons       
  account_tokens        
  accounts              
  event_destinations    
  events                

Flags:
  -h, --help   help for core

Global Flags:
      --config string         config file (default is $HOME/.config/stripe/config.toml)
      --log-level string      log level (debug, info, trace, warn, error) (default "info")
  -p, --project-name string   the project name to read from for config (default "default")

Use "stripe v2 core [command] --help" for more information about a command.
```

After: `stripe v2 core approval_requests` is a command so we see the help for it.
```
./stripe v2 core approval_requests
Usage:
  stripe v2 core approval_requests <operation> [parameters...]

Available Operations:
  cancel      
  execute     
  list        
  retrieve    
  submit      

Flags:
  -h, --help   help for approval_requests

Global Flags:
      --config string         config file (default is $HOME/.config/stripe/config.toml)
      --log-level string      log level (debug, info, trace, warn, error) (default "info")
  -p, --project-name string   the project name to read from for config (default "default")

Use "stripe v2 core approval_requests [command] --help" for more information about a command.
```


aditigoyal-stripe (approved):

pkg/cmd/resource/approval_requests.go
```diff
@@ -53,9 +53,21 @@ var approvalRequestsSpecs = []OperationSpec{
...
+			v2Cmd = cmd
+			break
+		}
+	}
+
+	if v2Cmd == nil {
+		return nil
+	}
+
+	for _, cmd := range v2Cmd.Commands() {
```
bernerd-stripe: I think this whole block could be replaced with:

```go
coreCmd, _, _ := rootCmd.Find([]string{"v2", "core"})
if coreCmd.Name() != "core" {
  return
}
```

it looks like Find is used for the map feature as well as [billing stubs](https://github.com/stripe/stripe-cli/blob/f3526fc6cc4a96c10f41c8bbee232da92df0c62f/pkg/cmd/root.go#L284) (which seems the most relevant). Note that claude says the billing case doesn't actually handle the case where the command isn't found correctly, because `Find` returns the last match,  so if you did `Find([]string{"v2", "foo"})` it would return 1) the `v2Cmd`, 2) `[]string{"foo"}` (unmatched args), 3) `nil` because it only errors for some legacy flag check behavior.

bernerd-stripe (approved):
Left an idea to simplify but not blocking.

vzhang-stripe (approved):

pkg/cmd/resource/approval_requests.go
vcheung-stripe: Updated with this suggestion. Had claude take a review:
```
  Overall assessment:                                                                                                        
                                                                                                                             
  The fix is correct and clean. A couple of notes:                                                                           
                                                                                                                             
  1. cobra.Command.Find returns (cmd, args, err) but its behavior when a path isn't found is to return the deepest command it
   did find (not necessarily core), which is why the coreCmd.Name() != "core" check is needed — that's the right way to guard
   it.                                                                                                                       
  2. The comment "Fail open if..." is accurate and appropriate given this is a private-preview API.
  3. The test only covers the happy path. A test for the "v2 not present" and "core not present" cases would improve         
  coverage, but given the simplicity of the fail-open logic it's not critical.                                               
```

PR #1550, opened 2026-04-17
Emit Claude Code plugin hint to stderr

tomer-stripe:
## Summary
- When the Stripe CLI detects it's running inside Claude Code (`CLAUDECODE` env var), it emits a `<claude-code-hint />` tag to stderr
- Claude Code strips the hint from output and may prompt users to install the official [Stripe plugin](https://claude.com/plugins/stripe)
- Protocol spec: https://code.claude.com/docs/en/plugin-hints

## Test plan
- [x] Unit tests for hint emission (set/unset env var)
- [x] `go build ./...` compiles
- [x] Verified hint is stripped by Claude Code when running inside it


pkg/cmd/claudecodehint.go
```diff
@@ -0,0 +1,24 @@
...
+// the user to install the Stripe plugin. See
+// https://code.claude.com/docs/en/plugin-hints for the protocol.
+func emitClaudeCodePluginHint() {
+	writeClaudeCodePluginHint(os.Getenv, os.Stderr)
+}
+
+func writeClaudeCodePluginHint(getEnv func(string) string, w io.Writer) {
+	if getEnv("CLAUDECODE") == "" {
+		return
+	}
```
bernerd-stripe, line 22: I think this could be replace with:

```suggestion
	if useragent.DetectAIAgent(getEnv) != "claude_code" {
		return
	}
```

pkg/cmd/claudecodehint.go
```diff
@@ -0,0 +1,24 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+const claudeCodePluginHint = `<claude-code-hint v="1" type="plugin" value="stripe@claude-plugins-official" />`
```
bernerd-stripe, line 9: html on stderr for CLIs is.. a choice, but quite interesting! TIL.

bernerd-stripe (approved):
One small suggestion, otherwise LGTM

PR #1552, opened 2026-04-17
Skip plugin install and upgrade if the requested version is already installed

vcheung-stripe:
### Summary

What it says on the tin. I also updated these commands to output what version was previously installed, if there was one.

Manually tested a bunch of situations:

Screenshot 2026-04-17 at 3 49 26 PM

### DIFF
--- a/pkg/cmd/plugin/install.go
+++ b/pkg/cmd/plugin/install.go
@@ -98,6 +98,8 @@ func (ic *InstallCmd) runInstallCmd(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	prevVersion := plugin.InstalledVersion(ic.cfg, ic.fs)
+
 	ctx := withSIGTERMCancel(cmd.Context(), func() {
 		log.WithFields(log.Fields{
 			"prefix": "cmd.installCmd.runInstallCmd",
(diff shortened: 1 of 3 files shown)


pkg/cmd/plugin/install.go
```diff
@@ -65,55 +66,69 @@ func parseInstallArg(arg string) (string, string) {
 	return plugin, version
 }
 
-func (ic *InstallCmd) installPluginByName(cmd *cobra.Command, arg string) (version string, err error) {
```
vcheung-stripe, line 68: This looks like a big change but most of it is from inlining `installPluginByName`.

pkg/plugins/plugin.go
```diff
@@ -187,6 +187,33 @@ func (p *Plugin) getReleaseForVersion(version string) *Release {
 	return nil
 }
 
+// IsVersionInstalled returns true if the given version of the plugin is already installed on disk.
```
vcheung-stripe, line 190: I think checking the file system has the expected paths and files is enough to say that a version is installed. We could validate the plugin checksum if we want to make this even more precise, but this feels good enough for now since this feature is more of a nicety.

vzhang-stripe (approved):
cool

PR #1555, opened 2026-04-20
Refactor plugin version lookup to use afero and fix shadowing bug in Run

vzhang-stripe:
It's annoying to run the local.build.dev version when developing a plugin, update the CLI version preference to always run the local version

## What                                                                                                                                                                                                 
                                                                                                                                                                                                          
  - Extracts `"local.build.dev"` into a `localDevelopmentVersion` constant                                                                                                                                
  - Introduces `lookUpInstalledVersion()` method that checks for a local dev build first, then falls back to a semver glob — using `afero.Fs` instead of `filepath.Glob` so it's testable                 
  - Adds `isLocalDevelopmentVersion()` predicate and uses it to skip checksum verification for local dev builds in both `verifyChecksum` and `Run`                                                        
  - Fixes a variable shadowing bug in `Run` where `:=` inside the `else` block created a new inner `version`, leaving the outer variable empty and causing `getPluginInstallPath` / `getChecksum` to      
  receive `""` for all non-dev plugin invocations                                                                                                                                                         
                                                                                                                                                                                                          
  ## No regression for production users                                                                                                                                                                   
                                                                  
  Regular users never have a `local.build.dev` directory, so `lookUpInstalledVersion` always falls through to the semver glob. `afero.OsFs` wraps the real filesystem, so behavior is identical to the    
  previous `filepath.Glob` call. The checksum skip only fires for the `local.build.dev` version string.
                                                                                                                                                                                                          
  ## Tests                                                        

  - `TestVerifyChecksumSkipsLocalDevelopmentVersion` — checksum is skipped for local dev builds                                                                                                           
  - `TestLookUpInstalledVersionPrefersLocalDevelopmentVersion` — local dev dir takes priority over an installed release
  - `TestLookUpInstalledVersionFallsBackToInstalledRelease` — falls back to semver glob when no local dev build exists

### DIFF
--- a/pkg/plugins/plugin.go
+++ b/pkg/plugins/plugin.go
@@ -35,6 +35,8 @@ var (
 	PluginsPath string
 )
 
+const localDevelopmentVersion = "local.build.dev"
+
 // CommandInfo describes a plugin subcommand for tree display (e.g. in --map).
 type CommandInfo struct {
 	Name     string        `toml:"Name" json:"name"`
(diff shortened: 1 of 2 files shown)


vcheung-stripe (approved):
Nice!

PR #1558, opened 2026-04-21
Add pkg/cmdutil with FindSubCmd, fix cobra command lookup bugs

bernerd-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

Introduces `pkg/cmdutil` with a `FindSubCmd` helper that correctly wraps `cobra.Command.Find`, and migrates all command-lookup sites in `pkg/cmd` to use it.

**Two silent bugs fixed:**

`cobra.Command.Find` never returns a non-nil error — it signals not-found via a non-empty remaining-args slice and returns the closest matching ancestor rather than nil. Two call sites were checking `err != nil` (always false):

- `addV2BillingStubs` in `root.go` — a missing `v2 billing` namespace would silently attach stubs to the wrong parent
- `AddApprovalRequestsSubCmds` in `approval_requests.go` — same issue; the `coreCmd.Name() != "core"` check was the only real guard

**Seven manual loops replaced:**

`apps.go`, `terminal.go`, `events.go`, `databases.go` (×2), `alias.go`, and `http.go` each had a `for _, cmd := range .Commands()` loop matching on `cmd.Use`. This is fragile — `Use` includes arg hints like `"get <id or path>"`. `FindSubCmd` matches on `cmd.Name()` via cobra's own traversal, which is correct.

**Why `pkg/cmdutil`:**

This helper doesn't belong in `pkg/cmd/resource/` — it has nothing to do with API resources. `pkg/cmdutil` is the right home for generic cobra utilities. The package is intentionally small for now; natural future candidates include `getVisibleCommands` (`map.go`), `wrapText`/`visibleLen` (`templates.go`), and `extractURLParams`/`formatURL` (`operation.go`).


vcheung-stripe (approved):
Nice!

PR #1570, opened 2026-04-30
Read consolidated Stripe-Notice header

kanishk-lgtm:
## Summary

- Read the new consolidated `Stripe-Notice` response header instead of the two separate `Stripe-Api-Version-Upgrade-Notice` and `Stripe-Api-Integration-Path-Upgrade-Notice` headers.
- Print the header value as-is (it's already human-readable, including multi-notice numbered format).

## Test plan

- [x] `TestMakeRequest_StripeNotice` — single notice via new header
- [x] `TestMakeRequest_StripeNoticeMultiNotice` — numbered multi-notice format printed as-is
- [x] `TestMakeRequest_NoUpgradeNotice` — no output when no notice headers present
- [x] `TestMakeRequest_UpgradeNoticeSuppressed` — `SuppressOutput` silences the new header

### DIFF
--- a/pkg/requests/base_test.go
+++ b/pkg/requests/base_test.go
@@ -634,9 +634,10 @@ func captureStderr(t *testing.T, fn func()) string {
 	return string(out)
 }
 
-func TestMakeRequest_VersionUpgradeNotice(t *testing.T) {
+func TestMakeRequest_StripeNotice(t *testing.T) {
+	notice := "You are using an outdated API version (2024-01-01.ember). We recommend upgrading to the latest API version, 2025-01-27.acacia. See https://docs.stripe.com/upgrades"
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Stripe-Api-Version-Upgrade-Notice", "Please upgrade to the latest API version.")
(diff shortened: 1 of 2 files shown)


pkg/requests/base_test.go
```diff
@@ -634,9 +634,10 @@ func captureStderr(t *testing.T, fn func()) string {
 	return string(out)
 }
 
-func TestMakeRequest_VersionUpgradeNotice(t *testing.T) {
+func TestMakeRequest_StripeNotice(t *testing.T) {
+	notice := "You are using an outdated API version (2024-01-01.ember). We recommend upgrading to the latest API version, 2025-01-27.acacia. See https://docs.stripe.com/upgrades"
```
vzhang-stripe, line 638: the message is fixed, would the specific version (2025.01.07.acacia) get outdated?

vzhang-stripe (commented):
just a question about the upgrade message!

pkg/requests/base_test.go
justiny-stripe: I think the content of this notice in the test shouldn't matter, as we're just testing the behavior of Stripe CLI and ensuring that the notice is included.

vzhang-stripe (approved):

PR #1571, opened 2026-05-01
Embed commit SHA in dev builds and skip update check for non-release …

vcheung-stripe:
…versions

`make build` now injects the git SHA via ldflags when the working tree is clean, matching how GoReleaser sets the version for releases. CheckLatestVersion skips the GitHub API check unless the version parses as a valid semver (i.e. a real release), so SHA and "master" builds are unaffected.

Committed-By-Agent: claude

 ### Reviewers
r? @
cc @stripe/developer-products

 ### Summary


vcheung-stripe:
Example:

Screenshot 2026-05-01 at 12 32 00 PM

When the working tree is dirty, we still show master:

Screenshot 2026-05-01 at 12 32 36 PM

vcheung-stripe:
TODO: This breaks `stripe listen` for these clients because the server expects a semver string, so let's revisit this later.

PR #1574, opened 2026-05-04
Refactor symlink protection for .env writes

etsai2-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

- Prevent stripe fixtures from following a symlinked `.env` before reading or writing fixture-generated env values. This closes the arbitrary file overwrite path where attacker-controlled fixture env entries could be written into a symlink target.
- Extract symlink detection and write refusal into a shared `pkg/fsutil` helper so `fixtures` and `samples` use the same implementation instead of duplicating the check.
- Keep existing behavior otherwise unchanged: if `.env` is missing, `fixtures` still no-op.

### DIFF
--- a/pkg/fixtures/fixtures.go
+++ b/pkg/fixtures/fixtures.go
@@ -17,6 +17,7 @@ import (
 	"github.com/spf13/afero"
 	"github.com/tidwall/gjson"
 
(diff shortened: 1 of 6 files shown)


etsai2-stripe:
r?

vzhang-stripe (approved):
lgtm!

PR #1575, opened 2026-05-05
Use plugin metadata endpoint for plugin installs when API key is available

vzhang-stripe:
Adds a new GetPluginMetadata request that fetches plugin-specific manifest and binary URL directly, falling back to the existing GetPluginData flow when the metadata endpoint is unavailable or the user has no API key.

This is the first step to migrate to the new metadata endpoint from jfrog's plugins.toml.

We will test it out on the docs plugin, since the read plugin_metadata method is only giving access to users with the docs gate, pay-server PR here: https://git.corp.stripe.com/stripe-internal/mint/pull/2146204 

Manually tested:

image

if logged in to a non-gated account: 
Screenshot 2026-05-07 at 5 45 10 PM

not logged in: (the new path endpoint is not called at all)
Screenshot 2026-05-07 at 5 47 44 PM

 ### Reviewers
r? @tomer-stripe 
cc @stripe/developer-products

 ### Summary

### DIFF
--- a/pkg/plugins/plugin.go
+++ b/pkg/plugins/plugin.go
@@ -449,6 +449,9 @@ func (p *Plugin) Uninstall(ctx context.Context, config config.IConfig, fs afero.
 
 	err = fs.RemoveAll(pluginDir)
 	if err != nil {
+		if rollbackErr := rollbackInstalledPluginState(config, fs, p.Shortname, previousState); rollbackErr != nil {
+			return fmt.Errorf("failed to remove plugin files for %s: %w; rollback failed: %v", p.Shortname, err, rollbackErr)
+		}
 		return err
 	}
 
--- a/pkg/plugins/plugin.go
+++ b/pkg/plugins/plugin.go
@@ -298,6 +298,7 @@ func (p *Plugin) Install(ctx context.Context, cfg config.IConfig, fs afero.Fs, v
 	apiKey, _ := cfg.GetProfile().GetAPIKey(false)
 	pluginToInstall := p
 	var pluginDownloadURL string
+	downloadURLFromMetadata := false
 
 	if apiKey != "" {
 		log.WithFields(log.Fields{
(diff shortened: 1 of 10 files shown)


pkg/plugins/plugin.go
```diff
@@ -247,21 +296,54 @@ func (p *Plugin) Install(ctx context.Context, cfg config.IConfig, fs afero.Fs, v
 	spinner := ansi.StartNewSpinner(ansi.Faint(fmt.Sprintf("installing '%s' v%s...", p.Shortname, version)), os.Stdout)
 
 	apiKey, _ := cfg.GetProfile().GetAPIKey(false)
+	pluginToInstall := p
```
tomer-stripe, line 299: kind of a nit but how come the copy here?

pkg/plugins/plugin.go
```diff
@@ -247,21 +296,54 @@ func (p *Plugin) Install(ctx context.Context, cfg config.IConfig, fs afero.Fs, v
...
+	var pluginDownloadURL string
+
+	if apiKey != "" {
```
tomer-stripe, line 303: so with this would we copy the manifest from the current unauth'd flow we have? (want to make sure we don't break anything for projects)

vzhang-stripe: Unauthenticated installs still use the old flow, the `p` comes from the existing manifest flow, because `plugin install` refreshes and looks up the plugin before calling `Install`;

pkg/plugins/plugin.go
vzhang-stripe: this is establishing a default plugin object for the install flow. If the metadata endpoint succeeds, that variable gets replaced at pkg/plugins/plugin.go:325; if it doesn’t, the code keeps using the original p for legacy URL lookup, runtime lookup, and download/save

tomer-stripe (approved):

vcheung-stripe (approved):
Did some testing with `generate` and `projects`, with and without auth, confirmed the fallback works.

PR #1579, opened 2026-05-08
Add keychain visibility retry for macOS

vzhang-stripe:
On macOS, keychain writes can be transiently invisible to subsequent reads. This adds a polling retry (up to 1.5s) after keychain set/get operations to handle the race condition where a newly written value isn't immediately readable.

Committed-By-Agent: claude

 ### Reviewers
r? @tomer-stripe 
cc @stripe/developer-products

### Motivation
https://stripe.slack.com/archives/C0A9VNT3PFU/p1778196782726039

### DIFF
--- a/pkg/plugins/core_cli_helper.go
+++ b/pkg/plugins/core_cli_helper.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"runtime"
+	"sync"
 	"time"
 
 	"github.com/99designs/keyring"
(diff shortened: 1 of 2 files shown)


pkg/plugins/core_cli_helper.go
```diff
@@ -154,6 +156,52 @@ type coreCLIHelper struct {
...
+var (
+	keychainVisibilityRetryInterval = 100 * time.Millisecond
+	keychainVisibilityRetryTimeout  = 1500 * time.Millisecond
+	keychainVisibilityRetryEnabled  = runtime.GOOS == "darwin"
+	keychainVisibilityNow           = time.Now
+	keychainVisibilitySleep         = time.Sleep
+)
+
+func readKeychainPassword(key string) (string, bool, error) {
+	item, err := config.KeyRing.Get(key)
```
vcheung-stripe, line 174: One thing I want to confirm here is that this won't prompt the user for their password on every poll. You can reset your "Always allow" setting in your own keychain.

vzhang-stripe: tested manually!

vcheung-stripe (approved):
Tested this locally, my steps:
1. Run a plugin command that uses the keychain: `go run cmd/stripe/main.go projects init`
2. Observe that the keychain pops up
3. Run the command again, this time click "Always allow" in the keychain pop up
4. Observe that it doesn't pop up

Issue #558, opened 2021-01-06
Add invoice.upcoming to the supported list of events


iamtekeste:
### Problem
Unable to trigger and listen to the `invoice.upcoming` event from the CLI

### Feature
Adding this event to the supported list of events would unblock me.

Thank you!

pepin-stripe:
Hi @iamtekeste, I've filed a task and will into adding support for this event!

iamtekeste:
Thank you!

wb-joona:
Would love to see this added 👍

uptownhr:
I want to call out that this option is available in the dropdown in the the Stripe UI 
![image](https://user-images.githubusercontent.com/1021110/193114402-740542a4-809a-4dcd-82d4-fa5941b9db2e.png)

danabrey:
Is this still being looked into? It's a given example for triggering in the Listen To Stripe Events docs.

penguinsource:
Bump!

willricketts:
Bumping this again as this would be incredibly helpful.

Swish24:
Would be really useful to have this added!

mjonjones-veygo:
Is there an update on this? Bumping to keep this unsupported feature alive.

ateeb-studio93:
Is there an update on this? Bumping to keep this unsupported feature alive.

de-raaf-media:
@ob-stripe maybe?

PR #1582, opened 2026-05-10
Add invoice.upcoming trigger support via test clock fixture

SumitKumar-17:
### Reviewers
r? @
cc @stripe/developer-products

  ### Summary
  Fixes the issue where `stripe trigger invoice.upcoming` failed
  with "event is not supported", even though `stripe listen --events invoice.upcoming`
  already worked.

  **Root cause:** `invoice.upcoming` is a system-generated event fired by
  Stripe's billing engine 3 days before a subscription renews. 

  **Fix:** Add `pkg/fixtures/triggers/invoice.upcoming.json` using Stripe's
  test clock API — the approach Stripe's own documentation recommends for
  testing time-based billing events.

  The fixture:
  1. Creates a test clock frozen at Jan 1, 2024
  2. Creates a customer tied to that test clock (with `pm_card_visa` as default payment method)
  3. Creates a product + monthly recurring price
  4. Creates a subscription (first invoice paid immediately at Jan 1)
  5. Advances the clock to Jan 29, 2024 — 3 days before the Feb 1 renewal —
     which causes Stripe's billing engine to emit `invoice.upcoming`

  Using static timestamps keeps the fixture deterministic. 

  **Verified end-to-end against live test keys:**
 
image

  Stripe returned `evt_...` with `"type": "invoice.upcoming"`,
  `"billing_reason": "upcoming"`, and a draft invoice (no ID) attached —
  exactly the payload a real `invoice.upcoming` webhook carries.

  ### Changes
  - `pkg/fixtures/triggers/invoice.upcoming.json` — new fixture file


SumitKumar-17:
If Approved adds the invoice.upcoming triggers also fixes #558

PR #1583, opened 2026-05-10
Fix --api-key flag silently ignored when STRIPE_API_KEY env var is set

### LINKED ISSUE #1581: `--api-key` flag is silently overridden by `STRIPE_API_KEY` env var
## Summary

When the `STRIPE_API_KEY` environment variable is set, the `--api-key` CLI flag is silently ignored. No warning or info is printed. This caused production-intended commands to operate on a sandbox account.

## Reproduction

```sh
export STRIPE_API_KEY=sk_test_AAA        # sandbox
stripe --api-key rk_live_BBB webhook_endpoints list
# expected: results from rk_live_BBB (production)
# actual:   results from sk_test_AAA (sandbox), no warning
```

Verified that:
- `unset STRIPE_API_KEY` then re-running the same command → `--api-key` is used as expected
- `curl -u "rk_live_BBB:" https://api.stripe.com/v1/webhook_endpoints` → correctly hits the production account

So the live key itself is valid; only the CLI's resolution order ignores the flag.

## Expected behavior

Standard CLI precedence is **flag > env var > config file**. Either:

1. `--api-key` should win over `STRIPE_API_KEY`, OR
2. The CLI should print a warning when both are set to different values, e.g. `warn: --api-key is being overridden by STRIPE_API_KEY`

Currently neither happens — `--api-key` is silently ignored.

## Why this matters

The Stripe CLI docs state that you can make live-mode requests on a per-command basis by attaching the `--api-key` flag (https://docs.stripe.com/cli/api_keys). Following that guidance while having `STRIPE_API_KEY` set in the shell silently routes the command to the wrong account. Users see "success" output but resources (webhook endpoints, prices, etc.) end up on a sandbox while production dashboards/logs are being checked, or vice versa.

This was reproducible 100% on our setup — flag position before/after the subcommand made no difference.

## Environment

- stripe-cli: 1.40.9
- macOS (zsh)

SumitKumar-17:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary
When `STRIPE_API_KEY` was set in the environment, the `--api-key` CLI flag was silently ignored. As per CLI precedence (flags should  beat env vars) and could cause commands to operate on the wrong account with no warning.

Fixes #1581 

I just changed the order in the code,no code added 😅

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -219,6 +219,15 @@ func (p *Profile) GetAccountID() (string, error) {
 
 // GetAPIKey will return the existing key for the given profile
 func (p *Profile) GetAPIKey(livemode bool) (string, error) {
+	if p.APIKey != "" {
+		err := validators.APIKey(p.APIKey)
+		if err != nil {
+			return "", err
+		}
+
+		return p.APIKey, nil
+	}
+
 	envKey := os.Getenv("STRIPE_API_KEY")
 	if envKey != "" {
 		err := validators.APIKey(envKey)
@@ -227,15 +236,6 @@ func (p *Profile) GetAPIKey(livemode bool) (string, error) {
 		}
 
 		return envKey, nil
-	}
-
-	if p.APIKey != "" {
-		err := validators.APIKey(p.APIKey)
-		if err != nil {
-			return "", err
-		}
-
-		return p.APIKey, nil
 	}
 
 	var key string


tomer-stripe:
Appreciate the PR but going to close this for now. This could be a breaking change for users and I want to be careful here

SumitKumar-17:
@tomer-stripe Sure

PR #1586, opened 2026-05-11
Protect writes from symlink targets

etsai2-stripe:
### Reviewers
r? @vcheung-stripe 
cc @stripe/developer-products

 ### Summary

This PR hardens remaining local file writes against write-through-symlink attacks using the same pattern as the earlier `samples` / `fixtures` fix, without duplicating one-off checks.

  - add a shared symlink guard in `pkg/fsutil` for both afero-backed writes and direct OS writes
  - apply it to PDF downloads in `pkg/requests/base.go`
  - apply it to plugin binary installs and `plugins.toml` writes in `pkg/plugins`
  - apply it to CLI config writes so both `Config.WriteConfigField` and `Profile.WriteConfigField` reject symlink targets consistently
  - move existing `.env` protections in `samples` / `fixtures` onto the shared helper

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -482,7 +482,7 @@ func writeConfig(runtimeViper *viper.Viper) error {
 	configType := strings.TrimPrefix(filepath.Ext(profilesFile), ".")
 	runtimeViper.SetConfigType(configType)
 
-	if err := fsutil.RefuseWriteThroughSymlinkOS(profilesFile, filepath.Base(profilesFile)); err != nil {
+	if err := fsutil.RefuseWriteThroughSymlinkOS(profilesFile, filepath.Dir(filepath.Dir(profilesFile)), filepath.Base(profilesFile)); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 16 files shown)


vzhang-stripe (commented):
1. _pkg/fsutil/symlink.go:43-56_ only checks the final path with `os.Lstat`. If a parent directory is a symlink and the destination file does not exist yet, the guard returns nil on ENOENT, and the later write still follows the symlink. That means the new protections in _pkg/config/config.go:485_, _pkg/plugins/plugin.go:382_, and _pkg/plugins/utilities.go:141_ are bypassable.


2. The plugin install flow is still exposed for plugins with runtime dependencies. _pkg/plugins/plugin.go:267-270_ calls `InstallNodeRuntime`, and _pkg/plugins/runtime.go:360_ plus _pkg/plugins/runtime.go:436_ still write extracted files with raw `OpenFile(...O_CREATE|O_TRUNC...)` and no symlink protection. So this PR does not actually close the whole write-through-symlink class it claims to fix.

vzhang-stripe (approved):
:shipit:

PR #1590, opened 2026-05-14
feat(fixtures): accept multiple fixture files (closes #910)

### LINKED ISSUE #910: Stripe fixtures support for multiple file inputs
### Feature
This is something that would be nice to have, I would like to request support for multiple file input support for  the`stripe fixtures` command.

Currently you could always do:
`stripe fixtures ./subscription.created.json && stripe fixtures ./teardown.json `

However it would be neat if the `fixtures` command allowed unlimited file input args:

`stripe fixtures ./subscription.created.json ./teardown.json`

0xdeadd:
Closes #910.

The `fixtures` command previously required exactly one file argument. This extends it to accept one or more files, executed sequentially in the order given.

```
stripe fixtures ./customer.json ./subscription.json ./teardown.json
```

## Behavior

- `UpdateEnv` is called after each fixture, so later files can reference vars produced by earlier ones via `${.env:VAR}`.
- Execution aborts on the first error rather than continuing, to preserve the established single-file behavior and surface failures immediately. Users who want try-each semantics can chain at the shell level.
- The existing single-file invocation continues to work unchanged.

## Implementation

- Added a `MinimumNArgs` validator to `pkg/validators/cmds.go`, mirroring the existing `MaximumNArgs`, since the package only had `Exact` and `Maximum` variants. Includes tests.
- Switched the `fixtures` command from `validators.ExactArgs(1)` to `validators.MinimumNArgs(1)` and wrapped the runner body in a per-file loop.
- Updated the command's `Use` and `Long` to reflect the new shape.

## Test

```
$ go test ./pkg/validators/... ./pkg/fixtures/...
ok  	github.com/stripe/stripe-cli/pkg/validators	0.204s
ok  	github.com/stripe/stripe-cli/pkg/fixtures	0.283s
```

`gofmt` and `go vet` clean.


0xdeadd:
Gentle nudge here. This implements #910 (accepting multiple fixture files). CLA is signed and checks pass, and I've just updated the branch against base so it's no longer behind. Glad to make any changes a reviewer wants. Thanks for taking a look!

PR #1595, opened 2026-05-20
Distribute Stripe CLI via npm (`npx @stripe/cli`)

johno-stripe:
### Reviewers
r? @
cc @stripe/developer-products

### Summary

Adds npm as an install channel for the Stripe CLI. After this ships, developers with Node >= 18 can run:

```
npx @stripe/cli login
npm install -g @stripe/cli
```

This doesn't replace existing install paths (Homebrew, apt, Scoop, Docker) — it adds a fast path for the majority of the CLI's audience who already have Node installed.

### Motivation

The Stripe CLI's primary audience is web developers integrating a payments API. Most already have Node.js. Today, installing the CLI on Windows requires Scoop (a package manager most developers haven't heard of), and every platform requires a separate tool. `npx @stripe/cli` is a universal one-liner.

### What's included

- **6 npm packages** under the `@stripe` scope: a wrapper (`@stripe/cli`) and 5 platform-specific binary packages (darwin-arm64, darwin-x64, linux-x64, linux-arm64, win32-x64)
- **`publish-npm` job** in the release workflow, gated on all three platform builds completing
- **Install-test coverage**: 3 new jobs (npm, npm-no-optional, npx) across platforms with PagerDuty alerting
- **Fallback for `--no-optional`**: the wrapper's postinstall downloads directly from GitHub Releases and verifies SHA256 checksums
- **Installer telemetry**: expands existing telemetry to include a `install_method` key, shim checks whether it's running via `npx` and reports accordingly. Also tries to check for installs by `brew`, `apt-get`, or `scoop`.

No changes to Go code. No changes to goreleaser configs. The existing release artifacts are reused as-is.  No need for an `NPM_TOKEN` variable in the repo, as we're using the Trusted Publisher workflow.

### Test plan

- [x] `node --check` passes on all JS files
- [x] `node npm/scripts/stamp-version.js 1.41.1` stamps all package.json files correctly
- [x] `VERSION=1.41.1 GITHUB_TOKEN=... node npm/scripts/fetch-binaries.js` downloads, verifies, and extracts all 5 platform binaries
- [x] Shim resolves platform package and launches binary (`node npm/wrapper/bin/shim.js --version` via local symlink)
- [x] Postinstall fallback exercises download → checksum → extraction path
- [x] Post-publish: `npm install -g @stripe/cli && stripe --version` on all 3 platforms (covered by new install-test jobs)
- [x] Post-publish: `npx --yes @stripe/cli --version` exits successfully
- [x] Post-publish: `npm install -g @stripe/cli --no-optional && stripe --version` exercises fallback path

<details>

<summary>
### How it works
</summary>

For anyone unfamiliar with the npm mechanisms at play:

### Platform selection via `optionalDependencies` + `os`/`cpu`

npm's `package.json` supports two fields that declare platform requirements:

```json
{ "os": ["darwin"], "cpu": ["arm64"] }
```

When a package declares these, npm will **skip installing it** on non-matching hosts. The wrapper package (`@stripe/cli`) lists all 5 platform packages as `optionalDependencies`:

```json
"optionalDependencies": {
  "@stripe/cli-darwin-arm64": "1.41.1",
  "@stripe/cli-darwin-x64": "1.41.1",
  ...
}
```

`optionalDependencies` means "install these if you can, don't fail if you can't." Combined with the `os`/`cpu` guards, npm installs **only** the one matching the user's machine (typically 15-50 MB). The other 4 are silently skipped.

#### The `bin` field and the shim

The wrapper declares `"bin": { "stripe": "bin/shim.js" }`. When installed globally (`npm install -g`), npm symlinks `stripe` into the user's PATH pointing at `shim.js`. The shim is 23 lines: it detects the current platform, finds the installed binary package via `require.resolve()`, and spawns it with all arguments forwarded and `stdio: 'inherit'` (transparent passthrough).

#### The `postinstall` fallback

If npm was run with `--no-optional` (or a corporate policy strips optional deps), no platform package is installed. The wrapper's `"scripts": { "postinstall": "node scripts/postinstall.js" }` detects this, downloads the correct archive from GitHub Releases, verifies the SHA256 checksum against goreleaser's published checksum files, and extracts the binary to a `vendor/bin/` directory. The shim falls back to this path when `require.resolve()` fails.

#### Why platform packages have a `bin` field too

Each platform package (e.g. `@stripe/cli-darwin-arm64`) also declares `"bin": { "stripe": "bin/stripe" }`. This is a safety net: if someone installs a platform package directly, the binary still gets symlinked correctly. It also allows npm to set the executable bit during installation on systems that require it.
```

</details>

### DIFF
--- a/pkg/useragent/useragent.go
+++ b/pkg/useragent/useragent.go
@@ -5,7 +5,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/stripe/stripe-cli/pkg/version"
 )
@@ -24,6 +26,34 @@ func GetEncodedStripeUserAgent() string {
 // the `User-Agent` HTTP header.
 func GetEncodedUserAgent() string {
 	return encodedUserAgent
+}
+
+// DetectInstallMethod detects how the CLI was installed.
+// Returns one of: "npm", "npx", "homebrew", "scoop", "apt", or "unknown".
+func DetectInstallMethod(
+	getEnv func(string) string,
+	getExecutable func() (string, error),
+	statFile func(string) error,
+) string {
+	if method := getEnv("STRIPE_INSTALL_METHOD"); method != "" {
+		return method
+	}
+
+	if exe, err := getExecutable(); err == nil {
+		exeLower := strings.ToLower(filepath.ToSlash(exe))
+		if strings.Contains(exeLower, "/cellar/") || strings.Contains(exeLower, "/homebrew/") {
+			return "homebrew"
+		}
+		if strings.Contains(exeLower, "/scoop/apps/") {
+			return "scoop"
+		}
+	}
+
+	if err := statFile("/var/lib/dpkg/info/stripe.list"); err == nil {
+		return "apt"
+	}
+
+	return "unknown"
 }
 
 // DetectAIAgent detects if the CLI was invoked by a coding agent, based on well-known env vars.
--- a/pkg/useragent/useragent.go
+++ b/pkg/useragent/useragent.go
@@ -29,7 +29,7 @@ func GetEncodedUserAgent() string {
 }
 
 // DetectInstallMethod detects how the CLI was installed.
-// Returns one of: "npm", "npx", "homebrew", "scoop", "apt", or "unknown".
+// Returns one of: "npm_global", "npm_run", "npx", "homebrew", "scoop", "apt", or "unknown".
 func DetectInstallMethod(
 	getEnv func(string) string,
 	getExecutable func() (string, error),
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+export GH_HOST=github.com
--- a/.github/workflows/install-test.yml
+++ b/.github/workflows/install-test.yml
@@ -102,3 +102,51 @@ jobs:
       - run: sh scripts/install-test.sh docker
         env:
           PAGERDUTY_INTEGRATION_KEY: ${{ secrets.PAGERDUTY_INTEGRATION_KEY }}
+
+  npm:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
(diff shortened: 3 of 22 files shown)

### CODE REVIEW

**Comment by varughese** [comment]
When I try this on my local machine I get a

```
sh: stripe: command not found
```

CleanShot 2026-06-01 at 22 03 01@2x

Which persists after I reload.

However, `npm install -g @stripe/cli` works:

CleanShot 2026-06-01 at 22 03 41@2x

`which stripe`


`/Users/matv/.nvm/versions/node/v22.14.0/bin/stripe`

So I think to make "installing stripe cli as a local dev dep using NPM" work we need to fix it so 

```
npm install @stripe/cli -D
```

and `npx @stripe/cli whoami` in that project would work. And maybe we could somehow make the profile deterministic based on the path

**Comment by johno-stripe** [comment]
Papertrailing what we discussed offline, @varughese -- that's expected behavior given that `npx` doesn't modify the PATH, but I think we can try adding something in a later release which allows for both installing & running a command in a single line!

**Comment by johno-stripe** [comment]
FYI @vcheung-stripe & @tomer-stripe , I added the `npx` telemetry!  It's got some tests and I also did a quick local validation.  Given that it's changing the actual binary and our publication pipeline pulls our latest release binary, I think we'd need to publish a new version before the telemetry started working properly.

CleanShot 2026-06-03 at 12 25 44

**Review by vcheung-stripe — APPROVED** [review]

File: pkg/useragent/useragent.go [hunk 1]
```diff
@@ -26,6 +28,34 @@ func GetEncodedUserAgent() string {
...
+	}
+
+	if exe, err := getExecutable(); err == nil {
+		exeLower := strings.ToLower(filepath.ToSlash(exe))
+		if strings.Contains(exeLower, "/cellar/") || strings.Contains(exeLower, "/homebrew/") {
+			return "homebrew"
+		}
+		if strings.Contains(exeLower, "/scoop/apps/") {
+			return "scoop"
+		}
```
vcheung-stripe, line 49: We added support for winget recently! Once I thoroughly test that, I'll update this to look for the winget paths

PR #1611, opened 2026-05-26
Add canary tests for v2 API endpoints and triggers

vcheung-stripe:
### Summary

v2 APIs have their own request shape, so we should cover them in the canary. This will ensure changes to auth don't break these flows unexpectedly.

Tested by triggering the workflow manually and seeing the requests succeed: https://github.com/stripe/stripe-cli/actions/runs/26467782853


vzhang-stripe (approved):
nice

PR #1643, opened 2026-06-08
Save UAT and account information in profile

vcheung-stripe:
### Summary

This saves the UAT and account information to the profile during login, only if this data is returned by the server during `stripe login` (requires a flag to be enabled).


vcheung-stripe:
Will be superseded by https://github.com/stripe/stripe-cli/pull/1662

PR #1644, opened 2026-06-08
Add terminal environment telemetry

tomer-stripe:
## Summary
- add telemetry fields for whether the CLI runs inside tmux or GNU Screen
- add terminal program detection using TERM_PROGRAM and common terminal-specific fallback env vars
- include the new metadata on CLI telemetry events and API request telemetry

## Testing
- go test ./pkg/useragent
- go test ./pkg/stripe
- git -c core.fsmonitor=false diff --check
- env -u CODEX_SANDBOX -u CODEX_THREAD_ID -u CODEX_SANDBOX_NETWORK_DISABLED -u CODEX_CI go test ./...

Note: plain `go test ./...` fails in this Codex shell because `pkg/stripeauth` has an existing user-agent regex expectation that does not allow the environment-added `AIAgent/codex_cli` suffix.

### DIFF
--- a/pkg/useragent/useragent.go
+++ b/pkg/useragent/useragent.go
@@ -56,6 +56,39 @@ func DetectInstallMethod(
 	return "unknown"
 }
 
+// DetectInTmux detects whether the CLI was invoked from within a tmux session.
+func DetectInTmux(getEnv func(string) string) bool {
+	return getEnv("TMUX") != ""
+}
+
+// DetectInScreen detects whether the CLI was invoked from within a GNU Screen session.
+func DetectInScreen(getEnv func(string) string) bool {
+	return getEnv("STY") != ""
+}
+
+// DetectTerminalProgram detects the terminal program the CLI was invoked from, when available.
+func DetectTerminalProgram(getEnv func(string) string) string {
+	if program := getEnv("TERM_PROGRAM"); program != "" {
+		return program
+	}
+	if getEnv("WT_SESSION") != "" {
+		return "windows_terminal"
+	}
+	if getEnv("KITTY_WINDOW_ID") != "" {
+		return "kitty"
+	}
+	if getEnv("ALACRITTY_WINDOW_ID") != "" || getEnv("ALACRITTY_LOG") != "" {
+		return "alacritty"
+	}
+	if getEnv("WEZTERM_EXECUTABLE") != "" || getEnv("WEZTERM_PANE") != "" {
+		return "wezterm"
+	}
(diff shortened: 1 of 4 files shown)


pkg/useragent/useragent.go
```diff
@@ -56,6 +56,39 @@ func DetectInstallMethod(
...
+// DetectInScreen detects whether the CLI was invoked from within a GNU Screen session.
+func DetectInScreen(getEnv func(string) string) bool {
+	return getEnv("STY") != ""
+}
+
+// DetectTerminalProgram detects the terminal program the CLI was invoked from, when available.
+func DetectTerminalProgram(getEnv func(string) string) string {
+	if program := getEnv("TERM_PROGRAM"); program != "" {
+		return program
+	}
```
martingordon: Can we move this to the bottom? Most terminals I checked when running in tmux will have this set to `tmux`.

martingordon: Also, I was able to detect iTerm2 with LC_TERMINAL, can we add that as well?

Image

martingordon (commented):
Looks great, thanks for the quick turnaround! Just a minor comment for better detection in tmux and detecting iTerm2.

pkg/useragent/useragent.go
martingordon: Also for Warp, we can use one of these (maybe `WARP_CLIENT_VERSION`?)

Image

tomer-stripe: ty! will make these changes. i feel like this is going to be somewhat messy and we'll always need to be slightly modifying this 😆

tomer-stripe:
Updated in 6f191b2:
- moved `TERM_PROGRAM` fallback after the more specific terminal env vars so `TERM_PROGRAM=tmux` does not mask the host terminal when another signal is present
- added `LC_TERMINAL` detection for iTerm2
- added `WARP_CLIENT_VERSION` detection for Warp
- added tests for the new cases and precedence over `TERM_PROGRAM=tmux`

Verification:
- `go test ./pkg/useragent`
- `go test ./pkg/stripe`
- `git -c core.fsmonitor=false diff --check`
- `env -u CODEX_SANDBOX -u CODEX_THREAD_ID -u CODEX_SANDBOX_NETWORK_DISABLED -u CODEX_CI go test ./pkg/cmd`

I also retried the previously failing `pkg/cmd` package; the earlier full-suite run hit a transient TLS handshake timeout fetching the plugin manifest, and the package passed on retry.

vzhang-stripe (approved):

PR #1661, opened 2026-06-11
Replace credential store: zalando/go-keyring with timeout + file fallback

tomer-stripe:
## Summary

- Replaces unmaintained `99designs/keyring` (last release 2022) with `zalando/go-keyring`
- Eliminates macOS Keychain password prompts when running `--live` commands
- Adds file-based fallback store for headless Linux, containers, and D-Bus-unavailable environments
- Adds debug logging for credential store access (`--log-level=debug`)

## Problem

The old `99designs/keyring` library set per-application ACLs on keychain items via the legacy `SecAccessCreate` API. Every time the CLI binary changed (upgrades, dev rebuilds), macOS treated it as an untrusted application and prompted for the keychain password — often twice per invocation. "Always Allow" didn't persist across binary changes.

## Solution

`zalando/go-keyring` shells out to `/usr/bin/security` on macOS. Items stored this way use `/usr/bin/security` as the trusted application in their ACL, so reads never prompt regardless of which Go binary calls them.

A single `zalandoStore` wraps every keyring operation in a goroutine with a 3-second timeout — the same approach used by the GitHub CLI — so a hung D-Bus daemon on Linux times out cleanly instead of blocking forever.

When the OS keyring is unavailable (headless Linux, containers, D-Bus timeout), a `fallbackStore` transparently retries against a `fileStore` that writes a 0600-permissioned `credentials.json` in the Stripe config directory — mirroring the GitHub CLI pattern.

**Platform backends:**
- **macOS**: `zalando/go-keyring` → `/usr/bin/security` (zero prompts)
- **Linux**: `zalando/go-keyring` → D-Bus Secret Service, with file fallback on timeout/unavailable
- **Windows**: `zalando/go-keyring` → Windows Credential Manager

Uses a new service name (`stripe-cli` vs `StripeCLI`) to avoid touching legacy ACL-protected items. Users do a one-time `stripe login --live` after upgrading.

## Upgrade UX

- **Test mode commands**: No change, work immediately
- **Live mode commands**: One-time re-authentication message:
  ```
  your live mode API key needs to be re-configured. Run `stripe login --live` to re-authenticate
  ```
- After re-login: zero prompts forever, including across future upgrades

## Test plan

- [x] All CI checks pass (macOS, Ubuntu, Windows)
- [x] Lint clean (golangci-lint v2)
- [x] Unit tests for `zalandoStore` timeout behavior (injectable backend functions)
- [x] Unit tests for `fallbackStore`: primary-success, primary-failure-fallback, remove-both paths
- [x] Unit tests for `fileStore`: CRUD, missing-file, corrupt-JSON
- [x] Manual test: store key → read → rebuild binary → read again = zero prompts on macOS
- [x] Cross-compilation: `GOOS=linux`, `GOOS=windows` both build clean


pkg/config/keyring_darwin.go
```diff
@@ -0,0 +1,104 @@
...
+		return ErrKeyNotFound
+	}
+	return err
+}
+
+func (k *darwinKeychain) Keys() ([]string, error) {
+	out, err := exec.Command("/usr/bin/security", "dump-keychain").Output()
+	if err != nil {
+		return []string{}, nil
+	}
```
tomer-stripe: hm i thought we didn't need this with the zalando package. looking into it

vzhang-stripe (approved):

vcheung-stripe (changes requested):
There are some issues here I'm working out, requesting changes so it doesn't get merged.

pkg/config/profile.go
```diff
@@ -343,7 +277,7 @@ func (p *Profile) GetAPIKey(livemode bool) (string, error) {
 		p.redactAllLivemodeValues()
 		key, err = p.retrieveLivemodeValue(LiveModeAPIKeyName)
 		if err != nil {
-			return "", err
+			return "", errors.New("your live mode API key needs to be re-configured. Run `stripe login --live` to re-authenticate")
```
vcheung-stripe: `--live` is not a valid flag for `stripe login`

pkg/keyring/keyring.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (s *fallbackStore) Get(key string) ([]byte, error) {
+	logger := log.WithFields(log.Fields{"prefix": "keyring", "key": key})
+	logger.Debug("reading from credential store")
+	data, err := s.primary.Get(key)
+	if err == nil {
+		return data, nil
+	}
+	logger.WithError(err).Debug("credential store unavailable, reading from fallback file")
```
tomer-stripe, line 129: i forget, `logger.WithError` doesn't exit, right?

pkg/keyring/keyring.go
```diff
@@ -0,0 +1,247 @@
...
+		del:     zkr.Delete,
+	}
+}
+
+func (s *zalandoStore) Get(key string) ([]byte, error) {
+	type result struct {
+		val string
+		err error
+	}
+	ch := make(chan result, 1)
```
tomer-stripe, line 68: ooc how come some of these take place in channels?

pkg/config/keyring.go
```diff
@@ -0,0 +1,247 @@
...
+		logger.WithError(primaryErr).Debug("credential store unavailable during remove")
+	}
+	return nil
+}
+
+// fileStore persists credentials as a JSON object in a 0600-permissioned file
+// inside the Stripe config directory.
+type fileStore struct {
+	path string
+}
```
tomer-stripe: is this separate from our config file? should it be the config file?

pkg/config/keyring.go
```diff
@@ -0,0 +1,247 @@
...
+
+func newFileStore() *fileStore {
+	dir := getConfigFolder(os.Getenv("XDG_CONFIG_HOME"))
+	return &fileStore{path: filepath.Join(dir, "credentials.json")}
```
tomer-stripe: ah yes different file. is there a kind of intermediate migration we need since the configs are changing?

pkg/keyring/keyring.go
```diff
@@ -0,0 +1,247 @@
...
+
+func newSecureStore() SecureStore {
+	return &fallbackStore{
+		primary:  newZalandoStore(KeyManagementService),
+		fallback: newFileStore(),
+	}
+}
+
+// MemoryStore is an in-memory SecureStore for use in tests.
+type MemoryStore struct {
```
tomer-stripe: this is a nit: do we use this outside of the config package? wondering if we should keep it in the keyring test file so test things are colocated. not a strong opinion though

tomer-stripe: actually should we user afero memfs?

pkg/config/profile.go
```diff
@@ -343,7 +277,7 @@ func (p *Profile) GetAPIKey(livemode bool) (string, error) {
 		p.redactAllLivemodeValues()
 		key, err = p.retrieveLivemodeValue(LiveModeAPIKeyName)
 		if err != nil {
-			return "", err
+			return "", errors.New("your live mode API key needs to be re-configured. Run `stripe login` to re-authenticate")
```
tomer-stripe, line 281: so we will only show this to a user if they do live operations right? does that mean the credentials file above moves only the live keys or test too?

pkg/plugins/core_cli_helper.go
```diff
@@ -71,7 +70,7 @@ func (c *CoreCLIHelperClient) KeychainDeletePassword(key string) (bool, error) {
 }
 
 func (c *CoreCLIHelperClient) KeychainFindCredentials() ([]string, error) {
-	resp, err := c.client.KeychainFindCredentials(context.Background(), &proto.KeychainFindCredentialsRequest{})
+	resp, err := c.client.KeychainFindCredentials(context.Background(), &proto.KeychainFindCredentialsRequest{}) //nolint:staticcheck
```
tomer-stripe, line 75: how come the nolint here and in the file?

tomer-stripe (commented):
can you also make it so if a user logs in and we don't use a secure store we show a warning as part of the login flow?

pkg/keyring/keyring.go
vcheung-stripe: Confirmed it doesn't exit:
```
$ ./stripe customers list --live --log-level debug
[Tue, 23 Jun 2026 21:49:40 UTC] DEBUG config.Config.InitConfig: Using profiles file path=/home/owner/.config/stripe/config.toml
[Tue, 23 Jun 2026 21:49:40 UTC] DEBUG keyring: reading from credential store key=default.live_mode_api_key
[Tue, 23 Jun 2026 21:49:40 UTC] DEBUG keyring: credential store unavailable, reading from fallback file error=The name org.freedesktop.secrets was not provided by any .service files key=default.live_mode_api_key
{
  "object": "list",
  "data": [
...
```

pkg/keyring/keyring.go
vcheung-stripe: This is so we can quit if it takes longer than 3 seconds to read from the store. It's what the github CLI does but is probably a premature optimization, we don't have to do this but I figured why not

pkg/config/keyring.go
vcheung-stripe: No migration needed, it's just the live key in here. Putting it in a file is no different from putting it in a different keychain collection insofar as backwards compatibility is concerned.

pkg/keyring/keyring.go
vcheung-stripe: I actually want to move the all the keyring code into its own separate package, in fact I'll do that in addition to this.

pkg/config/profile.go
vcheung-stripe: Yes this error is only shown when doing live operations, not test or sandbox. No change to test keys, they stay in the config.toml

pkg/plugins/core_cli_helper.go
vcheung-stripe: It's because I marked `KeychainFindCredentials` as deprecated since zalando doesn't support listing keys, and the linter flags this line. I confirmed only `projects` is using `KeychainFindCredentials` but they are using it as a probe into the core CLI, not reading the return value.

pkg/config/keyring.go
vcheung-stripe: I feel like it should be its own file since it's the "backend" of the keyring implementation, and so is conceptually different from config.

vcheung-stripe:
Tested these scenarios:
- `stripe login`
- `stripe login --non-interactive`
- `stripe login` then `stripe login` to another account
- `stripe login` then `stripe login` to a sandbox, confirmed live key for the first login gets removed
- `stripe logout`
- `stripe customers list --live`
- `stripe customers list --live` while not logged in
- `stripe customers list` (to confirm test mode still works)

Confirmed on these platforms:
- macOS: uses keychain, no prompting
- Windows: uses wincred, no prompting
- Ubuntu desktop: secret service, no prompting
- WSL: file fallback with 0600 permissions and warning during login
- Ubuntu server: file fallback with 0600 permissions and warning during login

Here's what the warning looks like:
```
./stripe login         
Your pairing code is: cajole-plush-skill-enough
To authenticate with Stripe, please go to: https://dashboard.stripe.com/stripecli/confirm_auth?t=AlhsdIJXtQrWtBz5TMln94BBzzbdzGai
> Done! The Stripe CLI is configured for Tomer's Chicken n' Waffles (v1 Workbench) with account id acct_1NRKwLLJDmqA11cn
Please note: this key will expire after 90 days, at which point you'll need to re-authenticate.
Warning: the system keyring is unavailable. Your credentials have been stored unencrypted in /home/owner/.config/stripe/credentials.json
```

With color:
Screenshot 2026-06-24 at 9 49 17 AM

vzhang-stripe (commented):
P2: Existing WSL/Linux users can lose access to already-stored live credentials because `pkg/keyring/keyring.go:29` no longer reads/migrates the old 99designs file backend

pkg/config/keyring.go
tomer-stripe: ah testmode will live in the config file and live will live in credentials.json?

vzhang-stripe (approved):

vcheung-stripe (approved):

PR #1663, opened 2026-06-11
feat(cmd): add subcommand `docs` to browse docs.stripe.com

joelzwarrington:
Introduces `stripe docs` as a built-in subcommand. It was prototyped as a standalone plugin and is bundled into the CLI directly without requiring installation.

https://github.com/user-attachments/assets/48b027a8-2194-4d42-9163-1e4fcc826aa7

#### `stripe docs <path>`

Opens a docs page by its path. In a terminal it launches an interactive browser; with `--no-tui` (or when called from an AI agent) it renders plain Markdown to stdout.

https://github.com/user-attachments/assets/4580ef20-846d-4538-9f73-23a5dfbc5227

---

#### `stripe docs search <query>`

Searches docs.stripe.com.

https://github.com/user-attachments/assets/d0c8f646-651c-46d5-98f6-55aa295833db

---

#### `stripe docs api <resource>` — API reference lookup

Looks up API reference pages by resource name (`product`), HTTP method + path (`GET /v1/products`), or event name (`product.created`). Resolves the canonical reference page and renders it.

https://github.com/user-attachments/assets/137e2e25-e7e8-4196-a831-ffb18b3c9152

---

#### `stripe docs skills list` — agent skills

Lists AI agent skills available from docs.stripe.com, with install instructions for connecting them to Claude Code, Cursor, Copilot, etc.

https://github.com/user-attachments/assets/c781893a-1818-430a-a6dc-5aa0b925b789

---

As the subcommand was developed as a plugin outside of this repository, it contains several new packages at `internal/`. To keep this PR as simple as possible and easier to review, minimal changes have been done to migrate the code from https://github.com/stripe/stripe-cli-docs-plugin-go to here.

Post merge, I'll be cleaning up some of the duplicated packages. i.e. `internal/agent` is very similar to `pkg/useragent`.


.github/workflows/canary-test.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v6
         with:
-          go-version: '1.26.0'
+          go-version: '1.26.3'
```
tomer-stripe: do yall need go 1.26.3? if not would prefer to keep these normal and do the workflow bump separately

internal/agent/agent.go
tomer-stripe, line 1: can you use the detectors for this we already have in the cli? this might be duplicative

internal/agentskills/skills.go
tomer-stripe, line 1: how is this different than github.com/stripe/ai? if these are just all of our skills, i don't think having it be namespaced to `stripe docs` is necessarily right

internal/browser/browser.go
tomer-stripe, line 1: can you use https://github.com/stripe/stripe-cli/blob/master/pkg/open/open.go

tomer-stripe: or maybe update that? this may be better 😅 (you have tests). just be careful since we've run into issues where tests will open the browser so we stub that out in a few places

internal/browser/browser.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"os/exec"
+	"runtime"
+)
+
+var allowedHosts = map[string]bool{
+	"docs.stripe.com": true,
```
tomer-stripe: dashboard.stripe.com?
marketplace.stripe.com?

how comprehensive should this be?

pkg/docs/cache.go
```diff
@@ -0,0 +1,97 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+)
+
+const defaultCacheTTL = 1 * time.Hour
```
tomer-stripe, line 14: why is the cache so short? we don't update docs that often, right?

pkg/docs/cache.go
```diff
@@ -0,0 +1,97 @@
...
+// Cache defines the interface for storing and retrieving page content.
+type Cache interface {
+	Get(key string) (data []byte, cachedAt time.Time, ok bool, err error)
+	Set(key string, data []byte) error
+}
+
+// FSCache implements Cache using the local filesystem with a TTL-based
+// eviction policy. Entries older than the configured TTL are treated as
+// misses and removed on access.
+type FSCache struct {
```
tomer-stripe, line 25: can you use afero? it abstracts system-specific fs issues and also provides an in-memory fs that makes testing easier. we use it in other places in the cli

pkg/docs/cache.go
```diff
@@ -0,0 +1,97 @@
...
+	dir string
```
tomer-stripe, line 27: are you using `XDG_CONFIG` or anything similar? we should have a default dir but a user should allow to override it if they want

also im generally assuming the cache is going into `~/.config/stripe`. right?

internal/docs/cache_test.go
```diff
@@ -0,0 +1,137 @@
...
+	assert.Nil(t, got)
+
+	entries, _ := os.ReadDir(dir)
+	assert.Empty(t, entries)
+}
+
+func TestFSCache_Get_UnreadableFile(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("chmod 000 does not restrict read access on Windows")
+	}
```
tomer-stripe: do we need windows specific tests?

internal/docs/cache_test.go
```diff
@@ -0,0 +1,137 @@
...
+	require.NoError(t, os.Chmod(path, 0o000))
+	t.Cleanup(func() { os.Chmod(path, 0o644) })
+
+	_, _, _, err = cache.Get("key") //nolint:dogsled
+	assert.Error(t, err)
+}
+
+func TestFSCache_Set_UnwritableDir(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("chmod 0555 does not restrict write access on Windows")
```
tomer-stripe: under what conditions would the cache dir be unwritable? i feel like that would break things like cli login since we need to write to config files

pkg/docs/client.go
tomer-stripe, line 1: answer might be no but can you use any of our base requests package here? https://github.com/stripe/stripe-cli/blob/1bc0f2f1c611d2e28e1b3686acfb526552100ab8/pkg/requests/base.go

there's some overlap/duplication but i know this is more docs-specific with caching

pkg/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+	"time"
+
+	"github.com/stripe/stripe-cli/internal/agentskills"
+)
+
+const defaultBaseURL = "https://docs.stripe.com"
+
+// Page holds the content and metadata of a fetched documentation page.
+type Page struct {
+	// Content is the raw page body returned by docs.stripe.com.
```
tomer-stripe, line 22: markdown or html?

pkg/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+// ClientOption configures a Client.
+type ClientOption func(*Client)
+
+// WithBaseURL overrides the default docs.stripe.com base URL.
+func WithBaseURL(u string) ClientOption {
+	return func(c *Client) { c.baseURL, _ = url.Parse(u) }
+}
+
+// WithHTTPClient sets a custom http.Client.
+func WithHTTPClient(hc *http.Client) ClientOption { return func(c *Client) { c.http = hc } }
```
tomer-stripe, line 59: under what circumstances would we need a custom http.client?

internal/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// WithHTTPClient sets a custom http.Client.
+func WithHTTPClient(hc *http.Client) ClientOption { return func(c *Client) { c.http = hc } }
+
+// WithCache enables response caching for FetchPage.
+func WithCache(cache Cache) ClientOption { return func(c *Client) { c.cache = cache } }
+
+// WithLogger sets a custom logger.
+func WithLogger(logger *slog.Logger) ClientOption { return func(c *Client) { c.logger = logger } }
```
tomer-stripe: under what circumstances would we need custom logger?

internal/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+	URL   string `json:"url"`
+}
+
+// Client fetches documentation pages and calls endpoints on docs.stripe.com.
+type Client struct {
+	http      *http.Client
+	baseURL   *url.URL
+	userAgent string
+	cache     Cache
+	logger    *slog.Logger
```
tomer-stripe: we're using github.com/sirupsen/logrus in other places, can you use that?

internal/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+
+// WithAgent appends an AIAgent token to the User-Agent header. Pass the result
+// of string(agent.Detect()) — a non-empty value is appended as " AIAgent/<name>".
+func WithAgent(a string) ClientOption {
+	return func(c *Client) {
+		if a != "" {
+			c.userAgent += fmt.Sprintf(" AIAgent/%s", a)
+		}
+	}
+}
```
tomer-stripe: should use cli's internal detector here

internal/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+
+// NewClient creates a Client configured with the given plugin version.
+func NewClient(version string) *Client {
+	base, _ := url.Parse(defaultBaseURL)
+	client := http.Client{Timeout: 10 * time.Second}
+
+	return &Client{
+		http:      &client,
+		baseURL:   base,
+		userAgent: fmt.Sprintf("stripe-cli docs-plugin/%s (%s; %s; %s)", version, runtime.GOOS, runtime.GOARCH, runtime.Version()),
```
tomer-stripe: not a plugin anymore! just cli :P

tomer-stripe: actually you should use our normal user agent generator so it's consistent with our telemetry

pkg/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+//
+//	page, err := c.FetchPage(ctx, &url.URL{Path: "/payments/accept-a-payment", RawQuery: "api_version=2024-06-30"})
+func (c *Client) FetchPage(ctx context.Context, ref *url.URL) (Page, error) {
+	resolvedURL := c.baseURL.ResolveReference(ref)
+	// url.Values.Encode() already sorts params, but we normalize here to ensure
+	// a consistent cache key regardless of how callers construct RawQuery.
+	resolvedURL.RawQuery = resolvedURL.Query().Encode()
+	rawURL := resolvedURL.String()
+
+	if c.cache != nil {
```
tomer-stripe, line 98: maybe because i haven't had enough coffee yet but this slightly tripped me up. i think this check is to make sure we have a cache folder, not that the page itself is cached, right? are the cases where there's no cache be because of internal error?

internal/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+func (c *Client) FetchPage(ctx context.Context, ref *url.URL) (Page, error) {
+	resolvedURL := c.baseURL.ResolveReference(ref)
+	// url.Values.Encode() already sorts params, but we normalize here to ensure
+	// a consistent cache key regardless of how callers construct RawQuery.
+	resolvedURL.RawQuery = resolvedURL.Query().Encode()
+	rawURL := resolvedURL.String()
+
+	if c.cache != nil {
+		if data, cachedAt, ok, err := c.cache.Get(rawURL); err != nil {
+			c.logger.Error("cache read failed", "url", rawURL, "err", err)
```
tomer-stripe: is a cache miss also an error here?

pkg/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	req.Header.Set("Accept", "text/plain")
+
+	res, err := c.do(req)
+	if err != nil {
+		return Page{}, err
+	}
+
+	if c.cache != nil {
+		if err := c.cache.Set(rawURL, res.body); err != nil {
```
tomer-stripe, line 125: something about having this check twice in the same function feels off. would this change somehow between the top of the call and here?

internal/docs/client.go
```diff
@@ -0,0 +1,239 @@
...
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return response{}, fmt.Errorf("docs: read body: %w", err)
+	}
+
+	return response{body: body, finalURL: resp.Request.URL}, nil
+}
+
+// FetchSkills retrieves the skills index from docs.stripe.com.
+func (c *Client) FetchSkills(ctx context.Context) (*agentskills.Index, error) {
```
tomer-stripe: isn't the source of truth github? is docs just proxying that?

pkg/docs/markdown/testdata/showcase.dark.golden
tomer-stripe, line 1: what is this?

pkg/docs/markdown/styles.go
tomer-stripe, line 1: im fine with this for now but i want to start standardizing colors/styles/etc for the cli and plugins. i suspect we'll end up moving and changing this later on

pkg/docs/pager/pager.go
```diff
@@ -0,0 +1,116 @@
...
+// Pager wraps a writer, optionally piping output through a terminal pager.
+type Pager struct {
+	w   io.Writer
+	pw  io.WriteCloser
+	cmd *exec.Cmd
+}
+
+// New returns a Pager that pipes through a terminal pager when w is a TTY and
+// enabled is true. The caller must call Close when done writing.
+func New(w io.Writer, enabled bool) *Pager {
```
tomer-stripe, line 25: struggling with this a little, why do we need this (the pager piece specifically)? could you share more about what this is doing and how it's being used?

pkg/docs/spinner/spinner.go
tomer-stripe, line 1: cc @yahanxing-stripe -- this is fine for now but i think we'll replace this later with something standardized

pkg/docs/tui/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package tui provides a Bubble Tea TUI for browsing Stripe documentation
```
tomer-stripe, line 1: should the entire tui here be scoped just to the docs experience? i don't know if everyone will want/use the same tui docs has

pkg/docs/tui/filter.go
```diff
@@ -0,0 +1,35 @@
...
+	"time"
+
+	tea "charm.land/bubbletea/v2"
+)
+
+const mouseThrottleDur = 15 * time.Millisecond
+
+// NewMouseEventFilter rate-limits mouse events before they reach the model.
+// This reduces the amount of in-flight wheel input during shutdown.
+func NewMouseEventFilter() func(tea.Model, tea.Msg) tea.Msg {
```
tomer-stripe, line 14: were you running into issues with this? i hadn't noticed anything but my testing's been very limited

internal/tui/model.go
```diff
@@ -0,0 +1,630 @@
...
+		return m, cmd
+	}
+
+	if !m.isLanding() {
+		m.viewport, cmd = m.viewport.Update(msg)
+	}
+	return m, cmd
+}
+
+// applyPage installs a newly loaded page into the model.
```
tomer-stripe: what does it mean to `install` a newly loaded page into the model? that word is throwing me off a bit.

pkg/docs/tui/page.go
```diff
@@ -0,0 +1,38 @@
...
+	URL     *url.URL
+}
+
+// Copy writes the raw markdown content to the system clipboard.
+func (p Page) Copy() error {
+	if err := clipboard.WriteAll(string(p.Content)); err != nil {
+		return fmt.Errorf("copying to clipboard: %w", err)
+	}
+	return nil
+}
```
tomer-stripe, line 29: does this work on every os? do we need os-specific testing?

pkg/docs/ui/styles.go
```diff
@@ -0,0 +1,80 @@
+// Package ui provides shared visual styles and components for the TUI.
```
tomer-stripe, line 1: noted this elsewhere but will probably standardize and change these in the future. fine for now though!

pkg/cmd/docs/api.go
```diff
@@ -0,0 +1,72 @@
...
+
+Look up by resource name:
+
+  docs api product
+  docs api customer
+
+Look up by HTTP method and path:
+
+  docs api GET /v1/products
+  docs api POST /v1/products/{id}
```
tomer-stripe: does this work by default or do we need quotes so the shell doesn't interpret as paths

pkg/cmd/docs/api.go
```diff
@@ -0,0 +1,72 @@
...
+
+Look up by event type:
+
+  docs api charge.succeeded
+  docs api payment_intent.created`,
```
tomer-stripe: should this be `docs event`? it's not technically an api

pkg/cmd/docs/api.go
tomer-stripe: oh wait it's coming back to me. i think this is fine

pkg/cmd/docs/root.go
```diff
@@ -0,0 +1,322 @@
...
+// RootCommand is the root command for the docs plugin.
+type RootCommand struct {
+	cmd *cobra.Command
+
+	cfg *cliconfig.Config
+
+	client       *docs.Client
+	renderer     markdown.Renderer
+	rendererOpts []markdown.RendererOption
+	logger       *slog.Logger
```
tomer-stripe: let's use our standard logger here too

pkg/cmd/docs/root.go
```diff
@@ -0,0 +1,322 @@
...
+
+	noPager bool
+	noTUI   bool
```
tomer-stripe: could you add `--non-interactive` and alias one of these to that? that's what we've been using in other places for a while (and what we're asking plugins to standardize on)

pkg/cmd/docs/root.go
```diff
@@ -0,0 +1,322 @@
...
+  docs /payments
+  docs /api/customers
+
+Search documentation by keywords:
+
+  docs search "payment intents"
+
+Read API Reference pages by their identifier:
+
+  docs api product
```
tomer-stripe: shouldn't this be `products`? everything is pluralized

pkg/cmd/docs/root.go
```diff
@@ -0,0 +1,322 @@
...
+  docs api product.created`,
+		Args:              cobra.ArbitraryArgs,
+		PersistentPreRunE: r.preRun,
+		RunE:              r.run,
+		SilenceUsage:      true,
+	}
+
+	agentDetected := agent.Detect() != agent.NotDetected
+	r.cmd.PersistentFlags().BoolVar(&r.noPager, "no-pager", agentDetected, "Write output directly to stdout")
+	r.cmd.PersistentFlags().BoolVar(&r.noTUI, "no-tui", agentDetected, "Write output directly without the interactive browser")
```
tomer-stripe: are we hiding this based on whether we detect an agent?

pkg/cmd/docs/skills.go
tomer-stripe, line 1: noted this elsewhere but i think this one will be worth pulling into a top level `stripe skills` command. we should talk to reese+nilofer+folks about that

tomer-stripe (commented):
couple high level notes but my high level is let's make this conform better to the existing design patterns in the cli. few main things
* let's reuse existing packages where we have them
* we don't have a top-level `internal` command. let's create a `docs` specific one in `pkg` for anything there
* i think we'll want to make `skills` its own top level command, not namespaced by `docs`. let's talk to folks about that one

internal/agent/agent.go
joelzwarrington: Fixed in 33fe9f3. Removed `internal/agent` entirely and replaced all usages with `pkg/useragent.DetectAIAgent`.

internal/agentskills/skills.go
joelzwarrington: Agreed, removed the `skills` subcommand and the `agentskills` package entirely in 3b98e2c. We'll revisit as a standalone top-level command once there's alignment on placement.

internal/browser/browser.go
joelzwarrington: Done in 81d0782. Moved `OpenURL` (with context + host allowlist validation) into `pkg/open`, migrated the tests there, and deleted `internal/browser`. I didn't update the `Browser` function since I didn't want to have to switch all of the consumers.

internal/docs/client.go
joelzwarrington: Switched to logrus in c579963. The custom `*slog.Logger` field is now a `*logrus.Entry` backed by the global logrus instance, and the `charm.land/log` handler is gone.

internal/docs/client.go
joelzwarrington: Switched to logrus in `c579963`. The custom `*slog.Logger` field is now a `*logrus.Entry` backed by the global logrus instance, and the `charm.land/log` handler is gone.

pkg/docs/client.go
joelzwarrington: This is a dependency injection tool. It allows us to test the CLI without relying on modifying globals. Added benefit of keeping our command(s) unaware of the specific HTTP client customizations set

internal/docs/client.go
joelzwarrington: Addressed in 33fe9f3.

internal/docs/client.go
joelzwarrington: Addressed in 33fe9f3.

internal/browser/browser.go
joelzwarrington: This can/should be configurable to the invocation. I've added a `OpenURL` function which accepts allowed hosts.

pkg/docs/cache.go
joelzwarrington: We update docs all of the time!

I think this value was blindly chosen, and could likely be increased. But I'm not very worried about this for now. I'd like to get some usage metrics first to see what would be a good balance.

pkg/docs/cache.go
joelzwarrington: Yes, updated in d54f7fa.

pkg/docs/cache.go
joelzwarrington: Yes, in our root command setup it passes the directory!

https://github.com/stripe/stripe-cli/pull/1663/changes#diff-74aa8af2a86d0e4722ceb23a73b4dc0b74d6a88ee0e06ce5a462c24893b0401aR153

internal/docs/cache_test.go
joelzwarrington: No longer needed when using afero

internal/docs/cache_test.go
joelzwarrington: This test is no longer necessary with the switch to afero

pkg/docs/client.go
joelzwarrington: Looking at pkg/requests/base.go it looks very coupled to the Stripe API. I think it makes sense to keep these separate, and possibly abstract later if needed.

pkg/docs/client.go
joelzwarrington: This is Markdown, but I can see how it's not explicit in the comment/naming.

pkg/docs/client.go
joelzwarrington: The two guards are separate concerns. The cache is intentionally optional (i.e. we always want to fetch live documentation), so we need to check on read and write.

pkg/docs/client.go
joelzwarrington: having the cache be nilable is intentional. it can be nil when we always want to fetch live docs. also this function is checking it twice on separate operations (1 for read and this 1 for writing) i don't really see how we could extract this logic without significantly increasing indirection or making it harder to read

internal/docs/client.go
joelzwarrington: docs is source of truth, github syncs from docs

pkg/docs/markdown/styles.go
joelzwarrington: sounds good. I think it'd be nice to standardize colours in here or in the /ui package!

pkg/docs/pager/pager.go
joelzwarrington: Without a pager, long docs pages scroll past the terminal with no way to scroll back. The abstraction handles this branching: when stdout is a TTY and the pager is enabled, it spawns `less` (or `$PAGER`) and pipes output through it; when stdout is piped, an agent is detected, or `--no-pager` is set, it falls through to a direct write. Callers just call `Write` and `Close` without caring which path they're on.

pkg/docs/tui/doc.go
joelzwarrington: yep, this was focused on just a docs tui, will consider moving this

pkg/cmd/docs/skills.go
joelzwarrington: yes! removed the skills subcommand

pkg/cmd/docs/root.go
joelzwarrington: no, `product` is the actual API entity.

CleanShot 2026-06-18 at 12 41 15

pkg/cmd/docs/root.go
joelzwarrington: changed wording to be `--non-interactive`

pkg/cmd/docs/root.go
joelzwarrington: swapped for the standard logger

pkg/cmd/docs/api.go
joelzwarrington: I don't think it's worth the distinction personally! events are technically API entities ;)

pkg/cmd/docs/root.go
joelzwarrington: Yes, the default is set based on wether an agent is detected!

.github/workflows/canary-test.yml
joelzwarrington: downgraded. had a dep which required 1.26, implemented it within the project (palette implementation)

internal/tui/model.go
joelzwarrington: updated wording which was confusing!

pkg/cmd/docs/root.go
```diff
@@ -0,0 +1,301 @@
...
+	r := &RootCommand{}
+
+	r.cmd = &cobra.Command{
+		Use:   "docs <path>",
+		Short: "Browse docs.stripe.com from the terminal",
+		Long: `Browse docs.stripe.com from the terminal.
+
+Read a page by path:
+
+  docs /payments
```
vcheung-stripe: Can we update all the examples to say `stripe docs ...` instead of just `docs ...`? In `api` and `search` as well.

pkg/cmd/docs/root.go
```diff
@@ -0,0 +1,301 @@
...
+  docs api product.created`,
+		Args:              cobra.ArbitraryArgs,
+		PersistentPreRunE: r.preRun,
+		RunE:              r.run,
+		SilenceUsage:      true,
+	}
+
+	agentDetected := useragent.DetectAIAgent(os.Getenv) != ""
+	r.cmd.PersistentFlags().BoolVar(&r.noPager, "no-pager", agentDetected, "Write output directly to stdout")
+	r.cmd.PersistentFlags().BoolVar(&r.nonInteractive, "non-interactive", agentDetected, "Write output directly without the interactive browser")
```
vcheung-stripe, line 98: It seems like `--no-pager` only works when `--non-interactive` is set. I wonder if `--no-pager` should just automatically make it non-interactive. Not sure what the right thing to do is.

pkg/cmd/docs/search.go
```diff
@@ -0,0 +1,105 @@
...
+		RunE: r.runSearch,
+	}
+}
+
+func (r *RootCommand) runSearch(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("search: missing search query argument")
+	}
+
+	query := strings.Join(args, " ")
```
vcheung-stripe, line 48: Wonder if we should have this take one arg wrapped in quotes rather than joining the args together?

vcheung-stripe (changes requested):
Only blocking thing I have is to update the help text examples, otherwise lgtm!

pkg/cmd/docs/search.go
joelzwarrington: It accepts both. So you can do `stripe docs search "How do I accept a payment"` or succinctly `stripe docs search payment intents`. I'd love to keep this like it is!

pkg/cmd/docs/root.go
joelzwarrington: I updated it so that `--no-pager` will work on it's own and not use the TUI.

pkg/cmd/docs/root.go
joelzwarrington: Yes, updated!

vcheung-stripe (approved):

PR #1674, opened 2026-06-15
1/12 Add co-op session domain model

tomer-stripe:
## What this adds
Introduces the core co-op session types, step state machine, session navigation helpers, and SDK snippet fetching used by later agent and TUI layers.

## Review focus
Review this PR as one slice of the co-op stack. It should be understandable on its own against `tomer/coop-stack-base`.

## Stack
Base: `tomer/coop-stack-base`  
Head: `tomer/coop-split-domain-core`

## Validation

### DIFF
--- a/pkg/coop/session.go
+++ b/pkg/coop/session.go
@@ -1,0 +1,230 @@
+package coop
+
+import (
+	"fmt"
+	"time"
+)
+
+// validTransitions defines allowed state transitions.
+var validTransitions = map[NodeState][]NodeState{
+	NodePending: {NodeActive, NodeSkipped},
+	NodeActive:  {NodeReview, NodeDone, NodeSkipped},
+	NodeReview:  {NodeDone, NodeActive, NodeSkipped}, // active = rejected, redo
+}
+
+// TotalNodes returns the total number of nodes across all steps.
+func (s *Session) TotalNodes() int {
+	count := 0
+	for _, ch := range s.Steps {
+		count += len(ch.Nodes)
+	}
+	return count
+}
+
+// NodeByNumber returns a pointer to the node at the given 1-based index,
+// counting sequentially across steps. Returns an error if out of range.
+func (s *Session) NodeByNumber(n int) (*SessionNode, error) {
+	if n < 1 {
+		return nil, fmt.Errorf("node number must be >= 1, got %d", n)
+	}
+	idx := 0
+	for i := range s.Steps {
+		for j := range s.Steps[i].Nodes {
+			idx++
+			if idx == n {
+				return &s.Steps[i].Nodes[j], nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("node %d out of range (session has %d nodes)", n, s.TotalNodes())
+}
+
+// StepByNodeNumber returns the step containing a 1-based node number.
+func (s *Session) StepByNodeNumber(n int) (*SessionStep, int, int, error) {
+	if n < 1 {
+		return nil, -1, -1, fmt.Errorf("node number must be >= 1, got %d", n)
+	}
+	idx := 0
+	for i := range s.Steps {
+		for j := range s.Steps[i].Nodes {
+			idx++
+			if idx == n {
+				return &s.Steps[i], i, j, nil
+			}
+		}
+	}
+	return nil, -1, -1, fmt.Errorf("node %d out of range (session has %d nodes)", n, s.TotalNodes())
+}
+
+// ReviewGranularityForNode returns the effective review granularity for a node.
+func (s *Session) ReviewGranularityForNode(n int) ReviewGranularity {
+	ch, _, _, err := s.StepByNodeNumber(n)
+	if err != nil || ch.ReviewGranularity == "" {
+		return ReviewGranularityNode
+	}
+	return ch.ReviewGranularity
+}
+
+// StepReadyForReview returns true when every reviewable node in a step is
+// no longer pending or active.
+func (s *Session) StepReadyForReview(stepIndex int) bool {
+	if stepIndex < 0 || stepIndex >= len(s.Steps) {
+		return false
+	}
+	hasReviewable := false
+	for _, n := range s.Steps[stepIndex].Nodes {
+		if n.AutoConfirm {
+			continue
+		}
+		hasReviewable = true
+		switch n.State {
+		case NodeReview, NodeDone, NodeSkipped:
+		default:
+			return false
+		}
+	}
+	return hasReviewable
+}
+
+// StepHasReview returns true if any node in the step is waiting for human review.
+func (s *Session) StepHasReview(stepIndex int) bool {
+	if stepIndex < 0 || stepIndex >= len(s.Steps) {
+		return false
+	}
+	for _, n := range s.Steps[stepIndex].Nodes {
+		if n.State == NodeReview {
+			return true
+		}
+	}
+	return false
+}
+
+// FirstReviewNodeInStep returns the 1-based node number of the first review node.
+func (s *Session) FirstReviewNodeInStep(stepIndex int) int {
+	if stepIndex < 0 || stepIndex >= len(s.Steps) {
+		return 0
+	}
+	nodeNumber := 0
+	for i := range s.Steps {
+		for j := range s.Steps[i].Nodes {
+			nodeNumber++
+			if i == stepIndex && s.Steps[i].Nodes[j].State == NodeReview {
+				return nodeNumber
+			}
+		}
+	}
+	return 0
+}
+
+// FirstActiveNodeInStep returns the 1-based node number of the first active node.
+func (s *Session) FirstActiveNodeInStep(stepIndex int) int {
+	if stepIndex < 0 || stepIndex >= len(s.Steps) {
+		return 0
+	}
+	nodeNumber := 0
+	for i := range s.Steps {
+		for j := range s.Steps[i].Nodes {
+			nodeNumber++
+			if i == stepIndex && s.Steps[i].Nodes[j].State == NodeActive {
+				return nodeNumber
+			}
+		}
+	}
+	return 0
+}
+
+// ActiveNode returns the first node in active state, or nil.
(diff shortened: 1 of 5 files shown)


pkg/coop/session.go
```diff
@@ -0,0 +1,226 @@
...
+var validTransitions = map[StepState][]StepState{
+	StepPending: {StepActive, StepSkipped},
+	StepActive:  {StepReview, StepDone, StepSkipped},
+	StepReview:  {StepDone, StepActive, StepSkipped}, // active = rejected, redo
+}
+
+// TotalSteps returns the total number of nodes across all chapters.
+func (s *Session) TotalSteps() int {
+	count := 0
+	for _, ch := range s.Chapters {
```
kwu-stripe: Just a heads up that we've renamed chapters to steps in the Blueprint's code.

pkg/coop/session.go
```diff
@@ -0,0 +1,226 @@
...
+		count += len(ch.Nodes)
```
kwu-stripe, line 19: I think this could still make sense, since we can mark some chapters and nodes are skipped, but it's possible that not all chapters / nodes will apply to all users.

pkg/coop/types.go
```diff
@@ -0,0 +1,148 @@
...
+	Check  string `json:"check"`
+	Passed bool   `json:"passed"`
+}
+
+// APIRequest describes the expected API call for a node.
+type APIRequest struct {
+	Key    string      `json:"key,omitempty"`
+	Path   string      `json:"path"`
+	Method string      `json:"method"`
+	Params interface{} `json:"params,omitempty"`
```
kwu-stripe, line 72: In the mongo blueprints we're also going to list params not to include.

pkg/coop/types.go
```diff
@@ -0,0 +1,148 @@
...
+	Nodes             []SessionNode     `json:"nodes"`
+}
+
+// Session is the shared state file between agent and TUI.
+type Session struct {
+	SchemaVersion   int               `json:"schema_version"`
+	ID              string            `json:"id"`
+	Blueprint       string            `json:"blueprint"`
+	Status          SessionStatus     `json:"status"`
+	Settings        map[string]string `json:"settings,omitempty"`
```
kwu-stripe, line 109: In mongo blueprints we're splitting the settings up into two halves: settings and params. Settings are going to be things that are consistent across the implementation (ex. platform vs account) while params are things that are variables (ex. account name).

pkg/coop/types.go
```diff
@@ -0,0 +1,148 @@
...
+	Chapters        []SessionChapter  `json:"chapters"`
+	ClaimURL        string            `json:"claim_url,omitempty"`
```
kwu-stripe: What's the claim URL for?

pkg/coop/session.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	return false
+}
+
+// FirstReviewStepInChapter returns the 1-based step number of the first review node.
+func (s *Session) FirstReviewStepInChapter(chapterIndex int) int {
+	if chapterIndex < 0 || chapterIndex >= len(s.Chapters) {
+		return 0
+	}
+	step := 0
```
kwu-stripe: nit: can we rename step to node, since in the core blueprints we are renaming chapters to steps.

pkg/coop/session.go
```diff
@@ -0,0 +1,226 @@
...
+			if i == chapterIndex && s.Chapters[i].Nodes[j].State == StepReview {
+				return step
+			}
+		}
+	}
+	return 0
+}
+
+// FirstActiveStepInChapter returns the 1-based step number of the first active node.
+func (s *Session) FirstActiveStepInChapter(chapterIndex int) int {
```
kwu-stripe: nit: could we also rename step here to node.

pkg/coop/snippet.go
```diff
@@ -0,0 +1,61 @@
...
+
+var httpClient = &http.Client{Timeout: 5 * time.Second}
+
+// FetchSDKSnippet fetches a generated SDK code snippet from the docs endpoint.
+// path is the API path (e.g. "/v1/customers"), method is "post"/"get"/etc,
+// params is the JSON-encoded request params, and language is "node"/"python"/etc.
+func FetchSDKSnippet(path, method string, params interface{}, language string) (string, error) {
+	argsJSON := "{}"
+	if params != nil {
+		data, err := json.Marshal(params)
```
kwu-stripe, line 22: If json.Marshal fails then the parameters would get dropped. Should we return the error or get notified about it in some way?

pkg/coop/session.go
```diff
@@ -0,0 +1,226 @@
...
+	if !valid {
+		return fmt.Errorf("invalid transition: step %d is %q, cannot move to %q", n, node.State, to)
+	}
+
+	node.State = to
+	now := time.Now().UTC()
+
+	switch to {
+	case StepActive:
+		node.StartedAt = &now
```
kwu-stripe: If a step is rejected and reactivated, startedAt is overwritten and the original time is lost. Did you want to track total time-in-flight?

pkg/coop/session.go
```diff
@@ -0,0 +1,226 @@
...
+		node.CompletedAt = nil
+	case StepDone:
+		node.CompletedAt = &now
+		node.RejectionNote = ""
+	case StepReview:
+		node.CompletedAt = &now
+	}
```
kwu-stripe, line 189: We could set a completedAt time if we chose to skip a step, so that it looks different from a step that we haven't touched at all.

kwu-stripe (changes requested):
I left some in-line comments:
- I'll try to make sure this stays in sync with Mongo blueprints. 
- Chapters are getting renamed to steps.

pkg/coop/session.go
tomer-stripe: oh cool! will fix. couldn't remember if we did that or not.

so it's steps as the group, and nodes as the lower-level actions?

pkg/coop/session.go
tomer-stripe: good point. how would we determine that in the code? ideally the tui is generated from the blueprint itself so i'd want the blueprint driving this

pkg/coop/types.go
tomer-stripe: this comes from the sandboxes creation flow. so if you do `stripe sandboxes create` we give you a url to the dashboard to convert the sandbox into a full account

i wanted to keep track of that throughout the flow so we could prompt the user to go create an account

pkg/coop/session.go
tomer-stripe: yep! will do

tomer-stripe:
thank you! do we have the blueprint schemas in our docs somewhere? agree we should stay in sync

tomer-stripe:
@kwu-stripe resolved most comments. the two i have open am curious about what the shape is before we add them here

kwu-stripe:
> thank you! do we have the blueprint schemas in our docs somewhere? agree we should stay in sync

I think its a little bit rough right now because we have both the typescript and mongo blueprints, which have slightly differing schemas. There's this [trailhead doc](https://trailhead.corp.stripe.com/docs/workbench/blueprints/blueprints-schema?entrypoint=home_search_result) that is mostly up to date for typescript blueprints. Once we move into Mongo blueprints and it becomes more stable, I will create a new schema doc. I think with that doc we could have it be [auto-generated](https://trailhead.corp.stripe.com/docs/trailhead/working-in-trailhead/auto-generate-docs) from the code.

pkg/coop/types.go
kwu-stripe: It's the same shape, just another field called `excluded_params`.

pkg/coop/session.go
kwu-stripe: For the typescript blueprints:
- chapter inclusion is determined by [includeIfConfigMatchesOne](https://stripe.sourcegraphcloud.com/r/stripe-internal/mint/-/blob/pay-server/frontend/workbench/shared/blueprints/src/types/Blueprint.tsx?L58)
- node inclusion is determined by both [includeIfConfigMatchesOne and hiddenIfConfigIncludesOne](https://stripe.sourcegraphcloud.com/r/stripe-internal/mint/-/blob/pay-server/frontend/workbench/shared/blueprints/src/types/BlueprintNode.tsx?L75-76)

For Mongo blueprints:
- chapter inclusion is determined by [is_included](https://stripe.sourcegraphcloud.com/r/stripe-internal/mint/-/blob/pay-server/lib/workbench/blueprints/model/step.rb?L363-368)
- node inclusion is determined by [is_included](https://stripe.sourcegraphcloud.com/r/stripe-internal/mint/-/blob/pay-server/lib/workbench/blueprints/model/step.rb?L313-317).

pkg/coop/types.go
```diff
@@ -0,0 +1,149 @@
...
+	Settings        map[string]string `json:"settings,omitempty"`
+	Params          map[string]string `json:"params,omitempty"`
+	Steps           []SessionStep     `json:"steps"`
+	UsedSandbox     bool              `json:"used_sandbox,omitempty"`
+	NextSteps       *NextStepsState   `json:"next_steps,omitempty"`
+	ParentSessionID string            `json:"parent_session_id,omitempty"`
+	ParentStepID    string            `json:"parent_step_id,omitempty"` // which next-step this session fulfills
+	CreatedAt       time.Time         `json:"created_at"`
+	UpdatedAt       time.Time         `json:"updated_at"`
+	Version         int               `json:"version"`
```
kwu-stripe, line 118: What are `schemaVersion` vs `Version`? When I was initially reading this file I thought `schemaVersion` was the version for the schema (the params), which is why it's set to 2 right now, and `version` is how we keep track of updates to the session, but when reading https://github.com/stripe/stripe-cli/pull/1675/changes it looks like we're updating schemaVersion.

kwu-stripe: Nevermind, I see that the updates to schemaVersion are for a version migration now.

kwu-stripe (approved):

PR #1675, opened 2026-06-15
2/12 Add co-op session store

tomer-stripe:
## What this adds
Adds JSON session persistence, optimistic version checks, lock files, heartbeat files, session listing, latest-session lookup, deletion, and Windows-safe session replacement.

## Review focus
Review this PR as one slice of the co-op stack. It should be understandable on its own against `tomer/coop-split-domain-core`.

## Stack
Base: `tomer/coop-split-domain-core`  
Head: `tomer/coop-split-domain-store`

## Validation

### DIFF
--- a/pkg/coop/store.go
+++ b/pkg/coop/store.go
@@ -1,0 +1,389 @@
+package coop
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"sort"
+	"strings"
+	"time"
+)
+
+// Store manages session persistence with atomic writes.
+type Store struct {
+	baseDir string
+}
+
+var (
+	ErrInvalidSessionID = errors.New("invalid session id")
+	ErrSessionNotFound  = errors.New("session not found")
+	ErrVersionConflict  = errors.New("version conflict")
+	ErrLockTimeout      = errors.New("timed out waiting for session lock")
+	ErrCorruptSession   = errors.New("corrupt session")
+)
+
+var (
+	sessionLockTimeout      = 5 * time.Second
+	sessionLockPollInterval = 25 * time.Millisecond
+)
+
+// NewStore creates a Store, ensuring the coop directory exists.
+func NewStore(configFolder string) (*Store, error) {
+	dir := filepath.Join(configFolder, "coop")
+	if err := os.MkdirAll(dir, 0700); err != nil {
+		return nil, fmt.Errorf("creating coop directory: %w", err)
+	}
+	return &Store{baseDir: dir}, nil
+}
+
+// NewStoreAt creates a Store at a specific path (for testing).
+func NewStoreAt(dir string) (*Store, error) {
+	if err := os.MkdirAll(dir, 0700); err != nil {
+		return nil, fmt.Errorf("creating coop directory: %w", err)
+	}
+	return &Store{baseDir: dir}, nil
+}
+
+func (s *Store) sessionPath(id string) (string, error) {
+	// Validate ID to prevent path traversal
+	base := filepath.Base(id)
+	if base != id || id == "" || id == "." || id == ".." {
+		return "", fmt.Errorf("%w: %q", ErrInvalidSessionID, id)
+	}
+	return filepath.Join(s.baseDir, id+".json"), nil
+}
+
+// Write atomically persists a session (write to .tmp then rename).
+// Uses optimistic locking: checks that the file's current version matches
+// the session's version before writing. Returns an error on conflict.
+func (s *Store) Write(session *Session) error {
+	path, err := s.sessionPath(session.ID)
+	if err != nil {
+		return err
+	}
+	return s.writePath(path, session)
+}
+
+func (s *Store) writePath(path string, session *Session) error {
+	unlock, err := s.acquireSessionLock(path)
+	if err != nil {
+		return err
+	}
+	defer unlock()
+
+	if existing, err := os.ReadFile(path); err == nil {
+		var current Session
+		if json.Unmarshal(existing, &current) == nil {
+			if current.Version != session.Version {
+				return fmt.Errorf("%w: expected %d, file has %d", ErrVersionConflict, session.Version, current.Version)
+			}
+		}
+	}
+
+	ensureSessionSchemaVersion(session)
+	session.UpdatedAt = time.Now().UTC()
+	session.Version++
+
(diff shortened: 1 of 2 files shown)


pkg/coop/store.go
```diff
@@ -0,0 +1,414 @@
...
+			if current.Version != session.Version {
+				// Re-read to get latest state
+				return fmt.Errorf("%w: expected %d, file has %d", ErrVersionConflict, session.Version, current.Version)
+			}
+		}
+	}
+
+	if session.SchemaVersion == 0 {
+		session.SchemaVersion = CurrentSessionSchemaVersion
+	}
```
kwu-stripe: The schema migration on here only migrates v0 to v2. `Update` (L225) migrates v0 to v1 then v0 and v1 to v2, and `Read` (L178) migrates v0 to v1 in memory. These three paths should share a `migrateSession` helper: right now adding v3 means updating three separate places.

pkg/coop/store.go
```diff
@@ -0,0 +1,414 @@
...
+		if err == nil {
+			f.Close()
+			return func() { os.Remove(lockPath) }, nil
+		}
+		if !os.IsExist(err) {
+			return nil, fmt.Errorf("creating lock file: %w", err)
+		}
+
+		if info, statErr := os.Stat(lockPath); statErr == nil && time.Since(info.ModTime()) > 30*time.Second {
+			os.Remove(lockPath)
```
kwu-stripe: In between the these two lines it's technically possible that another process could have released the stale lock and created a new valid lock, which the os.Remove would then remove.

kwu-stripe: If we're not expecting much concurrent access, this should be fine though.

pkg/coop/store.go
```diff
@@ -0,0 +1,414 @@
...
+		tmp.Close()
+		return fmt.Errorf("writing temp file: %w", err)
+	}
+	if err := tmp.Close(); err != nil {
+		return fmt.Errorf("closing temp file: %w", err)
+	}
+
+	if err := replaceSessionFile(tmpPath, path); err != nil {
+		return err
+	}
```
kwu-stripe: Nit: I think we could refactor the code in writeUnlocked and Update a bit, so writePath gets the lock and does the version check then calls writeUnlocked.

pkg/coop/store.go
```diff
@@ -0,0 +1,414 @@
...
+
+	var ids []string
+	for _, e := range entries {
+		if e.IsDir() || !strings.HasSuffix(e.Name(), ".json") {
+			continue
+		}
+		if strings.HasSuffix(e.Name(), ".tmp") {
+			continue
+		}
+		ids = append(ids, strings.TrimSuffix(e.Name(), ".json"))
```
kwu-stripe: This is dead code.

pkg/coop/store.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+	return os.Remove(path)
+}
+
+func (s *Store) heartbeatPath(id string) string {
+	path, err := s.sessionPath(id)
+	if err != nil {
+		return filepath.Join(s.baseDir, "invalid.heartbeat")
+	}
+	return path + ".heartbeat"
```
kwu-stripe: If session path fails this still returns without notifying about an error, which causes the other heartbeat functions to use that file without knowing that something is wrong.

pkg/coop/store_test.go
kwu-stripe, line 1: Should we also add tests for version conflict and the heartbeat functions?

pkg/coop/store.go
```diff
@@ -0,0 +1,414 @@
...
+		if err != nil {
+			continue
+		}
+		entries = append(entries, entry{id: id, modTime: mt})
+	}
+
+	sort.Slice(entries, func(i, j int) bool {
+		return entries[i].modTime.After(entries[j].modTime)
+	})
+
```
kwu-stripe, line 322: Nit: This is the same as the beginning of `latestSession` so maybe we could extract it.

kwu-stripe (changes requested):
Left some in-line comments but overall it looks good to me:
- We could clean up the code migrating the SchemaVersion.
- There are some places where I think we could extract and re-use code.

pkg/coop/store.go
```diff
@@ -0,0 +1,388 @@
...
+	if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("removing existing session file: %w", err)
+	}
+	if err := os.Rename(tmpPath, path); err != nil {
+		return fmt.Errorf("renaming temp file: %w", err)
+	}
+	return nil
+}
+
+func (s *Store) acquireSessionLock(path string) (func(), error) {
```
kwu-stripe, line 111: The 30 second stale lock cleanup was removed: the previous implementation had a possible race condition, but I think we should still have some sort of clean up for stale locks.

pkg/coop/store.go
```diff
@@ -0,0 +1,388 @@
...
+		return nil, fmt.Errorf("%w: parsing session %q: %v", ErrCorruptSession, id, err)
+	}
+	migrateSession(&session)
+
+	return &session, nil
+}
+
+func migrateSession(session *Session) {
+	if session.SchemaVersion == 0 {
+		session.SchemaVersion = 1
```
kwu-stripe: Is there a reason we're not just updating the schema version to the currentSessionSchemaVersion?

pkg/coop/store.go
```diff
@@ -0,0 +1,388 @@
...
+
+	var session Session
+	if err := json.Unmarshal(data, &session); err != nil {
+		return nil, fmt.Errorf("%w: parsing session %q: %v", ErrCorruptSession, id, err)
+	}
+	migrateSession(&session)
+	if err := fn(&session); err != nil {
+		return nil, err
+	}
+	migrateSession(&session)
```
kwu-stripe: I think we can just update migrateSession to migrate it to the current version and then call it once in this function.

kwu-stripe (changes requested):
Left some in-line comments:
- I think we should still have some sort of stale lock cleanup.
- A small fix to the way we migrate the schema version,

pkg/coop/store.go
tomer-stripe: no you are right. i had this because i was bouncing between a few branches but this is unnecessary

kwu-stripe (approved):

PR #1676, opened 2026-06-15
3/12 Add co-op blueprint loader

tomer-stripe:
## What this adds
Adds embedded blueprint loading, prefix matching, metadata listing, session creation from blueprints, and the minimal fixture catalog needed to validate loader behavior.

## Review focus
Review this PR as one slice of the co-op stack. It should be understandable on its own against `tomer/coop-split-domain-store`.

## Stack
Base: `tomer/coop-split-domain-store`  
Head: `tomer/coop-split-blueprint-core`

## Validation

### DIFF
--- a/pkg/coop/blueprint.go
+++ b/pkg/coop/blueprint.go
@@ -1,0 +1,207 @@
+package coop
+
+import (
+	"embed"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+//go:embed blueprints/*.json
+var blueprintFS embed.FS
+
+// BlueprintNode is a node definition within a blueprint step.
+type BlueprintNode struct {
+	Type          NodeType    `json:"type"`
+	Key           string      `json:"key"`
+	Title         string      `json:"title"`
+	Description   string      `json:"description,omitempty"`
+	ReviewPrompt  string      `json:"review_prompt,omitempty"`
+	ReviewCommand string      `json:"review_command,omitempty"`
+	AutoConfirm   bool        `json:"auto_confirm,omitempty"`
+	Request       *APIRequest `json:"request,omitempty"`
+	Events        []string    `json:"events,omitempty"`
+}
+
+// BlueprintStep is a step definition within a blueprint.
+type BlueprintStep struct {
+	Key               string            `json:"key"`
+	Title             string            `json:"title"`
+	Description       string            `json:"description,omitempty"`
+	ReviewGranularity ReviewGranularity `json:"review_granularity,omitempty"`
+	Required          bool              `json:"required,omitempty"`
+	Nodes             []BlueprintNode   `json:"nodes"`
+}
+
+// Blueprint is the CLI-friendly representation of a Workbench Blueprint.
+type Blueprint struct {
+	ID          string             `json:"id"`
+	Title       string             `json:"title"`
+	Description string             `json:"description,omitempty"`
+	Prompt      string             `json:"prompt,omitempty"`
+	Type        string             `json:"type"`
+	Products    []string           `json:"products,omitempty"`
+	Settings    []BlueprintSetting `json:"settings,omitempty"`
+	Params      []BlueprintParam   `json:"params,omitempty"`
+	Steps       []BlueprintStep    `json:"steps"`
+}
+
+// BlueprintSetting defines a configurable setting for a blueprint.
+type BlueprintSetting struct {
+	Key         string   `json:"key"`
+	Title       string   `json:"title"`
+	Description string   `json:"description,omitempty"`
+	Type        string   `json:"type"`
+	Default     string   `json:"default,omitempty"`
+	Options     []string `json:"options,omitempty"`
+}
+
+// BlueprintParam defines a variable value used while implementing a blueprint.
+type BlueprintParam struct {
+	Key         string   `json:"key"`
+	Title       string   `json:"title"`
+	Description string   `json:"description,omitempty"`
+	Type        string   `json:"type"`
+	Default     string   `json:"default,omitempty"`
+	Options     []string `json:"options,omitempty"`
+}
+
+// LoadBlueprint loads a blueprint by ID from the embedded filesystem.
+// If no exact match is found, it tries prefix matching (e.g. "deploy" → "deploy-stripe-projects").
+func LoadBlueprint(id string) (*Blueprint, error) {
+	filename := fmt.Sprintf("blueprints/%s.json", id)
+	data, err := blueprintFS.ReadFile(filename)
+	if err != nil {
+		// Try prefix match
+		match, matchErr := prefixMatchBlueprint(id)
+		if matchErr != nil {
+			return nil, matchErr
+		}
+		data, err = blueprintFS.ReadFile(fmt.Sprintf("blueprints/%s.json", match))
+		if err != nil {
+			return nil, fmt.Errorf("blueprint %q not found", id)
+		}
+	}
+
+	var bp Blueprint
+	if err := json.Unmarshal(data, &bp); err != nil {
+		return nil, fmt.Errorf("parsing blueprint %q: %w", id, err)
+	}
+
+	return &bp, nil
+}
+
+func prefixMatchBlueprint(prefix string) (string, error) {
+	ids, err := ListBlueprints()
+	if err != nil {
+		return "", fmt.Errorf("blueprint %q not found", prefix)
+	}
+
+	var matches []string
+	for _, id := range ids {
+		if strings.HasPrefix(id, prefix) {
+			matches = append(matches, id)
+		}
+	}
+
+	switch len(matches) {
+	case 0:
+		return "", fmt.Errorf("blueprint %q not found", prefix)
+	case 1:
+		return matches[0], nil
+	default:
+		return "", fmt.Errorf("ambiguous blueprint prefix %q matches: %s", prefix, strings.Join(matches, ", "))
+	}
+}
+
+// ListBlueprints returns all available blueprint IDs.
+func ListBlueprints() ([]string, error) {
+	entries, err := blueprintFS.ReadDir("blueprints")
+	if err != nil {
+		return nil, err
+	}
+
+	var ids []string
+	for _, e := range entries {
+		if e.IsDir() || !strings.HasSuffix(e.Name(), ".json") {
+			continue
+		}
+		ids = append(ids, strings.TrimSuffix(e.Name(), ".json"))
+	}
+	return ids, nil
+}
+
+// ListBlueprintsWithMetadata returns all blueprints with their metadata.
+func ListBlueprintsWithMetadata() ([]Blueprint, error) {
+	ids, err := ListBlueprints()
+	if err != nil {
+		return nil, err
+	}
+
+	var blueprints []Blueprint
+	for _, id := range ids {
+		bp, err := LoadBlueprint(id)
+		if err != nil {
+			continue
+		}
+		blueprints = append(blueprints, *bp)
+	}
+	return blueprints, nil
+}
+
+// NewSessionFromBlueprint creates a new Session from a Blueprint definition.
+// A context-gathering step is prepended so the agent scans the project first.
+func NewSessionFromBlueprint(bp *Blueprint, sessionID string, settings, params map[string]string) *Session {
+	// Prepend a context-gathering step (auto-confirmed, no human sign-off needed)
+	contextStep := SessionStep{
+		Key:   "context-step",
+		Title: "Project context",
+		Nodes: []SessionNode{
+			{
+				Key:         "scan-project",
+				Type:        NodeTestHelper,
+				Title:       "Understand the project",
+				Description: "Scan the codebase to identify language, framework, dependencies, and existing Stripe code. Report what you find.",
+				AutoConfirm: true,
+				State:       NodePending,
+			},
+		},
+	}
+
+	steps := make([]SessionStep, 0, len(bp.Steps)+1)
+	steps = append(steps, contextStep)
+
+	for _, ch := range bp.Steps {
+		nodes := make([]SessionNode, len(ch.Nodes))
+		for j, n := range ch.Nodes {
+			nodes[j] = SessionNode{
+				Key:           n.Key,
+				Type:          n.Type,
+				Title:         n.Title,
+				Description:   n.Description,
+				ReviewPrompt:  n.ReviewPrompt,
+				ReviewCommand: n.ReviewCommand,
+				AutoConfirm:   n.AutoConfirm,
+				State:         NodePending,
+				Request:       n.Request,
+				Events:        n.Events,
+			}
+		}
+		steps = append(steps, SessionStep{
+			Key:               ch.Key,
+			Title:             ch.Title,
+			ReviewGranularity: ch.ReviewGranularity,
+			Nodes:             nodes,
+		})
+	}
+
+	return &Session{
+		SchemaVersion: CurrentSessionSchemaVersion,
+		ID:            sessionID,
+		Blueprint:     bp.ID,
+		Status:        SessionActive,
+		Settings:      settings,
+		Params:        params,
+		Steps:         steps,
+	}
+}
--- a/pkg/coop/blueprint.go
+++ b/pkg/coop/blueprint.go
@@ -14,9 +14,11 @@ var blueprintFS embed.FS
 // BlueprintStep is a step definition within a blueprint.
 type BlueprintStep struct {
 	StepDefinition
-	Description string           `json:"description,omitempty"`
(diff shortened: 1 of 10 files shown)


pkg/coop/blueprints/flat-fee-and-overages.json
```diff
@@ -0,0 +1,151 @@
...
+      "nodes": [
+        {
+          "type": "apiRequest",
+          "key": "createCustomer",
+          "title": "Create a customer",
+          "request": {
+            "path": "/v1/customers",
+            "method": "post"
+          },
+          "review_prompt": "Confirm the implementation calls the intended Stripe API and reuses any IDs needed by later steps."
```
kwu-stripe: How're you getting the review prompt for the blueprint?

pkg/coop/blueprint.go
kwu-stripe, line 1: Some of the types here are a subset of the corresponding type in pkg/coop/types.go:

- BlueprintNode => SessionNode
- BlueprintStep => SessionStep
- Blueprint => Session

Could we make the type in pkg/coop/types.go reference these in some way so that updating this would also update that type?

pkg/coop/blueprint.go
```diff
@@ -0,0 +1,207 @@
...
+	Request       *APIRequest `json:"request,omitempty"`
+	Events        []string    `json:"events,omitempty"`
+}
+
+// BlueprintStep is a step definition within a blueprint.
+type BlueprintStep struct {
+	Key               string            `json:"key"`
+	Title             string            `json:"title"`
+	Description       string            `json:"description,omitempty"`
+	ReviewGranularity ReviewGranularity `json:"review_granularity,omitempty"`
```
kwu-stripe: Maybe this should be under Blueprint (and also the same thing for Session).

pkg/coop/blueprint.go
```diff
@@ -0,0 +1,207 @@
...
+	Required          bool              `json:"required,omitempty"`
+	Nodes             []BlueprintNode   `json:"nodes"`
+}
+
+// Blueprint is the CLI-friendly representation of a Workbench Blueprint.
+type Blueprint struct {
+	ID          string             `json:"id"`
+	Title       string             `json:"title"`
+	Description string             `json:"description,omitempty"`
+	Prompt      string             `json:"prompt,omitempty"`
```
kwu-stripe: What is prompt / how do we get it? I'm not seeing this on the stored blueprints in this PR.

pkg/coop/blueprint.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	return &Session{
+		SchemaVersion: CurrentSessionSchemaVersion,
+		ID:            sessionID,
+		Blueprint:     bp.ID,
+		Status:        SessionActive,
+		Settings:      settings,
+		Params:        params,
+		Steps:         steps,
```
kwu-stripe, line 188: I think this should also set `CreatedAt` to the current time.

pkg/coop/blueprint.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, fmt.Errorf("parsing blueprint %q: %w", id, err)
+	}
+
+	return &bp, nil
+}
+
+func prefixMatchBlueprint(prefix string) (string, error) {
+	ids, err := ListBlueprints()
+	if err != nil {
+		return "", fmt.Errorf("blueprint %q not found", prefix)
```
kwu-stripe: I think a more accurate error message here might be that it couldn't load the list of blueprints, or we could just pass on the error we got.

pkg/coop/blueprint.go
```diff
@@ -0,0 +1,207 @@
...
+	ids, err := ListBlueprints()
+	if err != nil {
+		return nil, err
+	}
+
+	var blueprints []Blueprint
+	for _, id := range ids {
+		bp, err := LoadBlueprint(id)
+		if err != nil {
+			continue
```
kwu-stripe: Do we want to log an error about failing to load a blueprint?

pkg/coop/blueprints/flat-fee-and-overages.json
kwu-stripe: Nevermind I realized it's the same for every apiRequest.

pkg/coop/blueprint_test.go
```diff
@@ -0,0 +1,332 @@
...
+	assert.Equal(t, "Understand the project", session.Steps[0].Nodes[0].Title)
+
+	// All nodes should be pending
+	for _, ch := range session.Steps {
+		for _, n := range ch.Nodes {
+			assert.Equal(t, NodePending, n.State)
+		}
+	}
+
+	// Total steps = blueprint steps (6) + context step (1)
```
kwu-stripe: nit: steps here should be nodes

pkg/coop/blueprints/flat-fee-and-overages.json
```diff
@@ -0,0 +1,151 @@
...
+        },
+        {
+          "type": "apiRequest",
+          "key": "addGraduatedRateToRateCard",
+          "title": "Add graduated rate to rate card",
+          "request": {
+            "path": "/v2/billing/rate_cards/${node.create-rate-card-step.createRateCard.createRateCardRequest:id}/rates",
```
kwu-stripe: I'm updating interpolation for the mongo blueprints so that we don't have specific request IDs on the API request node anymore because each API request node can only have one request: this would become `node.create-rate-card-step.createRateCard:id`.

pkg/coop/blueprints/one-time-payment.json
```diff
@@ -0,0 +1,116 @@
+{
+  "steps": [
+    {
+      "description": "Create a product and price in your Stripe account.",
+      "key": "setup-step",
+      "nodes": [
+        {
+          "description": "Create a Stripe Product with inline default_price_data, then save the returned default_price ID for the Checkout Session. Do not put generated IDs in .env files.",
```
kwu-stripe: Nit: can you ask the AI to make the order of the fields in each object the same (so the node would be ordered: type, key, title etc.)

pkg/coop/blueprints/one-time-payment.json
```diff
@@ -0,0 +1,116 @@
...
+          "key": "create-product",
+          "request": {
+            "key": "create-product-request",
```
kwu-stripe: We won't need to have the key on the API request for Mongo blueprints.

pkg/coop/blueprints/one-time-payment.json
```diff
@@ -0,0 +1,116 @@
...
+          "description": "Create a Checkout Session with the price from the previous step.",
+          "key": "create-checkout-session",
+          "request": {
+            "key": "create-checkout-session-request",
+            "method": "post",
+            "params": {
+              "cancel_url": "https://example.com/cancel",
+              "line_items": [
+                {
+                  "price": "${node.setup-step.create-product.create-product-request:default_price}",
```
kwu-stripe: Same update for interpolation (won't need to reference create-product-request).

pkg/coop/blueprints/one-time-payment.json
```diff
@@ -0,0 +1,116 @@
...
+          "type": "uiComponent",
+          "review_prompt": "Open the app and confirm the user-facing flow works as described."
+        },
+        {
+          "description": "Create a page to show after the customer completes payment.",
+          "key": "add-success-page",
+          "title": "Add a success page",
+          "type": "uiComponent",
+          "review_prompt": "Open the app and confirm the user-facing flow works as described."
+        }
```
kwu-stripe, line 70: This is a new node that isn't in the blueprint in pay-server.

pkg/coop/blueprints/one-time-payment.json
```diff
@@ -0,0 +1,116 @@
...
+        {
+          "description": "Run through the complete payment flow using a test card number.",
+          "key": "test-payment-flow",
+          "title": "Make a test payment",
+          "type": "testHelper",
+          "review_prompt": "Run the described test flow and confirm the expected result is observable."
+        }
+      ],
+      "title": "Test the integration"
+    }
```
kwu-stripe: This also isn't in the blueprint in payserver, maybe we should append this for all of the blueprints kind of like we prepend the context-step.

pkg/coop/blueprints/flat-fee-and-overages.json
```diff
@@ -0,0 +1,151 @@
...
+            "method": "post"
+          },
+          "review_prompt": "Confirm the implementation calls the intended Stripe API and reuses any IDs needed by later steps."
+        },
+        {
+          "type": "apiRequest",
+          "key": "attachRateCardToPricingPlan",
+          "title": "Attach rate card to Pricing Plan",
+          "request": {
+            "path": "/v2/billing/pricing_plans/${node.create-pricing-plan-step.createEmptyPricingPlan.createEmptyPricingPlanRequest:id}/components",
```
kwu-stripe: Same comment about interpolation.

pkg/coop/blueprints/flat-fee-and-overages.json
```diff
@@ -0,0 +1,151 @@
...
+            "method": "post"
+          },
+          "review_prompt": "Confirm the implementation calls the intended Stripe API and reuses any IDs needed by later steps."
+        },
+        {
+          "type": "apiRequest",
+          "key": "attachLicenseFeeToPricingPlan",
+          "title": "Attach license fee to Pricing Plan",
+          "request": {
+            "path": "/v2/billing/pricing_plans/${node.create-pricing-plan-step.createEmptyPricingPlan.createEmptyPricingPlanRequest:id}/components",
```
kwu-stripe: Same comment about interpolation.

pkg/coop/blueprints/flat-fee-and-overages.json
```diff
@@ -0,0 +1,151 @@
...
+            "method": "post"
+          },
+          "review_prompt": "Confirm the implementation calls the intended Stripe API and reuses any IDs needed by later steps."
+        },
+        {
+          "type": "apiRequest",
+          "key": "setLiveVersion",
+          "title": "Set live version to latest",
+          "request": {
+            "path": "/v2/billing/pricing_plans/${node.create-pricing-plan-step.createEmptyPricingPlan.createEmptyPricingPlanRequest:id}",
```
kwu-stripe: Sam comment about interpolation.

pkg/coop/blueprints/setup-future-payments.json
```diff
@@ -0,0 +1,109 @@
...
+          "description": "Redirect the customer to the Checkout Session URL.",
+          "key": "redirect-to-setup",
+          "title": "Redirect to setup page",
+          "type": "uiComponent",
+          "review_prompt": "Open the app and confirm the user-facing flow works as described."
+        },
+        {
+          "description": "Listen for the setup_intent.succeeded event to confirm the payment method was saved.",
+          "events": [
+            "setup_intent.succeeded"
```
kwu-stripe: The pay-server blueprint checks for checkout.session.completed.

pkg/coop/blueprints/flat-fee-and-overages.json
kwu-stripe: Also I think `create-rate-card-step` doesn't exist.

kwu-stripe (changes requested):
Left some in-line comments but in general:
- The types for Blueprint are also used in Session so we should make Session reference Blueprint.
- The interpolation is changing for API requests nodes requests: we won't have the request ID.
- I don't think the blueprints match what we have in pay-server and some of the interpolation is broken (I didn't fully check all of them right now but will go through again in a later view). It's ok that they're not 1-1 right now, just wanted to make sure we're aware of this.
- Some other small things.

pkg/coop/blueprint.go
tomer-stripe: i think i want to remove this entirely. after trying it out for a while im fairly sure the right review granularity is always "step". there was a while where i was confirming each node but that's way too noisy.

pkg/coop/blueprint.go
tomer-stripe: let me remove this. i was experimenting with some custom extra prompts but i really want to lean on the blueprint as the canonical artifact and not supplement it locally

pkg/coop/types.go
```diff
@@ -32,15 +32,6 @@ const (
 	NodeSetUpWebhooks NodeType = "setUpWebhooks"
 )
 
-// ReviewGranularity controls where human approval happens.
```
kwu-stripe, line 35: Does this mean that we don't want to have configurable review granularity?

pkg/coop/blueprint.go
```diff
@@ -0,0 +1,190 @@
...
+	"fmt"
+	"strings"
+	"time"
+)
+
+//go:embed blueprints/*.json
+var blueprintFS embed.FS
+
+// BlueprintStep is a step definition within a blueprint.
+type BlueprintStep struct {
```
kwu-stripe, line 15: For mongo blueprints we're going to introduce settings and params on blueprint steps so that we can reuse steps between blueprints.

pkg/coop/blueprints/flat-fee-and-overages.json
```diff
@@ -0,0 +1,187 @@
...
+          }
+        },
+        {
+          "type": "apiRequest",
+          "key": "createMeter",
+          "title": "Create a meter",
+          "review_prompt": "Confirm the implementation calls the intended Stripe API and reuses any IDs needed by later steps.",
+          "request": {
+            "path": "/v1/billing/meters",
+            "method": "post"
```
kwu-stripe: This needs to have a unique event_name on every run.

pkg/coop/blueprints/flat-fee-and-overages.json
```diff
@@ -0,0 +1,187 @@
...
+      "required": true,
+      "nodes": [
+        {
+          "type": "apiRequest",
+          "key": "createRateCard",
+          "title": "Create a rate card",
+          "review_prompt": "Confirm the implementation calls the intended Stripe API and reuses any IDs needed by later steps.",
+          "request": {
+            "path": "/v2/billing/rate_cards",
+            "method": "post"
```
kwu-stripe: Missing the display_name, service_interval, service_interval_count, currency, tax_behavior.

kwu-stripe (changes requested):
Do we want to get rid of the configurable review granularity? Also, for the blueprint definitions, some of them are missing params that are in the pay-server blueprints: some of these params are required for the API requests.

pkg/coop/blueprints/flat-subscription-with-entitlements.json
```diff
@@ -0,0 +1,174 @@
...
+      "required": true,
+      "nodes": [
+        {
+          "type": "apiRequest",
+          "key": "create-checkout-session",
+          "title": "Create a checkout session",
+          "review_prompt": "Confirm the implementation calls the intended Stripe API and reuses any IDs needed by later steps.",
+          "request": {
+            "path": "/v1/checkout/sessions",
+            "method": "post"
```
kwu-stripe: I think this node is missing required params.

pkg/coop/blueprints/flat-subscription-with-entitlements.json
```diff
@@ -0,0 +1,174 @@
...
+      "key": "next-billing-cycle-chapter",
+      "title": "Advance to the next billing cycle",
+      "required": true,
+      "nodes": [
+        {
+          "type": "testHelper",
+          "key": "advance-time",
+          "title": "Advance time to billing date",
+          "description": "Run the clock advance and confirm it moves to the next billing cycle.",
+          "review_prompt": "Run the described test flow and confirm the expected result is observable."
```
kwu-stripe: I'm not sure how we want to support test helpers. For this one there is technically a public API they could hit but that won't always be the case.

pkg/coop/blueprints/flat-subscription-with-entitlements.json
```diff
@@ -0,0 +1,174 @@
...
+          "description": "Run the test-clock setup and confirm the subscription billing flow can use the created clock.",
+          "review_prompt": "Run the described test flow and confirm the expected result is observable."
+        },
+        {
+          "type": "testHelper",
+          "key": "create-customer",
+          "title": "Create a customer",
+          "description": "Run the customer setup and confirm the customer is attached to the test clock.",
+          "review_prompt": "Run the described test flow and confirm the expected result is observable."
+        }
```
kwu-stripe, line 108: Same for these test helper nodes.

kwu-stripe (changes requested):
This looks good overall, I'm not sure how we want to handle test helper nodes though. For the ones in these blueprints there are public API endpoints we could give them.

kwu-stripe (approved):

PR #1678, opened 2026-06-15
5/12 Add co-op workflow service

tomer-stripe:
## What this adds
Adds the workflow service that applies agent lifecycle transitions, review/next-action helper logic, command responses, snippet lookup integration, and workflow tests.

## Review focus
Review this PR as one slice of the co-op stack. It should be understandable on its own against `tomer/coop-split-blueprint-catalog`.

## Stack
Base: `tomer/coop-split-blueprint-catalog`  
Head: `tomer/coop-split-workflow`

## Validation
Ran `go test ./pkg/coop/... -count=1` and `golangci-lint run --max-issues-per-linter=0 --max-same-issues=0 ./pkg/coop/...`.

### DIFF
--- a/pkg/coop/helpers/nextaction.go
+++ b/pkg/coop/helpers/nextaction.go
@@ -10,6 +10,9 @@ import (
 )
 
 var ErrNoSession = errors.New("no session found")
+var ErrSelectionTimeout = errors.New("timed out waiting for next-action selection")
+
+const NextActionSelectionTimeout = 10 * time.Minute
 
 type Input struct {
 	SessionID string
@@ -97,8 +100,16 @@ func ShowSuggestions(store Store, session *coop.Session, suggestions []Suggestio
 }
 
 func WaitForSelection(store Store, sessionID string) (string, error) {
+	return waitForSelection(store, sessionID, NextActionSelectionTimeout, time.Now, time.Sleep)
+}
+
+func waitForSelection(store Store, sessionID string, timeout time.Duration, now func() time.Time, sleep func(time.Duration)) (string, error) {
+	deadline := now().Add(timeout)
 	for {
-		time.Sleep(500 * time.Millisecond)
+		if now().After(deadline) {
+			return "", ErrSelectionTimeout
+		}
+		sleep(500 * time.Millisecond)
 		session, err := store.Read(sessionID)
 		if err != nil {
 			continue
--- a/pkg/coop/helpers/nextaction.go
+++ b/pkg/coop/helpers/nextaction.go
@@ -1,0 +1,214 @@
+// Package helpers contains shared support logic for co-op commands and workflows.
+package helpers
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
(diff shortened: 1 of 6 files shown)


pkg/coop/helpers/nextaction.go
```diff
@@ -0,0 +1,214 @@
...
+		if err := store.Write(session); err != nil {
+			return fmt.Errorf("marking next-action completed: %w", err)
+		}
+	}
+	return nil
+}
+
+func WaitForSelection(store Store, sessionID string) (string, error) {
+	for {
+		time.Sleep(500 * time.Millisecond)
```
kwu-stripe: Do we want to add a timeout?

pkg/coop/helpers/review.go
```diff
@@ -0,0 +1,21 @@
+package helpers
+
+import "github.com/stripe/stripe-cli/pkg/coop"
+
+func NextPendingNodeInStep(session *coop.Session, stepIndex, afterNode int) int {
+	nodeNumber := 0
```
kwu-stripe, line 6: Nit: it's a little confusing to have 0 indexed step and 1 indexed node

pkg/coop/helpers/review.go
```diff
@@ -0,0 +1,21 @@
...
+			nodeNumber++
+			if i == stepIndex && nodeNumber > afterNode && session.Steps[i].Nodes[j].State == coop.NodePending {
+				return nodeNumber
+			}
+		}
+	}
+	return 0
+}
+
+func StepReviewApplies(session *coop.Session, nodeNumber int) bool {
```
kwu-stripe: I think this function basically makes sure that the nodeNumber exists in the session. When would we expect this not to be the case?

kwu-stripe: I think I have a better understanding of how the coop and agent interact now: is this to guard against the agent calling review for an invalid node?

pkg/coop/workflow/service.go
```diff
@@ -0,0 +1,540 @@
...
+}
+
+func (s *Service) ConfirmReview(sessionID string, nodeNumbers []int) (*coop.Session, error) {
+	return s.store.Update(sessionID, func(session *coop.Session) error {
+		for _, nodeNumber := range nodeNumbers {
+			node, err := session.NodeByNumber(nodeNumber)
+			if err != nil {
+				return err
+			}
+			if node.State == coop.NodeDone {
```
kwu-stripe: NodeSkipped is also terminal.

kwu-stripe (approved):
Looks good overall

pkg/coop/helpers/review.go
tomer-stripe: yea. there were a few times i noticed agents putting in completely weird/unexpected numbers so i put this in as a small guardrail

tomer-stripe: was looking at the code, i thiiiink this is actually redundant given some of the other step checks we have in place. i removed it for now and if the agents run into issues again we can re-add

PR #1686, opened 2026-06-16
Add curl-based install script for macOS and Linux

yahanxing-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

- Adds `scripts/install.sh` so users can install the CLI with a single command:
    `curl -sSL https://raw.githubusercontent.com/stripe/stripe-cli/master/scripts/install.sh | sh`
- Detects OS and downloads the latest release from GitHub, verifies SHA256 checksum, and installs to `~/.stripe/bin/stripe`
- Adds the install directory to PATH (supports zsh, bash, fish) and warns if a Homebrew install already exists

### Testing
Linux:
Screenshot 2026-06-16 at 2 28 06 PM

macOS:
Screenshot 2026-06-16 at 2 34 13 PM

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-set -e
+set -euo pipefail
 
 INSTALL_DIR="${STRIPE_INSTALL_DIR:-$HOME/.stripe/bin}"
 GITHUB_REPO="stripe/stripe-cli"
@@ -22,7 +22,7 @@ detect_platform() {
     linux)  OS_LABEL="linux" ;;
     *)
       echo "Error: unsupported operating system: $OS"
-      echo "Supported: macOS, Linux. For Windows, use install.ps1."
+      echo "Supported: macOS, Linux."
       exit 1
       ;;
   esac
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -0,0 +1,168 @@
+#!/bin/sh
+set -e
```
vcheung-stripe: As a convention, could we do this for safety?
```suggestion
set -euo pipefail
```

scripts/install.sh
```diff
@@ -0,0 +1,168 @@
...
+detect_platform() {
+  OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+  ARCH=$(uname -m)
+
+  case "$OS" in
+    darwin) OS_LABEL="mac-os" ;;
+    linux)  OS_LABEL="linux" ;;
+    *)
+      echo "Error: unsupported operating system: $OS"
+      echo "Supported: macOS, Linux. For Windows, use install.ps1."
```
vcheung-stripe: I assume we'll add an `install.ps1` in another PR? If so, could we remove mention of this file until we add it?

scripts/install.sh
```diff
@@ -0,0 +1,168 @@
...
+  esac
+
+  echo "Detected: $OS $ARCH_LABEL"
+}
+
+get_latest_version() {
+  if ! command -v curl >/dev/null 2>&1; then
+    echo "Error: curl is required but not installed."
+    exit 1
+  fi
```
vcheung-stripe: Some linux environments don't have `curl` installed by default but do have `wget`. Could we add that as a fallback?

vcheung-stripe (changes requested):
Overall looks good, just a few small things

vcheung-stripe (approved):

PR #1689, opened 2026-06-16
Remove legacy plugins.toml manifest in favor of local plugin metadata

vzhang-stripe:
## Summary

  This changes plugin install/lookup to treat per-plugin local metadata as the primary persisted state for installed plugins.

  It:
  - backfills missing `plugin-metadata/*.toml` entries for already-installed plugins during startup
  - stops relying on a refreshed/downloaded `plugins.toml` manifest as the active source of truth
  - keeps an existing cached `plugins.toml` only as a compatibility fallback when local metadata is missing, and only if it matches the installed version
  - falls back to the live plugin metadata endpoint when the cached manifest is stale or absent
  - updates latest-version checks to resolve against metadata instead of the cached manifest
  - validates plugin shortnames before constructing install or metadata paths to block path traversal
  - simplifies uninstall to operate from a validated shortname instead of requiring a metadata lookup first

Manually played around with installing, uninstalling, and triggering the migration backfill
Screenshot 2026-06-17 at 3 43 11 PM

  ## Why

  We want to move installed-plugin state off the legacy global manifest and onto per-plugin local metadata without breaking older installs that predate `plugin-metadata/
  *.toml`. This keeps those installs working during the migration while tightening filesystem safety around plugin operations.

  ## Testing

  - `go test ./pkg/plugins ./pkg/cmd/plugin ./pkg/cmd`

### DIFF
--- a/pkg/cmd/plugin/install_test.go
+++ b/pkg/cmd/plugin/install_test.go
@@ -2,7 +2,6 @@ package plugin
 
 import (
 	"context"
-	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -10,7 +9,6 @@ import (
 
 	"github.com/stretchr/testify/require"
 
-	"github.com/stripe/stripe-cli/pkg/requests"
 	"github.com/stripe/stripe-cli/pkg/stripe"
 )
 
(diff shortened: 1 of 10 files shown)


vcheung-stripe (commented):
Could you write down the steps to test this manually?

vcheung-stripe (approved):

PR #1701, opened 2026-06-19
10/12 Add co-op launcher and debug harness

tomer-stripe:
Part 10 of the clean co-op stack targeting `coop`.

This PR adds the co-op start/join launcher path and the hidden debug-agent harness used for local/manual validation.

Changes:
- Adds start, join, launcher, and debug-agent commands.
- Adds launcher/debug-agent tests.
- Adds debug-agent documentation and the tmux validation script.

Review notes:
- The debug agent is intentionally retained as an internal/testing aid.
- This branch was adjusted to use the current step-scoped review helpers and session fixture shape.

### DIFF
--- a/pkg/cmd/coop/coop_join.go
+++ b/pkg/cmd/coop/coop_join.go
@@ -2,12 +2,12 @@ package coopcmd
 
 import (
 	"fmt"
+	"sort"
 
 	"charm.land/huh/v2"
 	"github.com/spf13/cobra"
 
 	"github.com/stripe/stripe-cli/pkg/coop"
-	"github.com/stripe/stripe-cli/pkg/coop/helpers"
 	"github.com/stripe/stripe-cli/pkg/coop/tui"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/cmd/coop/coop_launcher.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	err := helpers.Select(title,
+		[]huh.Option[string]{
+			huh.NewOption("Normal — agent asks before running commands", "normal"),
+			huh.NewOption("Auto-approve — skip all permission prompts (faster, less safe)", "auto"),
+		},
+		&choice,
+	)
+	if err != nil {
```
kwu-stripe, line 106: This turns errors into false and continues launching the agent: if I hit Ctrl-C at the permission prompt, I would expect `coop start` to stop, not launch the agent in normal mode.

pkg/cmd/coop/coop_launcher.go
```diff
@@ -0,0 +1,380 @@
...
+		if cleanup != nil {
+			cleanup()
+		}
+		rc.abortStartedSession(session, "tmux new-session failed")
+		return fmt.Errorf("tmux new-session failed: %w", err)
+	}
+
+	split := exec.Command("tmux", "split-window", "-h", "-t", sessionName, "-p", "60",
+		"bash", "-c", paneCmd)
+	if err := split.Run(); err != nil {
```
kwu-stripe: If the split fails after `tmux new-session` succeeded, we return with the newly-created `stripe-coop` session still running a `coop join` pane. That leaves the user with a stale tmux session and the next `coop start` will hit the "session already exists" prompt. Should we kill `sessionName` on this error path before returning?

pkg/cmd/coop/coop_join.go
```diff
@@ -0,0 +1,136 @@
...
+	return tui.Run(store, session.ID, tui.WithSandboxClaimURL(coopSandboxClaimURL()))
+}
+
+func (jc *coopJoinCmd) pickSession(store *coop.Store) (*coop.Session, error) {
+	ids, err := store.List()
+	if err != nil || len(ids) == 0 {
+		return nil, fmt.Errorf("no sessions found. Start one with 'stripe coop start <blueprint>'")
+	}
+
+	// Load sessions and build options (most recent first)
```
kwu-stripe: I don't think this sorts by time.

kwu-stripe (approved):

kwu-stripe (approved):

PR #1710, opened 2026-06-23
Fix plugin subcommand stubs to pass executing command instead of parent

vzhang-stripe:
Fix plugin subcommand stubs to pass executing command instead of parent Plugin subcommand stubs (e.g. `stripe projects catalog`) were passing `ptc.cmd` (the parent plugin command) to runPluginCmd instead of the actual subcommand being executed. This caused the wrong context to be used and incorrect argument slicing. Now the stub's own `cmd` is passed through, preserving the correct context and args.

Also adds commandContextOrBackground helper to safely handle nil command contexts, and makes runPluginCmd injectable for testing.

Before fix error: `panic: cannot create context from nil parent`

After:

Screenshot 2026-06-23 at 4 53 38 PM

Motivation: https://stripe.slack.com/archives/CFP9HVB7D/p1782251686812209?thread_ts=1781838798.384439&cid=CFP9HVB7D

 ### Reviewers
r? @
cc @stripe/developer-products

 ### Summary


vcheung-stripe (approved):
Confirmed it works:

Ran `stripe plugin install projects@0.20.0`, this is the bad version

`stripe project status` crashes:
Image

It works on this branch:
Image

PR #1727, opened 2026-07-01
Add plugin install/uninstall/upgrade analytics

vcheung-stripe:
Track plugin lifecycle events (Plugin Installed, Plugin Upgraded, Plugin Uninstalled) with name, version, CLI version, OS, CPU arch, a persistent machine UUID stored in config.toml, and merchant ID when authenticated.

Here's where the machine UUID is in the config file, we'll eventually use this in more than just analytics:
```
color = ''
installed_plugins = []
machine_uuid = '44d7bb37-055e-4aa3-9306-a39d7d2e083b'
project-name = 'default'
...
```

I verified the events in our backend logs.

### DIFF
--- a/pkg/cmd/plugin/upgrade.go
+++ b/pkg/cmd/plugin/upgrade.go
@@ -64,6 +64,10 @@ func (uc *UpgradeCmd) runUpgradeCmd(cmd *cobra.Command, args []string) error {
 		}).Debug("Ctrl+C received, cleaning up...")
 	})
 
+	if m := stripe.GetEventMetadata(cmd.Context()); m != nil {
+		m.SetPluginName(args[0])
+	}
+
 	resolvedPlugin, err := plugins.ResolvePluginForUpgrade(ctx, uc.cfg, uc.fs, args[0], uc.apiBaseURL, dashboardBaseURL)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


pkg/cmd/plugin/upgrade.go
```diff
@@ -87,6 +91,8 @@ func (uc *UpgradeCmd) runUpgradeCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	sendPluginLifecycleEvent(cmd.Context(), "Plugin Installed", version)
```
yahanxing-stripe: do we want `plugin upgraded` here?

yahanxing-stripe (approved):

pkg/cmd/plugin/upgrade.go
vcheung-stripe: Yeah I think that would be useful!

vcheung-stripe: Updated to "Plugin Upgraded" when using `stripe plugin upgrade`, or `stripe plugin install` while already having it installed. We could also capture the previous version, but I'm not sure how useful that is.

yahanxing-stripe (approved):
nice!

yahanxing-stripe (approved):

PR #1728, opened 2026-07-01
Add `stripe agent setup` CLI command 

jleong-stripe:
### Reviewers
r? @gusnguyen-stripe 
cc @stripe/developer-products

### Summary

Adds a new `stripe agent setup` command that detects installed AI coding clients, installs the Stripe plugin for each, and offers a client-agnostic Stripe skills installation path for agents without a dedicated plugin.

### What it does

**Detection** — scans the machine for installed AI coding clients and reports their Stripe plugin status:
- **Claude Code** — reads `~/.claude/plugins/installed_plugins.json`
- **Cursor** — scans `~/.cursor/plugins/cache/<marketplace>/stripe/<hash>/` (filesystem-based; Cursor has no CLI installer)
- **Codex CLI** — runs `codex plugin list --json` and matches on `pluginId`/`marketplaceName`

**Installation** — installs the Stripe plugin for each client using its native mechanism:
- **Claude** — `claude plugin install stripe@claude-plugins-official` 
- **Codex** — `codex plugin add stripe@openai-curated`
- **Cursor** — no shell installer exists; surfaces a manual-step hint (`/add-plugin stripe` inside Cursor) and counts it as skipped

**Skills** — native Go skills installer (no npx/Node dependency) that fetches `https://docs.stripe.com/.well-known/skills/index.json` and writes each skill's multi-file tree to `.agents/skills/` (local or global). 

**Agent-aware** — detects the calling agent from environment variables (`CLAUDECODE`, `CODEX_THREAD_ID`, `CURSOR_AGENT`, `GEMINI_CLI`, etc.) and auto-scopes behavior:
- Supported agent (Claude/Cursor/Codex) → installs only that agent's plugin, even with `--yes`
- Unsupported agent (Gemini CLI, Cline, etc.) → installs Stripe skills instead
- Never shows the interactive picker inside an agent (some runtimes allocate a PTY that would otherwise trick TTY detection)

**Interactive TUI** — Bubble Tea multi-select checklist for humans at a terminal:
- Shows detected agents (pre-selected) with a visually separated "Install Stripe skills directly" row below
- Skills row triggers a follow-up scope prompt (local vs global) showing resolved absolute paths
- When no clients are detected, shows supported-client guidance and immediately offers to install skills

### Flags

| Flag                           | Behavior                                                                                              |
| ------------------------------ | ----------------------------------------------------------------------------------------------------- |
| `--status`                     | Read-only: compact, color-coded status table (only shows detected clients)                            |
| `--json`                       | Machine-readable status + planned actions                                                             |
| `--yes`                        | Non-interactive: install all detected clients (human CLI) or just the calling agent (inside an agent) |
| `--client <id>`                | Limit to a single client (explicit override, wins over agent detection)                               |
| `--force`                      | Reinstall even when already installed                                                                 |

### Architecture

```
pkg/agentsetup/          — detection + install providers
  provider.go            — Provider interface, Status, Plan, ActionManual
  scanner.go             — shared Scanner (LookPath/ReadFile/HomeDir/WorkDir), pluginClient, scanPlugin
  claude.go              — ClaudeProvider (marketplace-update retry)
  cursor.go              — CursorProvider (filesystem cache detection, manual-step install)
  codex.go               — CodexProvider (JSON detection, post-install verify)
  *_test.go              — per-provider unit tests

pkg/agentskills/         — native skills installer (net/http only, no npx)
  agentskills.go         — FetchIndex, Install (multi-file: SKILL.md + references/*.md)
  agentskills_test.go    — httptest-based tests

pkg/cmd/
  agent.go               — Cobra command, orchestration, agent-aware selection, install loop
  agent_setup_tui.go     — Bubble Tea multi-select + scope radio prompt
  agent_test.go          — 15+ test cases covering all paths
  agent_setup_tui_test.go — model unit tests
```

### Testing
Demo video (may be stale)

https://github.com/user-attachments/assets/9fa8fadc-377b-4b20-a082-5b4bc96356fc

```bash
# Build the file
~/stripe/mint/gocode/bin/go build -o ./stripe ./cmd/stripe

# Smoke (read-only)
./stripe agent setup --status
./stripe agent setup --json

# Interactive (human terminal)
./stripe agent setup

# Simulate agent context
CODEX_THREAD_ID=x ./stripe agent setup      # scopes to Codex only
GEMINI_CLI=1 ./stripe agent setup            # installs skills (no Gemini plugin)
```

---

### DIFF
--- a/pkg/agentsetup/claude.go
+++ b/pkg/agentsetup/claude.go
@@ -91,7 +91,7 @@ func (p ClaudeProvider) Apply(ctx context.Context, out io.Writer, plan Plan) err
 
 	updateName, updateArgs := ClaudeMarketplaceUpdateCommand()
 	updateCommand := append([]string{updateName}, updateArgs...)
-	fmt.Fprintf(out, "Install failed. Updating Claude plugin marketplace and retrying: %s\n", strings.Join(updateCommand, " "))
+	fmt.Fprintf(out, "First install failed. Updating Claude plugin marketplace and retrying: %s\n", strings.Join(updateCommand, " "))
 	if updateErr := p.RunCommand(ctx, updateName, updateArgs...); updateErr != nil {
 		return fmt.Errorf("running %q after install failed: %w", strings.Join(updateCommand, " "), updateErr)
 	}
--- a/pkg/agentsetup/claude.go
+++ b/pkg/agentsetup/claude.go
@@ -1,0 +1,210 @@
+// Package agentsetup contains helpers for detecting and configuring AI coding
+// agent integrations.
+package agentsetup
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+const (
+	ClientClaudeCode      = "claude-code"
+	ClaudeBinaryName      = "claude"
+	TargetClaudePlugin    = "stripe@claude-plugins-official"
+	LocalClaudePlugin     = "stripe@stripe"
+	ClaudeMarketplace     = "claude-plugins-official"
+	ClaudePluginStatePath = ".claude/plugins/installed_plugins.json"
+
+	StatusNotDetected = "not_detected"
+	StatusInstalled   = "installed"
+	StatusMissing     = "missing"
+	StatusError       = "error"
+)
+
+// LookPathFunc matches exec.LookPath and exists to make detection testable.
+type LookPathFunc func(string) (string, error)
+
+// ReadFileFunc matches os.ReadFile and exists to make status parsing testable.
+type ReadFileFunc func(string) ([]byte, error)
+
+// HomeDirFunc matches os.UserHomeDir and exists to make status parsing testable.
+type HomeDirFunc func() (string, error)
+
+// WorkDirFunc matches os.Getwd and exists to make local plugin scope testable.
+type WorkDirFunc func() (string, error)
+
+// RunCommandFunc runs a command. The production implementation streams stdio.
+type RunCommandFunc func(context.Context, string, ...string) error
+
+// Scanner scans local agent installations without mutating them.
+type Scanner struct {
+	LookPath LookPathFunc
+	ReadFile ReadFileFunc
+	HomeDir  HomeDirFunc
+	WorkDir  WorkDirFunc
+}
+
+// ClaudeStatus is the read-only setup status for Claude Code.
+type ClaudeStatus struct {
+	Client          string `json:"client"`
+	Detected        bool   `json:"detected"`
+	ExecutablePath  string `json:"executable_path,omitempty"`
+	PluginInstalled bool   `json:"plugin_installed"`
+	PluginID        string `json:"plugin_id,omitempty"`
+	PluginVersion   string `json:"plugin_version,omitempty"`
+	PluginScope     string `json:"plugin_scope,omitempty"`
+	PluginProject   string `json:"plugin_project_path,omitempty"`
+	PluginStatePath string `json:"plugin_state_path,omitempty"`
+	Status          string `json:"status"`
+	Error           string `json:"error,omitempty"`
+}
+
+type installedPluginState struct {
+	Version int                             `json:"version"`
+	Plugins map[string][]installedPluginRef `json:"plugins"`
+}
+
+type installedPluginRef struct {
+	Scope       string `json:"scope"`
+	Version     string `json:"version"`
+	Installed   string `json:"installPath"`
+	ProjectPath string `json:"projectPath"`
+}
+
+// DefaultScanner returns a Scanner backed by the real OS.
+func DefaultScanner() Scanner {
+	return Scanner{
+		LookPath: exec.LookPath,
+		ReadFile: os.ReadFile,
+		HomeDir:  os.UserHomeDir,
+		WorkDir:  os.Getwd,
+	}
+}
+
+// ScanClaude returns Claude Code installation and Stripe plugin status.
+func (s Scanner) ScanClaude() ClaudeStatus {
+	s = s.withDefaults()
+
+	status := ClaudeStatus{
+		Client: ClientClaudeCode,
+		Status: StatusNotDetected,
+	}
+
+	claudePath, err := s.LookPath(ClaudeBinaryName)
+	if err != nil {
+		return status
+	}
+	status.Detected = true
+	status.ExecutablePath = claudePath
+	status.Status = StatusMissing
+
+	home, err := s.HomeDir()
+	if err != nil {
+		status.Status = StatusError
+		status.Error = fmt.Sprintf("locating home directory: %v", err)
+		return status
+	}
+	status.PluginStatePath = filepath.Join(home, ClaudePluginStatePath)
+
+	body, err := s.ReadFile(status.PluginStatePath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return status
+		}
+		status.Status = StatusError
+		status.Error = fmt.Sprintf("reading Claude plugin state: %v", err)
+		return status
+	}
+
+	var pluginState installedPluginState
+	if err := json.Unmarshal(body, &pluginState); err != nil {
+		status.Status = StatusError
+		status.Error = fmt.Sprintf("parsing Claude plugin state: %v", err)
+		return status
+	}
+
+	workDir, err := s.WorkDir()
+	if err != nil {
+		status.Status = StatusError
+		status.Error = fmt.Sprintf("locating working directory: %v", err)
+		return status
+	}
+
+	if id, ref, ok := findStripePlugin(pluginState.Plugins, workDir); ok {
+		status.PluginInstalled = true
+		status.PluginID = id
+		status.PluginVersion = ref.Version
+		status.PluginScope = ref.Scope
+		status.PluginProject = ref.ProjectPath
+		status.Status = StatusInstalled
+	}
+
+	return status
+}
+
+// ClaudeInstallCommand returns the command used to install the Stripe Claude
+// Code plugin.
+func ClaudeInstallCommand() (string, []string) {
+	return ClaudeBinaryName, []string{"plugin", "install", TargetClaudePlugin}
+}
+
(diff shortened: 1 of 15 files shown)


vzhang-stripe (changes requested):
Reject skill paths that escape the install directory — `/Users/vzhang/stripe/stripe-cli/pkg/agentskills/agentskills.go:94-98` 

Because both skill.Name and file come from the downloaded index, an absolute skill.Name or a file entry containing ../ will make filepath.Join escape destDir and os.WriteFile overwrite arbitrary files in the project or home directory. The installer should normalize the resolved path and reject anything that does not stay underneath destDir before creating directories or writing content.

pkg/agentsetup/claude.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	name, installArgs := plan.Command[0], plan.Command[1:]
+	if err := p.RunCommand(ctx, name, installArgs...); err == nil {
+		return nil
+	}
+
+	updateName, updateArgs := ClaudeMarketplaceUpdateCommand()
+	updateCommand := append([]string{updateName}, updateArgs...)
+	fmt.Fprintf(out, "Install failed. Updating Claude plugin marketplace and retrying: %s\n", strings.Join(updateCommand, " "))
```
gusnguyen-stripe:
```suggestion
	fmt.Fprintf(out, "First install attempt failed. Updating Claude plugin marketplace and retrying: %s\n", strings.Join(updateCommand, " "))
```

pkg/agentsetup/codex.go
```diff
@@ -0,0 +1,182 @@
...
+	if err := runCommand(ctx, plan.Command[0], plan.Command[1:]...); err != nil {
+		return err
+	}
+
+	// `codex plugin add` exits 0 even when it fails (e.g. the marketplace is not
+	// configured), so the exit code cannot be trusted. Confirm the plugin is
+	// actually installed before reporting success.
+	if _, ok := p.stripePluginVersion(ctx); !ok {
+		return fmt.Errorf("codex reported success but %s is not installed; run `%s` to see the underlying error",
+			TargetCodexPlugin, strings.Join(plan.Command, " "))
```
gusnguyen-stripe, line 142: Nitpicky, but I remember certain codex version did not support plugin install when we test. Can we ask user to upgrade codex if we can't find `codex plugin add`?

pkg/agentsetup/cursor.go
```diff
@@ -0,0 +1,150 @@
...
+	return CursorProvider{Scanner: scanner}
+}
+
+func (p CursorProvider) ID() string { return ClientCursor }
+
+func (p CursorProvider) Detect() Status { return p.Scanner.ScanCursor() }
+
+func (p CursorProvider) Plan(status Status, _ bool) Plan {
+	if status.Detected && !status.Plugin.Installed && status.Status != StatusError {
+		return Plan{Action: ActionManual, Manual: "run /add-plugin stripe inside Cursor to install the Stripe plugin"}
```
gusnguyen-stripe:
```suggestion
		return Plan{Action: ActionManual, Manual: "run /add-plugin stripe inside Cursor Agent to install the Stripe plugin"}
```

pkg/cmd/agent.go
```diff
@@ -0,0 +1,560 @@
...
+		}
+	}
+
+	enc := json.NewEncoder(w)
+	enc.SetIndent("", "  ")
+	if err := enc.Encode(result); err != nil {
+		return err
+	}
+	if len(result.Errors) > 0 {
+		return fmt.Errorf("%s", result.Errors[0])
```
gusnguyen-stripe: Maybe should document why we're only returning the first error

pkg/cmd/agent_setup_tui.go
```diff
@@ -0,0 +1,267 @@
...
+	containerStyle = lipgloss.NewStyle().Border(lipgloss.RoundedBorder()).Padding(1, 2)
+)
+
+func (m selectModel) View() tea.View {
+	body := "Select agents to configure with Stripe:\n\n"
+
+	skillsDividerShown := false
+	for i, r := range m.rows {
+		if r.kind == rowSkills && !skillsDividerShown {
+			body += "\n" + dividerStyle.Render("── or install Stripe skills directly ──") + "\n"
```
gusnguyen-stripe: Assuming we have auto updated skills, should this line just go away? Like we can have both right?

pkg/agentsetup/cursor.go
```diff
@@ -0,0 +1,150 @@
...
+
+	return status
+}
+
+// findCursorStripePlugin looks for a completed Stripe plugin install under
+// ~/.cursor/plugins/cache/<marketplace>/stripe/<hash>/. A `.cache-complete`
+// marker file indicates the install finished.
+func findCursorStripePlugin(s Scanner, pluginsRoot string) (string, cursorPluginRef, bool) {
+	cacheRoot := filepath.Join(pluginsRoot, "cache")
+	marketplaces, err := os.ReadDir(cacheRoot)
```
gusnguyen-stripe: Should this be using scanner instead of os.ReadDir?

pkg/agentsetup/cursor.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		status.Status = StatusError
+		status.Error = fmt.Sprintf("locating home directory: %v", err)
+		return status
+	}
+
+	pluginsRoot := filepath.Join(home, CursorPluginsDir)
+	status.Plugin.StatePath = pluginsRoot
+
+	if id, ref, ok := findCursorStripePlugin(s, pluginsRoot); ok {
```
gusnguyen-stripe: Hmm, when I uninstall the Cursor plugin and run the command it still says `  Cursor already set up`. I think we don't have a reliable way to determine if Cursor has installed Stripe
```
  What actually tracks the plugin: There is no global "installed plugins" registry for Cursor. Instead, Cursor creates a per-project ~/.cursor/projects/<project-name>/mcps/plugin-stripe-stripe/ directory each time
  a project window loads with the plugin active. These directories persist across sessions and accumulate across many projects — **they're not cleaned up on uninstall**.

  The .cache-complete file: Just marks that the download finished. Persists through uninstall. Survives indefinitely. This is what the current scanner checks — which is why you see "installed" even after removing the plugin.
```

I did some investigation and perhaps we can run a sqlite3 command to check the registry
```
gusnguyen@st-gusnguyen3(laptop) ~/.cursor/plugins/cache/cursor-public/stripe % sqlite3 "$HOME/Library/Application Support/Cursor/User/globalStorage/state.vscdb" \
  "SELECT key, value FROM ItemTable WHERE key LIKE 'cursor.plugins.installedIds%' AND value != '[]';"

cursor.plugins.installedIds.1762065|file:///Users/gusnguyen/stripe/ai|[{"id":"408","sources":["user"]}]
gusnguyen@st-gusnguyen3(laptop) ~/.cursor/plugins/cache/cursor-public/stripe % sqlite3 "/Users/gusnguyen/Library/Application Support/Cursor/User/globalStorage/state.vscdb" \
  "SELECT key, value FROM ItemTable WHERE key LIKE 'cursor.plugins.installedIds%' AND value != '[]';"

cursor.plugins.installedIds.1762065|file:///Users/gusnguyen/stripe/ai|[{"id":"408","sources":["user"]}]
```

pkg/cmd/agent.go
jleong-stripe: I just copied this over from the original POC, I can return all errors here

gusnguyen-stripe (changes requested):

pkg/agentsetup/codex.go
jleong-stripe: I think we can detect Codex incompatibility earlier and do something like this:
Screenshot 2026-07-01 at 6 04 21 PM

pkg/cmd/agent_setup_tui.go
jleong-stripe: hmm I think ideally we do not want to have duplicated skills for the agent (via the plugin) and the project/global dir as it can be confusing right?

pkg/agentsetup/codex.go
gusnguyen-stripe: That'd be great

pkg/cmd/agent_setup_tui.go
gusnguyen-stripe: Yea, I'm torn between allowing user to select both, or stick with our assumption that only Plugin or direct skill install but not both

jleong-stripe: Currently we allow them to select both, but the default is to prefer the plugins (hence the wording above)

pkg/agentsetup/cursor.go
jleong-stripe: yeah it is flakey, and we also don't have a good way to programmatically instead.

I think for the scope of this first PR let's punt on cursor, and just do something like this:

Screenshot 2026-07-01 at 6 29 33 PM

pkg/agentskills/agentskills.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		wrote := false
+		for _, file := range skill.Files {
+			if file == "" {
+				continue
+			}
+			target := filepath.Join(destDir, skill.Name, filepath.FromSlash(file))
+			if !isUnderDir(target, destDir) {
+				continue
+			}
```
jleong-stripe, line 93: check that the target is strictly in the dest directory

gusnguyen-stripe (approved):
LGTM

vzhang-stripe (changes requested):
Image

Does this review make sense?

jleong-stripe:
@vzhang-stripe 
1. In short, Cursor has no plugin CLI — plugins are managed inside the editor via slash commands (e.g.
  /add-plugin stripe). There's no reliable way to check install status from outside Cursor. Hence we default to showing a manual install instruction for Cursor. 
2. I think that's a fair point, the intersection of the `--skills` and `--yes`  flags can be confusing, and given that skills are meant to be a fallback for agent without Stripe plugin support, I'm going to remove the `--skills` flag from this initial PR. 
3. Good catch! Should be fixed now.

vzhang-stripe (approved):
thank you for addressing the comments

PR #1729, opened 2026-07-01
Fix expected/actual argument order in test assertions

vcheung-stripe:
Some of our tests have assertions that put the expected and actual values in the wrong order, which would give misleading messages when the tests would fail. It should be `require.Equal(t, expected, actual)` and not `require.Equal(t, actual, expected)`. This fixes it across the test files that Claude found, there could be more but I did not want to spend too much effort looking.


vzhang-stripe (approved):
nice

PR #1730, opened 2026-07-01
Handle VirusTotal large file uploads exceeding 32 MB limit

vzhang-stripe:
The CLI executable now exceeds VirusTotal's 32 MB direct-upload limit. Request a large-file upload URL when the executable is too big, and improve error handling and quoting throughout the script.

Committed-By-Agent: claude

 ### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

### DIFF
--- a/scripts/upload-zip-to-virustotal.sh
+++ b/scripts/upload-zip-to-virustotal.sh
@@ -22,6 +22,18 @@ if [ -z "$VIRUSTOTAL_API_KEY" ]; then
   exit 1
 fi
 
+VIRUSTOTAL_DIRECT_UPLOAD_LIMIT_BYTES=$((32 * 1024 * 1024))
+VIRUSTOTAL_SCAN_URL="https://www.virustotal.com/vtapi/v2/file/scan"
+VIRUSTOTAL_LARGE_FILE_UPLOAD_URL="https://www.virustotal.com/vtapi/v2/file/scan/upload_url"
+
+getResponseBody() {
+  printf '%s' "$1" | sed -e 's/HTTPSTATUS\:.*//g'
+}
+
+getResponseCode() {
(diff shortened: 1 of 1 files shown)


scripts/upload-zip-to-virustotal.sh
```diff
@@ -22,6 +22,18 @@ if [ -z "$VIRUSTOTAL_API_KEY" ]; then
   exit 1
 fi
 
+VIRUSTOTAL_DIRECT_UPLOAD_LIMIT_BYTES=$((32 * 1024 * 1024))
+VIRUSTOTAL_SCAN_URL="https://www.virustotal.com/vtapi/v2/file/scan"
+VIRUSTOTAL_LARGE_FILE_UPLOAD_URL="https://www.virustotal.com/vtapi/v2/file/scan/upload_url"
```
vcheung-stripe, line 27: Other than allowing larger files, does /vtapi/v2/file/scan/upload_url do anything meaningfully different from /vtapi/v2/file/scan? If not then can we use /vtapi/v2/file/scan/upload_url for all files?

vzhang-stripe: `/vtapi/v2/file/scan` is the actual direct upload-and-scan endpoint.
`/vtapi/v2/file/scan/upload_url` is a helper endpoint that returns a special upload_url; you then do a second POST of the file to that returned URL.

 Other than the upload mechanism, there’s no meaningful documented difference in scan semantics. The resulting upload response is the same kind of thing you already use: sha256, scan_id, permalink, etc.

Iwould not switch to upload_url for all files. VirusTotal documents it specifically as the path for files larger than 32 MB:

` /file/scan:` direct uploads up to 32 MB
`/file/scan/upload_url`: get a special upload URL for files above that limit

vcheung-stripe (approved):

PR #1737, opened 2026-07-03
Add query runs v2 API

kamenrider-cyber:
### Reviewers
r? @
cc @stripe/developer-products

### Summary
Adds a new `stripe reporting query-runs` command for the v2 Data
Reporting Query Runs API (`/v2/data/reporting/query_runs`), letting users run
ad hoc SQL queries against their Stripe data from the CLI.

Previously the CLI could only `GET` Sigma scheduled queries via the older
`/v1/sigma` endpoint; it had no way to kick off new or ad hoc queries. This adds:

- `stripe reporting query-runs create` — `POST /v2/data/reporting/query_runs`.
  Specify custom SQL inline with `--sql`, from a file with `--sql-file <path>`,
  or via stdin with `--sql-file -`. Supports `--compress-file`
  (`result_options.compress_file`) and the standard `--dry-run` / `--live`.
- `stripe reporting query-runs retrieve <id>` — `GET /v2/data/reporting/query_runs/{id}`
  to poll status (`running` → `succeeded`/`failed`) and fetch the result's
  download URL (`result.file.download_url.url`).

The API is in public preview, so the request builder sets the
`Stripe-Version: 2026-06-24.preview` header and JSON content-type automatically
(via `IsPreviewCommand`). Implemented as a manual command (modeled on
`stripe analytics query`) rather than relying on the auto-generated resource
command, to give first-class SQL flags and a stable command path.

The command is registered as a normal top-level command (like `analytics`,
`listen`, etc.) — it is not hidden and not nested under the `stripe preview`
namespace, since the API is public preview rather than private. Note the
`.preview` in the version header is just the API version pin and is independent
of the CLI `preview` command namespace.

Includes unit + HTTP-integration tests covering request body construction, SQL
resolution (inline/file/stdin/non-SQL pass-through/error cases), the outgoing
request shape (method, path, auth, content-type, version header), and error
handling.

### Test plan
- `go build ./...`, `go vet ./pkg/cmd/`, and `go test ./pkg/cmd/` pass.
- Manually verified against a test-mode account:
  - `stripe reporting query-runs create --sql "SELECT * FROM charges LIMIT 10"`
    returns a `qryrun_...` id with `status: running`.
  - `stripe reporting query-runs retrieve <id>` transitions `running` →
    `succeeded`, and `result.file.download_url.url` is populated on success
    (the URL is short-lived and expires ~5 min after retrieval).
  - `--dry-run` shows the expected POST, JSON body, and preview version header.
  
Screenshot 2026-07-06 at 11 51 29 AM
Screenshot 2026-07-06 at 11 51 39 AM

### DIFF
--- a/pkg/cmd/reporting.go
+++ b/pkg/cmd/reporting.go
@@ -1,0 +1,251 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/requests"
+	"github.com/stripe/stripe-cli/pkg/stripe"
+	"github.com/stripe/stripe-cli/pkg/validators"
+)
+
+const queryRunsPath = "/v2/data/reporting/query_runs"
+
+type reportingCmd struct {
+	cmd *cobra.Command
+}
+
+type reportingQueryRunsCmd struct {
+	cmd *cobra.Command
+}
+
+type reportingQueryRunsCreateCmd struct {
+	cmd          *cobra.Command
+	rb           requests.Base
+	sql          string
+	sqlFile      string
+	compressFile bool
+}
+
+type reportingQueryRunsRetrieveCmd struct {
+	cmd *cobra.Command
+	rb  requests.Base
+}
+
+func newReportingCmd() *reportingCmd {
+	rc := &reportingCmd{}
+	rc.cmd = &cobra.Command{
+		Use:   "reporting",
+		Short: "Run Stripe Sigma queries via the Data Reporting API",
+		Long: `Run ad hoc SQL queries against your Stripe data using the Data Reporting API.
+
+Use the query-runs subcommands to kick off a new query and retrieve its
+results. This uses the /v2/data/reporting/query_runs preview API.`,
+		Args: validators.NoArgs,
+		// Hidden while the underlying Query Runs API is in preview. The command
+		// is still runnable; it's just omitted from help output and --map.
+		Hidden: true,
+	}
+
+	rc.cmd.AddCommand(newReportingQueryRunsCmd().cmd)
+	return rc
+}
+
+func newReportingQueryRunsCmd() *reportingQueryRunsCmd {
+	qrc := &reportingQueryRunsCmd{}
+	qrc.cmd = &cobra.Command{
+		Use:   "query-runs",
+		Short: "Create and retrieve QueryRun objects",
+		Long: `Create and retrieve QueryRun objects.
+
+A QueryRun runs a custom SQL query against your Stripe data. Create a query run
+to kick off a query, then retrieve it to poll its status and fetch the download
+URL of the result once the query has completed.`,
+		Args: validators.NoArgs,
+	}
+
+	qrc.cmd.AddCommand(newReportingQueryRunsCreateCmd().cmd)
+	qrc.cmd.AddCommand(newReportingQueryRunsRetrieveCmd().cmd)
+	return qrc
+}
+
+func newReportingQueryRunsCreateCmd() *reportingQueryRunsCreateCmd {
+	cc := &reportingQueryRunsCreateCmd{}
+
+	cc.rb = requests.Base{
+		Method:           http.MethodPost,
+		Profile:          &Config.Profile,
+		IsPreviewCommand: true,
+	}
+
+	cc.cmd = &cobra.Command{
+		Use:   "create",
+		Short: "Create a query run from custom SQL",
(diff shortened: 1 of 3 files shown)


pkg/cmd/reporting.go
```diff
@@ -0,0 +1,251 @@
...
+		Use:   "reporting",
+		Short: "Run Stripe Sigma queries via the Data Reporting API",
+		Long: `Run ad hoc SQL queries against your Stripe data using the Data Reporting API.
+
+Use the query-runs subcommands to kick off a new query and retrieve its
+results. This uses the /v2/data/reporting/query_runs preview API.`,
+		Args: validators.NoArgs,
+		// Hidden while the underlying Query Runs API is in preview. The command
+		// is still runnable; it's just omitted from help output and --map.
+		Hidden: true,
```
benanthony-stripe: I'm not familiar with how CLI visibility works—the API is in Public Preview, which means any merchant can access it. Should this still be hidden?

pkg/cmd/reporting_test.go
```diff
@@ -0,0 +1,256 @@
...
+	cc := newReportingQueryRunsCreateCmd()
+	cc.sql = "SELECT 1"
+	cc.sqlFile = "query.sql"
+
+	_, err := cc.resolveSQL(cc.cmd)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "mutually exclusive")
+}
+
+func TestResolveSQL_EmptyFile(t *testing.T) {
```
benanthony-stripe, line 91: Should we also have a test for malformed sql or non-sql in the file?

pkg/cmd/root.go
```diff
@@ -268,6 +268,12 @@ func init() {
 	rootCmd.AddCommand(newPluginCmd().cmd)
 	resources.AddAllResourcesCmds(rootCmd, &Config)
 	registerHTTPCmds(rootCmd)
+
+	// Data Reporting query runs is a preview API, so nest it under `stripe
+	// preview` to signal it's not GA. The command itself is also hidden.
+	if preview, ok := cmdutil.FindSubCmd(rootCmd, "preview"); ok {
```
benanthony-stripe: Sorry for all the dumb questions, never interacted with CLI commands before, is this idiomatic? It feels odd that we're the only ones doing this preview pattern in this whole file

pkg/cmd/reporting.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	body := cc.buildRequestBody(sql)
+
+	if cc.rb.DryRun {
+		output, err := cc.rb.BuildDryRunOutput(apiKey, cc.rb.APIBaseURL, queryRunsPath, &requests.RequestParameters{}, body)
+		if err != nil {
+			return err
+		}
+		b, _ := json.MarshalIndent(output, "", "  ")
```
benanthony-stripe: Is it safe to assume we can ignore `err` here?

pkg/cmd/reporting.go
kamenrider-cyber: sorry, I saw they are preview and assume they are private preview...

pkg/cmd/root.go
kamenrider-cyber: sorry, it wasn't idiomatic and since it is not private preview I will removed it.

pkg/cmd/reporting.go
kamenrider-cyber: Sorry, I add the error handling there.

benanthony-stripe (approved):
LGTM, great work! First PR in the books!!!!!!!!!!!!

pkg/cmd/reporting.go
```diff
@@ -0,0 +1,251 @@
...
+	rc.cmd.Flags().BoolVarP(&rc.rb.Livemode, "live", "", false, "Make a live request (default: test)")
+	rc.cmd.Flags().BoolVarP(&rc.rb.DarkStyle, "dark-style", "", false, "Use a darker color scheme better suited for lighter command-lines")
+
+	rc.cmd.Flags().StringVar(&rc.rb.APIBaseURL, "api-base", stripe.DefaultAPIBaseURL, "Sets the API base URL")
+	rc.cmd.Flags().MarkHidden("api-base") // #nosec G104
+
+	return rc
+}
+
+func (cc *reportingQueryRunsCreateCmd) runReportingQueryRunsCreateCmd(cmd *cobra.Command, args []string) error {
```
vzhang-stripe, line 160: I think we need to do stripe.ValidateAPIBaseURL like [here](https://stripe.sourcegraphcloud.com/r/stripe/stripe-cli/-/blob/pkg/requests/base.go?L144) so no API key will be leaked

vzhang-stripe (changes requested):

vzhang-stripe (approved):
nice

PR #1739, opened 2026-07-03
coop: validate blueprint node references

tomelm:
## Summary
- Validate `${node.<step>.<node>...}` references when a blueprint is loaded (`LoadBlueprint`), so malformed or unknown node references fail fast instead of surfacing as confusing errors mid-session.
- Collect valid `step.node` (and `step.node.request`) keys, then check every `${node...}` reference in the marshaled blueprint against them.
- Allow integer-indexed reference segments (e.g. an indexed test-request) as valid.

## Tests
- `TestValidateBlueprintReferences` — covers valid references, malformed references, and unknown referenced nodes.
- `go test ./pkg/coop/... ./pkg/cmd/coop/...` — all green.

## Base
Based directly on `coop`. First of the co-op product-quality follow-ups; `agent-quality` and `blueprint-schema` build on top and will be opened as `coop` advances.

### DIFF
--- a/pkg/coop/blueprint.go
+++ b/pkg/coop/blueprint.go
@@ -4,12 +4,15 @@ import (
 	"embed"
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 )
 
 //go:embed blueprints/*.json
 var blueprintFS embed.FS
+
+var blueprintNodeRefRE = regexp.MustCompile(`\$\{node\.([^}:]+):[^}]+\}`)
 
 // BlueprintStep is a step definition within a blueprint.
 type BlueprintStep struct {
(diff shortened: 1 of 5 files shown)


pkg/coop/blueprint.go
```diff
@@ -4,13 +4,16 @@ import (
...
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 )
 
 //go:embed blueprints/*.json
 var blueprintFS embed.FS
 
+var blueprintNodeRefRE = regexp.MustCompile(`\$\{node\.([^}:]+):[^}]+\}`)
```
kwu-stripe: This only validates placeholders that already match the full ${node.<ref>:<field?} shape: it wouldn't catch something like ${node.setup.create-product}.

kwu-stripe (approved):

kwu-stripe (approved):

PR #1745, opened 2026-07-06
Send telemetry on unrecognized commands

keshavc-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

https://jira.corp.stripe.com/browse/DEVAI-1146

### Testing
https://logscale.corp.stripe.com/main-view/search?live=false&query=%22ARECIBO-EVENT%22%0A%7C%20%22context.p_client_id%22%20%3D%20%22stripe-cli%22%0A%7C%20groupBy(%22context.p_event_name%22)&start=2h&tz=UTC

```
keshavc@st-keshavc2(laptop) ~/stripe/stripe-cli % ./bin/stripe foobar
Unknown command "foobar" for "stripe".
See "stripe --help" for a list of available commands.
keshavc@st-keshavc2(laptop) ~/stripe/stripe-cli % cat ~/.config/stripe/unknown_commands.json
cat: /Users/keshavc/.config/stripe/unknown_commands.json: No such file or directory
keshavc@st-keshavc2(laptop) ~/stripe/stripe-cli % CLAUDECODE=1 ./bin/stripe foobar
<claude-code-hint v="1" type="plugin" value="stripe@claude-plugins-official" />
Unknown command "foobar" for "stripe".
See "stripe --help" for a list of available commands.
keshavc@st-keshavc2(laptop) ~/stripe/stripe-cli % cat ~/.config/stripe/unknown_commands.json
[{"command":"foobar","timestamp":"2026-07-06T22:06:12Z","agent":"claude_code"}]% 
```

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -177,6 +177,7 @@ func Execute(ctx context.Context) {
 
 		case strings.Contains(errString, "unknown command"):
 			showSuggestion()
+			recordUnknownCommand(updatedCtx, strings.Join(os.Args[1:], " "))
 
 		default:
 			fmt.Fprintln(os.Stderr, err)
--- a/pkg/cmd/unknown_cmd_telemetry.go
+++ b/pkg/cmd/unknown_cmd_telemetry.go
@@ -1,0 +1,100 @@
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/stripe/stripe-cli/pkg/stripe"
+	"github.com/stripe/stripe-cli/pkg/useragent"
+)
+
+const (
+	unknownCmdBatchFile  = "unknown_commands.json"
+	unknownCmdBatchSize  = 10
+	unknownCmdEventName  = "Unknown Command Attempted"
+)
+
+type unknownCommandEntry struct {
+	Command   string `json:"command"`
+	Timestamp string `json:"timestamp"`
+	Agent     string `json:"agent"`
+}
+
+// recordUnknownCommand records an unknown command attempt in agent environments.
+// It batches entries locally and sends telemetry every unknownCmdBatchSize occurrences.
+func recordUnknownCommand(ctx context.Context, command string) {
+	agent := useragent.DetectAIAgent(os.Getenv)
+	if agent == "" {
+		return
+	}
+
+	telemetryClient := stripe.GetTelemetryClient(ctx)
+	if telemetryClient == nil {
+		return
+	}
+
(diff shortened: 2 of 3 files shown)


pkg/cmd/root.go
```diff
@@ -177,6 +177,7 @@ func Execute(ctx context.Context) {
 
 		case strings.Contains(errString, "unknown command"):
 			showSuggestion()
+			recordUnknownCommand(updatedCtx, strings.Join(os.Args[1:], " "))
```
keshavc-stripe, line 180: Note that this is synchronous but the timeout is 3 seconds. Making this async doesn't work since the os.Exit(1) kills all subroutines. https://github.com/stripe/stripe-cli/blob/7c98c546afddeed2f5e1a39e318067ac21771f09/cmd/stripe/main.go#L22-L24

pkg/cmd/unknown_cmd_telemetry.go
```diff
@@ -0,0 +1,100 @@
...
+	log "github.com/sirupsen/logrus"
+
+	"github.com/stripe/stripe-cli/pkg/stripe"
+	"github.com/stripe/stripe-cli/pkg/useragent"
+)
+
+const (
+	unknownCmdBatchFile = "unknown_commands.json"
+	unknownCmdBatchSize = 10
+	unknownCmdEventName = "Unknown Command Attempted"
```
matv-stripe, line 19: is this a valid event? do we have to update anything on the spark side internally? @vcheung-stripe would know

keshavc-stripe: Hmm showed up on logscale but might be missing some more context for spark stuff

vcheung-stripe: This is fine, whatever you put here is what you'll query for. Whitespace/capitalization is all fine too. If it's in logscale then you should be good.

pkg/cmd/unknown_cmd_telemetry.go
```diff
@@ -0,0 +1,100 @@
...
+
+	entries = append(entries, unknownCommandEntry{
+		Command:   command,
+		Timestamp: time.Now().UTC().Format(time.RFC3339),
+		Agent:     agent,
+	})
+
+	if len(entries) >= unknownCmdBatchSize {
+		sendUnknownCommandBatch(ctx, telemetryClient, entries)
+		_ = os.Remove(batchPath)
```
vcheung-stripe, line 54: If the Remove() fails, then we might send the same events more than once right? Is there any concern about that or should we have a way to dedup them?

vcheung-stripe (approved):

pkg/cmd/unknown_cmd_telemetry.go
keshavc-stripe: True, but I don't think it'll happen that often and don't think the impact is that high if it does happen

PR #1747, opened 2026-07-07
Add per-version PagerDuty alerting to plugin canary tests

yahanxing-stripe:
### Summary

- Parameterize `scripts/notify.sh` to accept dedup key, summary, resolve summary, and severity as required env vars
- Update `install-test.yml` to pass its values explicitly
- Wire PagerDuty alerting into `plugin-canary.yml`: triggers an alert for each failing plugin version (dedup key: plugin-canary-projects-<version>) and auto-resolves when it passes again
- Add dryrun input to plugin-canary


vzhang-stripe (approved):
nicee

PR #1749, opened 2026-07-07
[agentskills] Add skills freshness checking

gusnguyen-stripe:
### Reviewers
r? @jleong-stripe  
cc @stripe/developer-products

### Summary

Adds `agentskills.Check()` to compare installed Stripe skills against the remote skills index and report freshness per destination directory.

### What it does

**Freshness checking** — compares local `.agents/skills` installs to `https://docs.stripe.com/.well-known/skills/index.json`:
- Fetches remote skill file content and compares SHA256 hashes with local copies
- Returns per-skill status: `not_installed`, `current`, `out_of_date`, or `error`
- Aggregates directory-level status with installed/out-of-date counts

**Performance** — parallelizes remote fetches with a concurrency cap to avoid overwhelming docs.stripe.com.

**Tests** — covers current, out-of-date, missing, and error cases.

### Stack

This is PR 1 of 4 in a stack for skills status in `stripe agent setup`. Later PRs wire this into CLI flows.

### DIFF
--- a/pkg/agentskills/agentskills.go
+++ b/pkg/agentskills/agentskills.go
@@ -5,7 +5,9 @@
 package agentskills
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -14,6 +16,8 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+
+	"golang.org/x/sync/errgroup"
 )
 
 // IndexURL is the canonical Stripe skills index. It is a var (not a const) so
@@ -21,7 +25,11 @@ import (
 // from it, so overriding it redirects file fetches too.
 var IndexURL = "https://docs.stripe.com/.well-known/skills/index.json"
 
-const requestTimeout = 10 * time.Second
+const (
+	requestTimeout         = 10 * time.Second
+	skillCheckConcurrency  = 8
+	remoteFetchConcurrency = 8
+)
 
 // Index is the top-level response from IndexURL.
 type Index struct {
@@ -34,6 +42,32 @@ type Skill struct {
 	Name        string   `json:"name"`
 	Description string   `json:"description"`
 	Files       []string `json:"files"`
+}
+
+const (
+	StatusNotInstalled = "not_installed"
+	StatusCurrent      = "current"
+	StatusOutOfDate    = "out_of_date"
+	StatusPartial      = "partial"
+	StatusError        = "error"
+)
+
+// SkillCheck is the per-skill result of comparing a local install to the index.
+type SkillCheck struct {
+	Name         string   `json:"name"`
+	Status       string   `json:"status"`
+	MissingFiles []string `json:"missing_files,omitempty"`
+	ChangedFiles []string `json:"changed_files,omitempty"`
+}
+
+// DirStatus summarizes installed skills under a single destination directory.
+type DirStatus struct {
+	Dir            string       `json:"dir"`
+	Status         string       `json:"status"`
+	Skills         []SkillCheck `json:"skills,omitempty"`
+	InstalledCount int          `json:"installed_count"`
+	OutOfDateCount int          `json:"out_of_date_count"`
+	Error          string       `json:"error,omitempty"`
 }
 
 func clientOrDefault(httpClient *http.Client) *http.Client {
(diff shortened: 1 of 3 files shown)


pkg/agentskills/agentskills.go
```diff
@@ -64,6 +92,112 @@ func FetchIndex(ctx context.Context, httpClient *http.Client) (*Index, error) {
...
+			result.OutOfDateCount++
+		}
+		result.Skills = append(result.Skills, check)
+	}
+
+	result.Status = aggregateDirStatus(result)
+	return result, nil
+}
+
+func checkSkill(ctx context.Context, client *http.Client, destDir, base string, skill Skill) SkillCheck {
```
jleong-stripe: how long does this take?

pkg/cmd/agent.go
```diff
@@ -350,33 +411,108 @@ func (asc *agentSetupCmd) installSkills(ctx context.Context, out io.Writer, scop
...
+	}
+
+	local, err := asc.skillsCheck(ctx, localDir)
+	if err != nil {
+		return skillsScopes{}, err
+	}
+	global, err := asc.skillsCheck(ctx, globalDir)
+	if err != nil {
+		return skillsScopes{}, err
+	}
```
jleong-stripe: should we fail close here?

jleong-stripe: also should we parallelize both calls ?

pkg/cmd/agent.go
```diff
@@ -145,24 +161,42 @@ func (asc *agentSetupCmd) runSetup(cmd *cobra.Command, _ []string) error {
...
-
 	if asc.statusOnly {
-		if len(detected) == 0 {
+		skills, err := asc.loadSkillsScopes(ctx)
+		if err != nil {
+			return err
+		}
+		if len(detected) > 0 {
+			printStatusTable(out, detected)
+			fmt.Fprintln(out)
```
jleong-stripe: Should we display the skills in `--status`? Assuming we want to prioritize plugin install, showing skills status may be misleading and prompt the user to install both. Maybe we should only show if the skills already exist?

jleong-stripe (commented):
left some questions, I tested it locally and it works well!

pkg/agentskills/agentskills.go
gusnguyen-stripe: It depends on which skill, but longest is about ~720ms for Stripe's wifi
image

pkg/cmd/agent.go
gusnguyen-stripe: Not at this level, but it's worth to parallelize when fetching index and skills in a follow up. And making this fail close!

pkg/cmd/agent.go
gusnguyen-stripe: Good call, if this is called by an agent we should not be showing both and prioritize plugins. However, when the user has Claude + stale skills, should we show skills in status?

pkg/agentskills/agentskills.go
gusnguyen-stripe: I think right now it takes about 1s-2s and we only have a handful number of skills. do you want to address it now or in a follow up PR?

pkg/cmd/agent.go
jleong-stripe: yeah I think if the skills exist then we show it, but if not we don't have to show `not installed`

pkg/agentskills/agentskills.go
jleong-stripe: How do you plan to address this?

gusnguyen-stripe: Just added a few commits to address the sequential HTTP calls bottleneck. I add goroutines to parallelize network calls: inside `checkSkill` and `Check`. Concurrency capped to 8 to avoid hammering the CDN. Attached pic for improvements 
image

jleong-stripe (approved):

pkg/agentskills/agentskills.go
```diff
@@ -64,6 +97,186 @@ func FetchIndex(ctx context.Context, httpClient *http.Client) (*Index, error) {
...
+				return nil
+			}
+
+			outcome := fileOutcome{file: job.file, hasLocal: true}
+			remoteContent, err := remote.get(ctx, base+skill.Name+"/"+job.file)
+			if err != nil {
+				outcome.changed = true
+				outcomes[i] = outcome
+				return nil
+			}
```
vcheung-stripe, line 234: If the request fails, does it make sense to have `outcome.changed = true` or do we need a way to fail gracefully?

pkg/agentskills/agentskills.go
```diff
@@ -64,6 +97,186 @@ func FetchIndex(ctx context.Context, httpClient *http.Client) (*Index, error) {
...
+			if !bytes.Equal(hashBytes(local), hashBytes(remoteContent)) {
+				outcome.changed = true
+			}
+			outcomes[i] = outcome
+			return nil
+		})
+	}
+	_ = g.Wait()
```
vcheung-stripe, line 245: nit: seems like we're ignoring the error on purpose so that we can process it in `fileOutcome`, could we add a comment to explain that to the reader?

vcheung-stripe (approved):

pkg/agentskills/agentskills.go
gusnguyen-stripe: Added some comments to clarify

pkg/agentskills/agentskills.go
gusnguyen-stripe: Add some comments to clarify

PR #1752, opened 2026-07-07
fix: only alert PagerDuty when install-test.sh itself fails

vcheung-stripe:
Update the package manager test so that it doesn't alert when setup steps fail, only when the actual tests themselves fail.

Here's what that looks like:
| Condition | PagerDuty action |
| ------------- | ------------- |
| Test fails | trigger alert |
| Test succeeds  | resolve alert  |
| Test setup fails  | do nothing, active alerts stay active, resolved alerts stay resolved  |

Testing:
- All passing: https://github.com/stripe/stripe-cli/actions/runs/28974869520
- Simulated setup failure: https://github.com/stripe/stripe-cli/actions/runs/28974649224
- Simulated actual failure: https://github.com/stripe/stripe-cli/actions/runs/28974761538


yahanxing-stripe (approved):
nice!

PR #1753, opened 2026-07-07
[sandboxes cli] Add hidden `stripe sandbox new` command scaffold

shashwat-stripe:
## Summary

Scaffolds a new, hidden `stripe sandbox new` subcommand as the entry point for creating a sandbox for the logged-in account from the CLI.

This is scaffolding only. The command is registered (hidden from help and completion while it is experimental), declares its flags, and prints a placeholder. The intended behavior is written out as plain steps in the command body so the follow-up PR can implement it directly; this PR intentionally ships no request logic.

It also assigns the sandbox commands and the `pkg/sandbox` client package to the sandboxes team in `CODEOWNERS`.

## What's included

- `pkg/cmd/sandbox.go`: a hidden `stripe sandbox new` subcommand (`--name`, `--replica-of`, `--stripe-context`, and a hidden `--api-base`), a placeholder handler, and a documented step-by-step plan for the implementation.
- `CODEOWNERS`: `/pkg/cmd/sandbox*.go` and `/pkg/sandbox/` assigned to `@stripe/sandboxes-engineering`.

## Test plan

- `go build ./pkg/...` passes.
- `go vet ./pkg/cmd/` passes.
- `go test ./pkg/cmd/ -run Sandbox` passes (existing sandbox tests unaffected).
- The command is hidden: it does not appear in `stripe sandbox --help`, and invoking it prints the placeholder line.

## Rollout / revert

Safe to revert. The command is hidden and does nothing beyond printing a placeholder.

### DIFF
--- a/pkg/cmd/sandbox.go
+++ b/pkg/cmd/sandbox.go
@@ -59,8 +59,10 @@ func newSandboxCmd() *sandboxCmd {
 
 	createCmd := newSandboxCreateCmd()
 	claimCmd := newSandboxClaimCmd()
+	newCmd := newSandboxNewCmd()
 	sc.cmd.AddCommand(createCmd.cmd)
 	sc.cmd.AddCommand(claimCmd.cmd)
+	sc.cmd.AddCommand(newCmd.cmd)
 	return sc
 }
 
@@ -400,6 +402,14 @@ type sandboxClaimCmd struct {
 	nonInteractive bool
 }
 
+type sandboxNewCmd struct {
+	cmd           *cobra.Command
+	name          string
+	replicaOf     string
+	stripeContext string
+	apiBase       string
+}
+
 func newSandboxClaimCmd() *sandboxClaimCmd {
 	scc := &sandboxClaimCmd{}
 	scc.cmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


pkg/cmd/sandbox.go
```diff
@@ -449,6 +459,58 @@ func (scc *sandboxClaimCmd) runSandboxClaimCmd(cmd *cobra.Command, args []string
...
+		// Remove this when the command is ready to GA.
+		Hidden: true,
+	}
+
+	snc.cmd.Flags().StringVar(&snc.name, "name", "", "Name for the new sandbox")
+	snc.cmd.Flags().StringVar(&snc.replicaOf, "replica-of", "", "Livemode workspace ID to replicate (wksp_...)")
+	snc.cmd.Flags().StringVar(&snc.stripeContext, "stripe-context", "", "Stripe-Context compartment to scope the request to")
+	snc.cmd.Flags().StringVar(&snc.apiBase, "api-base", stripe.DefaultAPIBaseURL, "Sets the Stripe API base URL")
+	_ = snc.cmd.Flags().MarkHidden("api-base")
+
```
echeng-stripe, line 484: can we add the flag for number of sandboxes that need to create here? The default value is `1`.

pkg/cmd/sandbox.go
```diff
@@ -449,6 +459,58 @@ func (scc *sandboxClaimCmd) runSandboxClaimCmd(cmd *cobra.Command, args []string
...
+		Args: validators.NoArgs,
+		RunE: snc.runSandboxNewCmd,
+		// Hidden while this is an experimental POC: keep it out of help/completion.
+		// The real access gate is the backend (the UAT flag + hzn_sandbox_create).
+		// Remove this when the command is ready to GA.
+		Hidden: true,
+	}
+
+	snc.cmd.Flags().StringVar(&snc.name, "name", "", "Name for the new sandbox")
+	snc.cmd.Flags().StringVar(&snc.replicaOf, "replica-of", "", "Livemode workspace ID to replicate (wksp_...)")
```
echeng-stripe, line 480: The `replicaOf` fits better for a boolean type outside of organization, right?

pkg/cmd/sandbox.go
shashwat-stripe: i would rather have batch be a different command (since it would require a different shape)

pkg/cmd/sandbox.go
shashwat-stripe: Right now, this command has no compartment context when being run, thus needs this to be a string

PR #1756, opened 2026-07-08
[sandboxes cli] Implement stripe sandbox new: v2 CreateSandbox (copy + blank)

shashwat-stripe:
Business logic for the hidden `stripe sandbox new` command. Stacked on #1753 (the scaffold).

Wires the command to POST /v2/sandboxes using the logged-in UAT (STRIPE-V2-SIG). Two create modes (a oneof; exactly one required):
- copy: `--replica-of <wksp_>` clones a livemode workspace
- blank: `--business-location <country>` creates a fresh sandbox

Stripe-Context must be a playground (`play_`), validated client-side since the backend rejects a non-playground context. `--activate` toggles capability activation (default true). `--batch` is optional (default 1; >1 rejected today, with a future-shape comment).

Command stays Hidden and is registered only behind `STRIPE_CLI_ENABLE_SANDBOX_NEW`; the authoritative access gate is server-side (UAT allowlist + hzn_sandbox_create).

Stacking note: based on the scaffold branch for a clean logic-only diff; will retarget to `sandboxes-cli` once #1753 merges. Part of the long-lived `sandboxes-cli` feature branch, not for merge to master until the feature is release-ready.

### DIFF
--- a/pkg/cmd/sandbox.go
+++ b/pkg/cmd/sandbox.go
@@ -68,21 +68,13 @@ func newSandboxCmd() *sandboxCmd {
 	sc.cmd.AddCommand(createCmd.cmd)
 	sc.cmd.AddCommand(claimCmd.cmd)
 
-	// `sandbox new` is an experimental POC. It is gated behind an env flag so it
-	// is not registered at all (invisible AND unrunnable — `unknown command`)
-	// unless explicitly enabled, on top of being Hidden. This keeps it out of
-	// users' hands pre-GA; the authoritative access gate remains server-side
-	// (the UAT allowlist + hzn_sandbox_create), not the client.
-	if sandboxNewEnabled() {
-		sc.cmd.AddCommand(newSandboxNewCmd().cmd)
-	}
+	// `sandbox new` is an experimental POC. It stays Hidden (kept out of
+	// help/completion) but is always registered — this ships only to the
+	// long-lived sandboxes-cli feature branch, never to master pre-GA, so no
+	// client-side env gate is needed. The authoritative access gate remains
+	// server-side (the UAT allowlist + hzn_sandbox_create), not the client.
+	sc.cmd.AddCommand(newSandboxNewCmd().cmd)
 	return sc
-}
-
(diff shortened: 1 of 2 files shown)


pkg/cmd/sandbox.go
```diff
@@ -477,40 +500,153 @@ that you have previously logged in with your Stripe account credentials.`,
...
+		fmt.Fprintln(cmd.OutOrStdout(), string(respBytes))
+	}
+
 	return nil
 }
 
+// newIdempotencyToken builds a stable-per-invocation idempotency token from the
+// sandbox name and a random UUID so retried requests within a run don't create
+// duplicate sandboxes.
+func newIdempotencyToken(name string) string {
```
echeng-stripe: The comment says retries within a run won't create duplicates, but uuid.NewString() produces a fresh token every invocation. So re-running the same CLI command (the most common retry scenario) will create a duplicate. 
Either reword the comment to scope the guarantee to HTTP-layer retries within a single process, or derive the token deterministically from the inputs so repeated invocations actually dedup.

pkg/cmd/sandbox.go
```diff
@@ -477,40 +500,153 @@ that you have previously logged in with your Stripe account credentials.`,
...
+	}
+	uatBytes, err := config.KeyRing.Get(config.UATKeychainItemKey)
+	if err != nil || len(uatBytes) == 0 {
+		return fmt.Errorf("no user access token found; run `stripe login` first")
+	}
+	uat := strings.TrimSpace(string(uatBytes))
+
+	if strings.TrimSpace(snc.name) == "" {
+		return fmt.Errorf("--name is required")
+	}
```
echeng-stripe: `--name` should be a required flag, not a runtime check?

shashwat-stripe: fixed!

pkg/cmd/sandbox.go
shashwat-stripe: addressed to match dashboard idempotency

echeng-stripe (approved):

PR #1757, opened 2026-07-08
[stripe agent setup] Surface skills status in setup flows

gusnguyen-stripe:
### Reviewers
r? @jleong-stripe  
cc @stripe/developer-products

### Summary

Wires `agentskills.Check()` into `stripe agent setup` so skills status appears in `--status`, `--json`, and the interactive TUI, and supports updating already-installed skills.

### What it does

**Skills status** — adds local and global skills checks for flows that display skills:
- `--status` shows installation state (but hide the other one to discourage installing in both)
image

- `--json` includes skills status and planned skills actions
- Interactive TUI shows skills installation status
image

**Skills update flow** — uses freshness checks before installing:
- Skips installation when skills are already current unless `--force` is passed
- Returns install/update counts from `installSkills` for JSON output

**Interactive TUI** — improves skills presentation for humans at a terminal:
- Pre-selects the skills row when outdated skills are detected
image

- Defaults the scope prompt toward the directory that needs install/update

image

- Labels local/global scope choices with their resolved paths and current status
image
image

### Stack

PR 2 of 4. Depends on #1749.

---
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pkg/cmd/agent_setup_tui.go
+++ b/pkg/cmd/agent_setup_tui.go
@@ -279,8 +279,6 @@ func scopeStatusHint(status agentskills.DirStatus) string {
 		return "(up to date)"
 	case agentskills.StatusOutOfDate:
 		return "(out of date)"
-	case agentskills.StatusPartial:
-		return "(partially installed)"
 	case agentskills.StatusNotInstalled:
 		return "(not installed)"
 	default:
--- a/pkg/cmd/agent_setup_tui.go
+++ b/pkg/cmd/agent_setup_tui.go
@@ -3,11 +3,13 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	tea "charm.land/bubbletea/v2"
 	"charm.land/lipgloss/v2"
 
 	"github.com/stripe/stripe-cli/pkg/agentsetup"
+	"github.com/stripe/stripe-cli/pkg/agentskills"
 )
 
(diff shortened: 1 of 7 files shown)


jleong-stripe (approved):

pkg/cmd/agent_setup_tui.go
```diff
@@ -210,24 +229,77 @@ func RunSelectionTUI(statuses []agentsetup.Status) (*Selection, error) {
...
+	}
+	return fmt.Sprintf("%-14s %s", prefix, path)
+}
+
+func scopeStatusHint(status agentskills.DirStatus) string {
+	switch status.Status {
+	case agentskills.StatusCurrent:
+		return "(up to date)"
+	case agentskills.StatusOutOfDate:
+		return "(out of date)"
```
vcheung-stripe, line 281: Small suggestion: might be nice to have more positive language like "new version available" but it's clear enough either way

vcheung-stripe (approved):

pkg/cmd/agent_setup_tui.go
gusnguyen-stripe: "up to date" and "out of date" are better pairing so i'm going to keep them as-is

PR #1759, opened 2026-07-08
[stripe agent setup] Polish skills status display

gusnguyen-stripe:
### Reviewers
r? @jleong-stripe  
cc @stripe/developer-products

### Summary

Polishes skills status presentation in `stripe agent setup` after the core integration lands: simpler status lines, scope-aware output, and smarter install suggestions.

### What it does

**Simpler status lines** — the `--status` skill rows show a plain `installed` / `outdated` state (dropping the per-count text), and the setup TUI skills row is tidied up.

**Scope-aware display** — when one scope is installed and the other is not, show only the installed scope instead of redundant "not installed" lines. This rule is shared between the `--status` text and `--json` output so they stay consistent.

**Install suggestions** — do not suggest a skills install command when the skill is already installed in one of the scopes.

### Stack

PR 4 of 4. Depends on #1758.

### Demo
**With no supported agents**

https://github.com/user-attachments/assets/0b977371-207d-42a8-9ecd-88bf70435e16

**With a supported agent**

https://github.com/user-attachments/assets/ebbcf504-b588-4e26-98c5-8b991317d336

---
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pkg/cmd/agent.go
+++ b/pkg/cmd/agent.go
@@ -79,8 +79,8 @@ type agentSetupJSON struct {
 }
 
 type skillsScopesJSON struct {
-	Local  agentskills.DirStatus `json:"local"`
-	Global agentskills.DirStatus `json:"global"`
+	Local  *agentskills.DirStatus `json:"local,omitempty"`
+	Global *agentskills.DirStatus `json:"global,omitempty"`
 }
 
 // skillsScopes holds the check result for local and global skill directories.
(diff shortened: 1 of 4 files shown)


jleong-stripe (approved):

pkg/cmd/agent.go
```diff
@@ -751,36 +767,43 @@ func printSkillsStatusTable(w io.Writer, skills skillsScopes, allowInstall bool)
...
-	total := len(d.Skills)
 	switch d.Status {
 	case agentskills.StatusError:
 		return color.Red("✗").String(), color.Red("error: " + d.Error).String()
 	case agentskills.StatusCurrent:
-		return color.Green("✔").String(), fmt.Sprintf("%d skills up to date", total)
+		return color.Green("✔").String(), "installed"
 	case agentskills.StatusOutOfDate:
-		return color.Yellow("⚠").String(), fmt.Sprintf("%d of %d skills out of date", d.OutOfDateCount, total)
+		return color.Yellow("⚠").String(), "outdated"
```
vcheung-stripe, line 804: Similar to the last one: "new version available" feels a little nicer to me

vcheung-stripe (approved):

pkg/cmd/agent.go
gusnguyen-stripe: I'm going to keep "outdated" because it's a stronger word for agents than "new version available"

PR #1768, opened 2026-07-09
autoupdate 2/3: Add version check and update marker

yahanxing-stripe:
## Summary

- After each command completes, check GitHub for a newer CLI version (rate-limited to once per 24h via timestamp file)
- If a minor/patch update is available, write a marker file to `~/.stripe/state/update-available` with version, download URL, and checksum
- **Skip major version changes** entirely — no prompt, no auto-update to avoid silently introducing breaking changes 
- Includes SHA256 checksum verification utility used by the apply step

This is **part 2/3** of the auto-update feature for curl-installed binaries.
Stack: #1767 (config) → **this** → #1769 (apply)

## Test plan

- [x] Unit tests for major version detection logic
- [x] Unit tests for marker file read/write/clear
- [x] Unit test for `recordLastCheck` timestamp
- [x] Unit test for checksum verification

### DIFF
--- a/pkg/autoupdate/checker.go
+++ b/pkg/autoupdate/checker.go
@@ -15,10 +15,13 @@ import (
 	"time"
 
 	"github.com/google/go-github/v72/github"
+	goversion "github.com/hashicorp/go-version"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/stripe/stripe-cli/pkg/version"
 )
+
+const httpTimeout = 10 * time.Second
 
 const checkInterval = 24 * time.Hour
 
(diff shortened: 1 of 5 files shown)


pkg/autoupdate/checker.go
```diff
@@ -0,0 +1,293 @@
...
+	return currentMajor != "" && latestMajor != "" && currentMajor != latestMajor
+}
+
+func majorVersion(v string) string {
+	parts := strings.SplitN(v, ".", 2)
+	if len(parts) == 0 {
+		return ""
+	}
+	return parts[0]
+}
```
vcheung-stripe, line 85: I think there's a semver library that we use in other places, could we use that here? Just to avoid having to maintain this or have to handle [non-semver values](https://git.corp.stripe.com/stripe-internal/mint/pull/2318444)

pkg/autoupdate/checker.go
```diff
@@ -0,0 +1,293 @@
...
+	if err != nil {
+		return true
+	}
+
+	return time.Since(time.Unix(ts, 0)) >= checkInterval
+}
+
+func fetchLatestRelease() (ver string, downloadURL string, checksum string) {
+	client := github.NewClient(nil)
+	release, _, err := client.Repositories.GetLatestRelease(context.Background(), "stripe", "stripe-cli")
```
vcheung-stripe: Should we time out after some amount of time?

pkg/autoupdate/checker.go
```diff
@@ -0,0 +1,293 @@
...
+
+	if checksumURL != "" {
+		checksum = fetchChecksumForAsset(checksumURL, assetName)
+	}
+
+	return ver, binaryURL, checksum
+}
+
+func fetchChecksumForAsset(checksumURL, assetName string) string {
+	resp, err := http.Get(checksumURL) //nolint:gosec
```
vcheung-stripe: Same question about timeout here

vcheung-stripe (approved):

PR #1769, opened 2026-07-09
autoupdate 3/3: Apply pending update and re-exec

yahanxing-stripe:
## Summary

This PR adds a step to command execution that automatically upgrades the CLI when a newer version is available.

- On startup, if a marker file exists, download the new binary, verify checksum, atomically replace the current binary, and re-exec the process
- User sees:
  ```
  Automatically updating Stripe CLI from 1.23.0 to 1.23.1.
  To disable auto-update, set STRIPE_NO_AUTO_UPDATE=1 or add auto_update = false to ~/.config/stripe/config.toml
  Updated successfully ✓
  ```
- If download or checksum fails: prints a one-line warning, clears marker, continues with current version
- Binary is never left in a broken state (download to temp → atomic rename)
- Supports both tar.gz (macOS/Linux) and zip (Windows) archives

This is **part 3/3** of the auto-update feature for curl-installed binaries.
Stack: #1767 (config) → #1768 (checker) → **this**

## Test plan
- [x] Automated test: added `pkg/autoupdate/updater_test.go`
- [x] Manual test: build with an old version string, place in `~/.stripe/bin/`, run a command, verify marker is written
- [x] Manual test: with marker present, run a command, verify stripe is updated to latest version
Screenshot 2026-07-15 at 3 54 07 PM


vcheung-stripe (approved):
Nice! The only thing I would suggest is making sure this works on Windows, it handles things a little differently. It's probably worth using github actions to do that.

yahanxing-stripe:
> Nice! The only thing I would suggest is making sure this works on Windows, it handles things a little differently. It's probably worth using github actions to do that.

yes once I merge the window install script, i'll update this to handle auto upgrade for windows

PR #1772, opened 2026-07-10
Fix co-op tmux launch with custom config dirs

tomer-stripe:
## Summary

While doing post-merge validation of `coop`, the outside-tmux `coop start --debug-agent` path failed when `XDG_CONFIG_HOME` was set. The debug-agent pane inherited the custom config dir, but the TUI `coop join` command in the newly-created tmux session did not, so it looked in the default config dir and could not find the session.

This updates tmux command construction so both the TUI pane and agent/debug-agent pane preserve the co-op config environment.

## Validation

- `env -u CODEX_CI -u CODEX_THREAD_ID -u CODEX_SANDBOX -u CODEX_SANDBOX_NETWORK_DISABLED go test ./... -count=1`
- `golangci-lint run ./...`
- `go test -race ./pkg/coop/... ./pkg/cmd/coop -count=1`
- `scripts/test-coop-debug-agent-tmux.sh`
- Manual outside-tmux smoke with custom `XDG_CONFIG_HOME`, wide terminal size, debug agent review/confirm flow, and completion view

### DIFF
--- a/pkg/cmd/coop/coop_launcher.go
+++ b/pkg/cmd/coop/coop_launcher.go
@@ -346,9 +346,9 @@ func (rc *coopRunCmd) runFallbackWithCommand(stripeBin string, blueprintID strin
 			return err
 		}
 		fmt.Printf("Session started: %s\n", session.ID)
-		fmt.Printf("Open another terminal and run: stripe coop join %s\n", session.ID)
+		fmt.Printf("Open another terminal and run: %s\n", shellCommandWithCoopEnv("stripe coop join "+session.ID))
 	} else {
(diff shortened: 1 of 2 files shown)


vzhang-stripe (approved):

kwu-stripe (approved):
Do we also want to cover the no tmux path?

tomer-stripe:
Addressed the no-tmux fallback path in `6644792`:

- The fallback instructions now preserve `XDG_CONFIG_HOME` for both `stripe coop join <session>` and `stripe coop join --wait`.
- Added tests for both fallback instruction variants so the custom config-dir case is covered outside tmux as well.

kwu-stripe (approved):

PR #1773, opened 2026-07-10
coop: fix shell-injection in tmux launcher + wide-layout divider wrapping

tomer-stripe:
**Stack 1/3** (base: `coop`). The two release-blocking findings from the co-op mode review.

### 1. Shell-quoting in the tmux launcher (security)
The launcher built a `bash -c` script by interpolating paths with `strconv.Quote`, which produces a **Go** string literal, not shell quoting. Inside bash double quotes, `$(...)` and backticks still execute — so an env-controlled path (`TMPDIR`, `XDG_CONFIG_HOME`) or a hostile `--agent` value could inject commands. Replaced with a POSIX `shellQuote` helper (single-quote + `'\''` escaping) at all interpolation sites; kept `strconv.Itoa` for numeric args.

### 2. Wide-terminal layout wrapping (UX)
In the split workspace (width ≥ 100) the outline's divider rules and titles were sized to the full terminal width (up to 80) and wrapped into a broken multi-line mess inside the ~43-col left column — visible on any normal wide terminal. Added an `outlineWidthOverride` so the outline renders to the left-column width; the right detail panel keeps full width.

Dropped a third reported item (completion-header "tab leak") after verifying it's a `tmux capture-pane` artifact of bubbletea's cursor-movement optimization, not a defect — the real `View()` output is tab-free at every stage.

**Tests:** unit test for `shellQuote` metacharacter neutralization; regression test that the split outline fits the left column; updated an existing debug-agent test that had encoded the old unsafe double-quoting. Full race suite + golangci-lint clean; debug-agent tmux smoke test 16/16.

### DIFF
--- a/pkg/cmd/coop/coop_launcher.go
+++ b/pkg/cmd/coop/coop_launcher.go
@@ -175,7 +175,11 @@ func (rc *coopRunCmd) agentPaneCommandBuilder(agent *agentInfo, discoveryPrompt 
 			os.Remove(promptPath)
 			return "", nil, err
 		}
-		return agentCmd, func() {
+		// The returned command is run via `bash -c`, so the launcher path itself
+		// must be shell-quoted — otherwise a temp dir (TMPDIR) containing a space
+		// or shell syntax would be parsed by bash before the launcher runs. The
+		// cleanup closure keeps the raw path for os.Remove.
+		return shellQuote(agentCmd), func() {
 			os.Remove(promptPath)
 			os.Remove(agentCmd)
 		}, nil
--- a/pkg/cmd/coop/coop_launcher.go
+++ b/pkg/cmd/coop/coop_launcher.go
@@ -21,6 +21,15 @@ type agentInfo struct {
 	path string
 }
 
+// shellQuote wraps s in POSIX single quotes so it is safe to interpolate into a
+// `bash -c` command line. Unlike strconv.Quote (which produces a Go string
+// literal), single quoting neutralizes shell metacharacters including command
+// substitution ($(...), backticks), which would otherwise execute when the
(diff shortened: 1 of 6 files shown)


pkg/cmd/coop/coop_launcher.go
vzhang-stripe, line 173: `paneCmd` is still just the raw script path returned from `buildAgentCmd`. So if the temp path contains a space or shell syntax, bash -c parses it before the launcher script even starts

vzhang-stripe (changes requested):
this fixes quoting inside the generated launcher script, but the launcher script path itself is still passed raw to bash -c at `pkg/cmd/coop/coop_launcher.go:219` and `pkg/cmd/coop/coop_launcher.go:364`. Since that path comes from `os.CreateTemp("", ...)`, it can still inherit `TMPDIR` and contain spaces or shell metacharacters. 
So the outer shell can still misparse or expand the path before the script runs. I think we need to quote the launcher path too, or avoid bash -c and exec the script directly.

pkg/cmd/coop/coop_launcher.go
tomer-stripe: Good catch — fixed in 98e324e. `agentPaneCommandBuilder` now returns `shellQuote(agentCmd)`, so the launcher path is single-quoted before it reaches `bash -c` (the cleanup closure keeps the raw path for `os.Remove`). This covers all three `bash -c` call sites since they all consume the builder's output. Added a regression test that builds the pane command under a `TMPDIR` containing a space and `$(...)` and asserts it's exactly the single-quoted launcher path.

vzhang-stripe (approved):
🚢

PR #1774, opened 2026-07-10
coop: medium correctness and UX fixes

tomer-stripe:
**Stack 2/3** (base: `coop-fixes-high`). Eight medium-severity findings from the co-op mode review.

- **session:** `IsComplete()` returns false for a session with no nodes, so a fresh/malformed empty session isn't shown as "Integration complete."
- **tui:** gate the completion view to suggestion nav / select / quit / claim-open; work-view keys (expand, collapse, confirm, follow, reject, copy) no longer leak in to mutate hidden state or fire stray commands against the completion cursor.
- **tui:** `g`/`G`/`Home`/`End` now move the outline selection to the first/last item (and scroll to it) instead of being no-ops when the outline fits on screen (they only scrolled the viewport).
- **cmd/coop:** agent commands emit a structured `ok:false` JSON response with a hint on infra errors (e.g. heartbeat/store write failures mid-await) instead of plain text, so an agent parsing stdout always gets a recoverable response.
- **cmd/coop:** surface the specific blueprint error (e.g. an ambiguous prefix + its candidate list) from `coop run`/`coop start` instead of a generic "not found."
- **store:** `LatestSession()` skips unreadable/corrupt files so one bad file doesn't mask valid sessions from `status`/`join` (matches `LatestActiveSession`).
- **cmd/coop:** create the store *before* launching the tmux agent pane, so a store failure can't orphan the pane and leave a dangling active session.
- **workflow:** remove the unreachable single-node await path (`awaitNodeReview` + now-dead `rejectedResponse`); review always waits at step granularity.

**Tests:** regression tests for empty-session completion, completion-view key gating, top/bottom selection movement, and corrupt-file resilience; verified ambiguous-prefix output via the built CLI. Full race suite + golangci-lint clean.

### DIFF
--- a/pkg/cmd/coop/coop_agent.go
+++ b/pkg/cmd/coop/coop_agent.go
@@ -289,7 +289,15 @@ func validateFollowupParent(parent *coop.Session, actionID string) error {
 
 func outputAgentResponse(resp coop.CommandResponse, err error) error {
 	if err != nil {
-		return err
+		// Emit a structured ok:false response (on stdout, like every other agent
+		// command) so an agent parsing JSON always gets an error + recovery hint,
+		// even on infra failures (e.g. a heartbeat/store write error mid-await).
+		resp = coop.CommandResponse{
+			OK:    false,
+			Error: err.Error(),
+			Hint:  "stripe coop status",
+			Next:  "stripe coop status",
+		}
 	}
 	if outErr := outputJSON(resp); outErr != nil {
 		return outErr
--- a/pkg/cmd/coop/coop_agent.go
+++ b/pkg/cmd/coop/coop_agent.go
@@ -58,7 +58,7 @@ func newCoopAgentStartWorkCmd() *coopAgentActionCmd {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			service, err := newWorkflowService()
 			if err != nil {
-				return err
+				return outputAgentError(err)
 			}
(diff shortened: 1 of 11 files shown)


pkg/cmd/coop/coop_agent.go
```diff
@@ -289,7 +289,15 @@ func validateFollowupParent(parent *coop.Session, actionID string) error {
 
 func outputAgentResponse(resp coop.CommandResponse, err error) error {
 	if err != nil {
-		return err
+		// Emit a structured ok:false response (on stdout, like every other agent
```
kwu-stripe, line 300: Do we also use this structured response for other agent commands that return before this helper (ex. newWorkflowService() errors in start-work / report-work / etc)?

kwu-stripe (approved):

pkg/cmd/coop/coop_agent.go
tomer-stripe: Good catch — those paths did fall through to a plain-text error. Fixed in 68e103e: the `newWorkflowService()` (store-creation) failure in start-work / report-work / report-check / skip / await-review now routes through a new `outputAgentError` helper, which renders the same structured `ok:false` JSON (error + next hint, exit 1) as `outputAgentResponse`. Added a regression test for it. `next-action` / `start-followup` already emit structured errors via `outputCoopError`.

kwu-stripe (approved):

PR #1775, opened 2026-07-10
coop: durability, lock resilience, and dead-code polish

tomer-stripe:
**Stack 3/3** (base: `coop-fixes-medium`). Low-severity polish and hardening from the co-op mode review.

- **store:** fsync the temp file before rename and fsync the parent directory after, so a crash can't leave a renamed-but-truncated session file (atomic rename gives name-swap atomicity, not data durability).
- **store:** reclaim a stale lock left by a crashed writer. Lock files now record pid+time; a lock untouched longer than 30s (healthy writers hold it for milliseconds) is removed and retried, instead of wedging the session until manual deletion. The 30s threshold stays well above the 5s lock timeout, so active-writer contention still times out as before.
- **tui:** truncate the header claim URL with `ansi.Truncate` (width-aware) instead of byte-slicing, which could split a multibyte rune.
- **helpers:** `filterCompletedSuggestions` allocates a fresh slice instead of reusing the caller's backing array via `suggestions[:0]`.
- **workflow:** remove the dead `SelectNextAction` (no callers) and its now-orphaned `contains` helper; they encoded a next-steps write policy that diverged from the live lock-held `Update` path in the TUI.
- **test:** the debug-agent tmux smoke test now asserts the wide split workspace has no wrapped divider rules (previously it only checked text presence, so it passed even with the wide-layout bug).

Intentionally did **not** touch the bundled blueprint JSON (two blueprints use a single coarse review step) — per the README those are exported from the upstream Workbench source and shouldn't be hand-edited here.

**Tests:** full race suite + golangci-lint clean; debug-agent tmux smoke test 16/16 including the new divider assertion.

### DIFF
--- a/pkg/coop/store.go
+++ b/pkg/coop/store.go
@@ -158,37 +158,56 @@ func (s *Store) acquireSessionLock(path string) (func(), error) {
 }
 
 // lockAbandoned reports whether a lock file can be safely reclaimed. The lock
-// records its owner's PID: if that process is alive the lock must be left alone,
-// even if it is old, so a slow Store.Update callback can't have its lock deleted
-// out from under it. Only when the owner is known-dead is the lock reclaimed
-// immediately. When the owner can't be determined — an unparseable PID, or a
-// platform where liveness can't be checked (e.g. Windows) — fall back to
-// reclaiming by age so a crashed writer still can't wedge the session forever.
+// records its owner's PID and creation time:
+//   - If that process is alive the lock is left alone, even if old, so a slow
+//     Store.Update callback can't have its lock deleted out from under it.
+//   - Unless the process started *after* the lock was created: then the original
+//     writer crashed and the PID was reused by an unrelated process, so the lock
+//     is stale and reclaimed.
+//   - A known-dead owner's lock is reclaimed immediately.
+//   - When the owner can't be determined — an unparseable PID, or a platform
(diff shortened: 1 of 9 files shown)


pkg/coop/store.go
```diff
@@ -126,13 +131,23 @@ func (s *Store) acquireSessionLock(path string) (func(), error) {
...
 			f.Close()
 			return func() { os.Remove(lockPath) }, nil
 		}
 		if !os.IsExist(err) {
 			return nil, fmt.Errorf("creating lock file: %w", err)
 		}
 
+		// Reclaim a stale lock left behind by a crashed writer. O_EXCL on the next
+		// iteration ensures only one contender wins the recreate.
+		if info, statErr := os.Stat(lockPath); statErr == nil && time.Since(info.ModTime()) > sessionLockStale {
```
kwu-stripe: Could this accidentally remove a lock that still belongs to a running writer? Right now we treat the lock as stale based only on file age. If a `Store.Update` callback takes longer than `sessionLockStale` another process could delete its lock and start writing to the same session. 

Since the lock file records the PID, could we check whether that process is still alive before removing the lock?

kwu-stripe (approved):

kwu-stripe (approved):
It's technically possible that the writer crashes and the PID is reused: could we check that the the lock was created before the process was?

kwu-stripe (approved):

PR #1778, opened 2026-07-10
Add `--stripe-account` selector to `sandbox new`, drop --replica-of

shashwat-stripe:
### Summary

PR2 of two for `stripe sandbox new`, stacked on #1777.

The live parent (the account a sandbox belongs to) applies to both modes: `--copy-live-account` replicates its settings and `--create-blank` is still placed under its playground. Naming it by workspace id (`--replica-of wksp_...`) exposed an internal id and was a second, confusing way to say the same thing.

This replaces it with **`--stripe-account acct_...`**, the friendly, dashboard-visible id:
- Valid in **both** modes; defaults to the logged-in account when omitted.
- Resolves `acct_ → wksp_` via `GET /v2/compartments/user_accessible`, matching `merchant_id` across **both** `standalone_workspaces` and org-nested `organizations[].workspaces` (org-nested filtered to livemode), so an account under an organization still resolves without a raw workspace id.
- Input is `acct_` only; an `org_` is rejected up front, a sandbox belongs to an account, not an organization.

`--replica-of` is removed entirely.

### Test plan

- [x] `go test ./pkg/cmd/` (full package) passes.
- [x] New tests: `ResolvesByStripeAccount`, `StripeAccountBlankMode` (proves `--stripe-account` works without `--copy-live-account`), `StripeAccountNotFound`, `StripeAccountRejectsOrg`. Existing tests converted from `--replica-of` to `--stripe-account` with a `user_accessible` mock, assertions preserved.
- [x] `go build ./...`, `go vet ./pkg/cmd/...`, and `make lint` (golangci-lint, 0 issues) all clean.
- [ ] Live QA follow-up: exercise `--create-blank --stripe-account acct_...` end-to-end (backend call shape is unchanged, only the resolved playground differs, so low risk).

### Rollout/revert plan

Safe to revert.

cc @stripe/developer-products

### DIFF
--- a/pkg/cmd/sandbox.go
+++ b/pkg/cmd/sandbox.go
@@ -417,7 +417,7 @@ type sandboxClaimCmd struct {
 type sandboxNewCmd struct {
 	cmd              *cobra.Command
 	name             string
-	replicaOf        string
+	stripeAccount    string
 	businessLocation string
 	copyLiveAccount  bool
 	createBlank      bool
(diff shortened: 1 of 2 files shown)


echeng-stripe (approved):
Once the command-line parameters are stable, we should create a .md file with instructions on how to use the command.

PR #1788, opened 2026-07-14
APPLCM-3762 Add user id to the response

angelosolitario-stripe:
### Reviewers
r? @
cc @stripe/developer-products

 ### Summary

  Part of APPLCM-3762. The backend now returns a user_id field in the stripecli/auth polling response (captured at login time when the user authenticates via browser). This change plumbs that value through to the CLI profile config so plugins can read it without making
  additional API calls.

### Changes:
  - PollAPIKeyResponse — adds UserID string mapped to "user_id" JSON field
  - RAKConfigurer.SaveLoginDetails — writes UserID from the poll response to the profile
  - Profile — adds UserID field, UserIDName constant, persists it in writeProfile, exposes it via GetUserID(), and includes it in authFieldNames so it's cleared on logout/login

### Testing
After `stripe login` check `~/.config/stripe/config.toml` to see if user_id is stored

Screenshot 2026-07-16 at 2 49 47 PM

### DIFF
--- a/pkg/config/profile.go
+++ b/pkg/config/profile.go
@@ -39,6 +39,7 @@ type Profile struct {
 	TerminalPOSDeviceID    string
 	DisplayName            string
 	AccountID              string
+	UserID                 string
 	SandboxClaimURL        string
 	SandboxExpiresAt       string
 	UAT                    string
@@ -48,6 +49,7 @@ type Profile struct {
 // config key names
 const (
 	AccountIDName              = "account_id"
+	UserIDName                 = "user_id"
 	DeviceNameName             = "device_name"
 	DisplayNameName            = "display_name"
 	IsTermsAcceptanceValidName = "is_terms_acceptance_valid"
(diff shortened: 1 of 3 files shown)


pkg/config/profile.go
```diff
@@ -48,6 +49,7 @@ type Profile struct {
 // config key names
 const (
 	AccountIDName              = "account_id"
+	UserIDName                 = "user_id"
 	DeviceNameName             = "device_name"
```
NMandapaty, line 53: lol, this probably should have been called `Key` instead to avoid this. But not important, you don't need to fix it.

NMandapaty (approved):

pkg/config/profile.go
angelosolitario-stripe: 😆 I noticed this but didnt know if it was done on purpose so left it as is

yahanxing-stripe (approved):

PR #1790, opened 2026-07-15
Implement `stripe feedback` command

johno-stripe:
### Reviewers
r? @
cc @stripe/developer-products

CleanShot 2026-08-13 at 12 58 59

 ### Summary
Replaces the `stripe feedback` stub with a full implementation that submits feedback to `POST /v1/_unstable/feedback` (added server-side in a companion pay-server change). Fixes the `newFeedbackdCmd` typo (now `newFeedbackCmd`) and its registration in `root.go`.

The command supports two modes:
- **Interactive** (TTY, no flags): prompts for feature area, sentiment, message, and optional context; `actor` defaults to `human`.
- **Non-interactive**: requires `--sentiment`, `--message`, and `--context`; `--feature`/`--context`/`--actor` are optional.

Requests are sent as form-encoded POSTs via `stripe.Client.PerformRequest` with the `Stripe-Version` preview header, along with caller-supplied metadata (device name, CLI version, OS, detected terminal program, detected AI agent, CI flag). Adds `--json` output and a hidden `--api-base` flag for testability.

The options for features are currently hardcoded, but if this works as expected, we'll get them autogenerated from the API to reduce maintenance load.


pkg/cmd/feedback.go
```diff
@@ -1,39 +1,408 @@
...
+const feedbackEndpoint = "/v1/_unstable/feedback"
+
+// feedbackMinLen and feedbackMaxLen bound the length of the free-text
+// message and context fields, after sanitization.
+const (
+	feedbackMinLen = 10
+	feedbackMaxLen = 2000
+)
+
+var feedbackFeatureAreas = []string{
```
NMandapaty, line 36: Could we add `"apps"` as an option here?

johno-stripe: This field is validated server-side, so that'd be a separate change, but I'll add it in once this lands!

johno-stripe:
Heads up, @NMandapaty , I added in the `apps` option -- it'll be available on this first release :)

pkg/cmd/feedback.go
```diff
@@ -1,39 +1,454 @@
...
+
+func (feedback *feedbackCmd) submitFeedback(cmd *cobra.Command, apiKey string) (*feedbackResponse, error) {
+	baseURL, err := url.Parse(feedback.apiBaseURL)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &stripe.Client{
+		BaseURL:     baseURL,
+		Credentials: stripe.NewAPIKeyCredentials(apiKey),
```
vcheung-stripe: Could we use this factory to create the credentials (feel free to make it public or move it to another package): https://github.com/stripe/stripe-cli/blob/5cd954d8218ff89c9417c24905f2e5578b355416/pkg/requests/plugin.go#L20

Pretty soon we're going to migrate off of API keys and this will give you the right credential automatically.

vcheung-stripe (changes requested):

pkg/cmd/feedback.go
johno-stripe: Thanks for the tip -- I moved it into the config package, with the other profile code.  There was also a duplicate internal helper in the `plugins` package (`request/plugin.go` vs `plugins/plugin.go`, v odd), so I took that out and they both now call this shared function.

vcheung-stripe (approved):
Looks great! I tried it out with the old and new login flows and it works with both

PR #1795, opened 2026-07-15
Show spinner during agent setup status checks

gusnguyen-stripe:
### Summary
Show a spinner while `stripe agent setup --status` runs, covering both client detection and skills checks.

### Motivation
Status checks can take several seconds (subprocess calls + remote skills verification) with no terminal output, which makes it look hung. The install path already uses spinners; this brings the same feedback to `--status`.

### Test plan
- [x] Existing `stripe agent setup --status` tests pass
- [x] `make lint`
- [x] Spinner writes to stderr and auto-disables on non-TTY (tests/scripts unaffected)

### Rollout/revert plan
- [x] Safe to revert

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/cmd/agent.go
+++ b/pkg/cmd/agent.go
@@ -154,23 +154,27 @@ func (asc *agentSetupCmd) runSetup(cmd *cobra.Command, _ []string) error {
 	ctx := commandContextOrBackground(cmd)
 	out := cmd.OutOrStdout()
 
-	statuses := detectAll(providers)
-
-	for _, s := range statuses {
-		if s.Detected {
-			sendAgentEvent(ctx, "Agent Setup: Client Detected", s.Client)
-		}
-	}
-
-	detected := detectedStatuses(statuses)
-
-	if asc.jsonOutput || asc.statusOnly {
-		view, err := asc.skillsStatusView(ctx, detected)
+	if asc.statusOnly {
+		var statuses []agentsetup.Status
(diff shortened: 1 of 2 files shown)


pkg/cmd/agent.go
```diff
@@ -154,24 +154,28 @@ func (asc *agentSetupCmd) runSetup(cmd *cobra.Command, _ []string) error {
...
 	out := cmd.OutOrStdout()
 
-	statuses := detectAll(providers)
-
-	for _, s := range statuses {
-		if s.Detected {
-			sendAgentEvent(ctx, "Agent Setup: Client Detected", s.Client)
+	if asc.statusOnly {
+		var statuses []agentsetup.Status
+		err := spinner.New().
```
jleong-stripe: One small regression here is if a user does `stripe agent setup --status --json`, previously JSON would take precedence and it will output the JSON. Now it outputs the spinner + human readable status since we do `if asc.statusOnly` check first. 

I _think_ we may want --json to take precedence because that's likely called by agent and returns a richer output (that includes the status too).

jleong-stripe (commented):
left one comment

pkg/cmd/agent.go
gusnguyen-stripe: That's fair. I fixed it and now `--status --json` don't show spinner anymore and just show the json

kwu-stripe (approved):

PR #1803, opened 2026-07-17
Attempt to fix winget error: "0x8a15000f : Data required by the source is missing"

vcheung-stripe:
This tries to fix this winget error in the package manager test: https://github.com/stripe/stripe-cli/actions/runs/29601198297/job/87953455639

Tested the happy case: https://github.com/stripe/stripe-cli/actions/runs/29605558834/job/87967798594

Since I can't reproduce the issue, I'm going to just wait and see if this works.

### DIFF
--- a/scripts/install-test.sh
+++ b/scripts/install-test.sh
@@ -40,7 +40,6 @@ run_install() {
 
     winget)
         # Re-register the WinGet source AppX package to fix 0x8a15000f "data required is missing".
-        # shellcheck disable=SC2016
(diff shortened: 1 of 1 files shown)


matv-stripe (approved):
how do we test this? do we have a windows vm or something?

vcheung-stripe:
> how do we test this? do we have a windows vm or something?

I'm running this branch in github actions: https://github.com/stripe/stripe-cli/actions/workflows/install-test.yml

I don't really know how to reliably reproduce the issue, so I'm just planning to make sure this works in the happy case and hope the fix will work when the error comes up.

PR #1811, opened 2026-07-21
Add ability to set and unset docs preferences

joelzwarrington:
### Summary and motivation

Adds `stripe docs prefs set` and `stripe docs prefs unset` commands to persist documentation preferences locally. Also updates `stripe docs prefs list` to show the currently configured value for each preference.

Preferences are stored in the CLI config file under the active profile, so they carry over between sessions. `set` validates the ID and value against the live list from docs.stripe.com before writing.

### Test plan

- [X] Changes are covered by unit tests (including failures and edge cases)
- [X] Changes were tested manually

To test manually:

```bash
make
./stripe docs prefs list
./stripe docs prefs set server go
./stripe docs prefs list   # should show [current: go] next to server
./stripe docs prefs unset server
./stripe docs prefs list   # [current: ...] should be gone
```

You should see a styled success message on set/unset (`✓ Preference server set to go`), and the list should reflect the current value inline with each preference's allowed values.

CleanShot 2026-07-21 at 10 48 30

### DIFF
--- a/pkg/cmd/docs/prefs_test.go
+++ b/pkg/cmd/docs/prefs_test.go
@@ -243,3 +243,26 @@ func TestPrefsUnsetCommand(t *testing.T) {
 	assert.Contains(t, stripANSI(out.String()), "Preference lang unset")
 	assert.Equal(t, "", viper.GetString("default.docs_prefs.lang"))
 }
+
+func TestPrefsUnsetCommand_NoOps(t *testing.T) {
+	tests := []struct {
+		name string
+		args []string
+	}{
+		{name: "unknown pref", args: []string{"prefs", "unset", "unknown_pref"}},
+		{name: "not set", args: []string{"prefs", "unset", "lang"}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg, cleanup := setupPrefsTestConfig(t)
+			defer cleanup()
+
+			client := docs.NewClient("test").WithOptions(docs.WithBaseURL("http://unused"))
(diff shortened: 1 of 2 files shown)


pkg/cmd/docs/prefs_test.go
```diff
@@ -112,3 +137,109 @@ func TestPrefsListCommand_ServerError(t *testing.T) {
...
+	client := docs.NewClient("test").WithOptions(docs.WithBaseURL("http://unused"))
+	var out bytes.Buffer
+	root := cmd.New().WithOptions(cmd.WithClient(client), cmd.WithConfig(cfg)).Root()
+	root.SetOut(&out)
+	root.SetArgs([]string{"prefs", "unset", "lang"})
+
+	require.NoError(t, root.ExecuteContext(context.Background()))
+	assert.Contains(t, stripANSI(out.String()), "Preference lang unset")
+	assert.Equal(t, "", viper.GetString("default.docs_prefs.lang"))
+}
```
rayc-stripe, line 245: Can we add some failure cases for the unset command? What should we see if a pref is invalid or the value is not set but unset is run?

rayc-stripe (commented):
out-of-scope question: do we have a plan to set the pref for logged-in users by default (if they have some prefs stored on the server)?

pkg/cmd/docs/prefs_test.go
joelzwarrington: I added more tests which show that a failure isn't really possible and they're just no-ops.

rayc-stripe: should we consider showing some error/warning message when the unset command hits a invalid value?

rayc-stripe (approved):

PR #1823, opened 2026-07-22
coop: install pinned repo-scoped Stripe guidance

tomelm:
## Summary

- add pinned repo-scoped Stripe best-practices guidance for Codex and Claude Code
- prepare only the agent skill state needed for discovery or an explicit blueprint
- configure Claude's cost-effective Co-op worker and clarify permission-mode labels
- isolate the skill-preparation launcher test from host tmux installation and inherited tmux state
- make the no-API-key plugin manifest test use its local fixture server instead of live JFrog

## Test plan

- `go test -race ./pkg/cmd/coop -run TestCoopStartPreparesOnlyTheNeededAgentSkillState -count=100`
- `go test -race ./pkg/cmd/coop -count=1`
- `go test -failfast -race -coverpkg=./... -covermode=atomic -coverprofile=coverage.txt ./pkg/cmd/coop -run . -timeout=2m`
- `go test -race ./pkg/plugins -run '^TestRefreshPluginManifestSucceedsIfNoAPIKey$' -count=100`
- `go test -race ./pkg/plugins -count=1`
- `make test` (with Codex-only User-Agent environment markers removed to match GitHub CI)
- `git diff --check`


pkg/cmd/coop/coop_agent_skill.go
```diff
@@ -0,0 +1,285 @@
...
+	if _, err := project.Lstat(target); err == nil {
+		return false, nil
+	} else if !os.IsNotExist(err) {
+		return false, fmt.Errorf("checking existing Stripe skill: %w", err)
+	}
+
+	if err := project.MkdirAll(filepath.Dir(target), 0o755); err != nil {
+		return false, fmt.Errorf("creating project skills directory: %w", err)
+	}
+	if err := project.Mkdir(target, 0o755); err != nil {
```
kwu-stripe: This could leave a broken still if the directory is created by a later file write files or the process is interrupted. Future runs only check whether the target path exists, so they will skip repair even if `.agents/skills/stripe-best-practices` or `.claude/skills/stripe-best-practices` is empty or partial.

kwu-stripe (changes requested):

kwu-stripe (approved):

PR #1832, opened 2026-07-23
Add Sentry error reporting integration

vcheung-stripe:
This adds client-side error reporting to Sentry.
- Whenever a command fails with an unexpected error, we call `reporting.CaptureException(err)`. Some expected errors are like being unauthenticated and command typos, we won't report those.
- We time out after 2 seconds so that we don't block the command line for too long. We can tune this value if decide to later.
- Only the error message is logged. We redact secrets and don't track the user's machine.
- HTTP errors from resource commands (e.g. `stripe payment_intents create`) are not reported because they technically aren't errors semantically (they don't return a Go error and always exit 0)
- This respects the `STRIPE_CLI_TELEMETRY_OPTOUT` and `DO_NOT_TRACK` env vars. Errors will not be reported if these are set.

I ran a bunch of bad commands and observed them going into our Sentry project.


pkg/reporting/scrub.go
```diff
@@ -0,0 +1,60 @@
...
+	emailPattern = regexp.MustCompile(`[a-zA-Z0-9._%+\-]+@[a-zA-Z0-9.\-]+\.[a-zA-Z]{2,}`)
+)
+
+func redactSensitiveStrings(s string) string {
+	s = apiKeyPattern.ReplaceAllString(s, "${1}[REDACTED]")
+	s = oauthKeyPattern.ReplaceAllString(s, "${1}[REDACTED]")
+	s = webhookSecretPattern.ReplaceAllString(s, "whsec_[REDACTED]")
+	s = emailPattern.ReplaceAllString(s, "[REDACTED]")
+	return s
+}
```
yahanxing-stripe, line 27: nice!

pkg/reporting/scrub_test.go
```diff
@@ -0,0 +1,34 @@
...
+	}{
+		{"sk_live_abc123", "sk_live_[REDACTED]"},
+		{"sk_test_abc123", "sk_test_[REDACTED]"},
+		{"rk_live_abc123", "rk_live_[REDACTED]"},
+		{"rk_test_abc123", "rk_test_[REDACTED]"},
+		{"pk_live_abc123", "pk_live_[REDACTED]"},
+		{"pk_test_abc123", "pk_test_[REDACTED]"},
+		{"whsec_abc123==", "whsec_[REDACTED]"},
+		{"oak_abc123", "oak_[REDACTED]"},
+		{"oak_live_abc123", "oak_[REDACTED]"},
```
yahanxing-stripe, line 22: should the `oak_ ` pattern preserve the live/test segment like the others do?

yahanxing-stripe (approved):

pkg/reporting/scrub_test.go
vcheung-stripe: This is gonna be a new token type that won't be tied to a specific mode, so I don't think it'll be too meaningful

PR #1836, opened 2026-07-24
Load co-op blueprints from the API

tomelm:
## Summary

- load blueprint summaries and definitions from the unstable blueprint API using the configured test-mode key and preview version header
- make `coop recommend --all` return learning blueprint summaries for agent selection, then retrieve exact blueprint keys for `coop run` and `coop start`
- preserve the source blueprint step/node model, resolve live settings and parameter mappings, merge matching configured details, and retain structured request, event, UI, and reference data
- pin blueprint, step, and template versions together with the retrieved digest and effective definition in each local session
- remove the checked-in runtime catalog, exporter, and sync target while keeping representative API responses under `testdata`

## Validation

- `go test ./pkg/coop/... ./pkg/cmd/coop`
- `go test -race ./pkg/coop/... ./pkg/cmd/coop`
- `go vet ./...`
- `make build`
- manually exercised `coop recommend --all`, `coop run` with one-time payment and Issuing blueprints, and the full `coop start` split-TUI/review lifecycle

The full `go test ./...` sweep still encounters an unrelated environment-dependent failure: the Codex user-agent suffix does not match the existing stripeauth test regex.


tomelm:
Superseded by #1843 after renaming the fork branch.

PR #1843, opened 2026-07-25
Load co-op blueprints from the API

tomelm:
## Summary

- load blueprint summaries and definitions from the unstable blueprint API using the configured test-mode key and preview version header
- make `coop recommend --all` return learning blueprint summaries for agent selection, then retrieve exact blueprint keys for `coop run` and `coop start`
- preserve the source blueprint step/node model, resolve live settings and parameter mappings, merge matching configured details, and retain structured request, event, UI, and reference data
- pin blueprint, step, and template versions together with the retrieved digest and effective definition in each local session
- remove the checked-in runtime catalog, exporter, and sync target while keeping representative API responses under `testdata`

## Validation

- `go test ./pkg/coop/... ./pkg/cmd/coop`
- `go test -race ./pkg/coop/... ./pkg/cmd/coop`
- `go vet ./...`
- `make build`
- manually exercised `coop recommend --all`, `coop run` with one-time payment and Issuing blueprints, and the full `coop start` split-TUI/review lifecycle

The full `go test ./...` sweep still encounters an unrelated environment-dependent failure: the Codex user-agent suffix does not match the existing stripeauth test regex.


pkg/coop/blueprint_api.go
```diff
@@ -0,0 +1,307 @@
...
+		profile:    profile,
+		httpClient: httpClient,
+	}
+}
+
+func (c *BlueprintClient) List(ctx context.Context) ([]BlueprintSummary, error) {
+	var response struct {
+		Data []BlueprintSummary `json:"data"`
+	}
+	if err := c.get(ctx, blueprintsPath, &response, nil); err != nil {
```
kwu-stripe, line 55: The blueprints list endpoint has pagination so we might want to make sure we're getting all of the blueprints in the future.

pkg/coop/blueprint.go
```diff
@@ -1,310 +1,496 @@
...
+				request.Key = strconv.Itoa(index)
+			}
+		}
+	case NodeUIComponent:
+		if node.UIComponentDetails == nil {
+			return fmt.Errorf("uiComponent node is missing ui_component_details")
 		}
+		resolveUIComponent(node.UIComponentDetails, settings)
+	default:
+		return fmt.Errorf("unsupported node type %q", node.NodeType)
```
kwu-stripe, line 112: The resolver currently rejects every node type other than apiRequest/asyncHandler/testHelper/uiComponent. That is fine for the current visible catalog, but the Mongo blueprint model has additional node types, and Co-op already has generic lifecycle/guidance for some of them (`dashboard`, `setUpWebhooks`), but we could also just add handling for these nodes in the future.

kwu-stripe (approved):

PR #1842, opened 2026-07-24
coop: make the review card say what to do and where to do it

tomelm:
> **Stacked below #1844.** That PR builds on this one and revises parts of it as
> the design settled: `Where:` became `Check `, the check summariser was
> rewritten to lead with the failure's cause, and the hanging indents described
> below now apply only to lines that open with a status glyph. Review this PR on
> its own diff — the notes here are accurate for it — but expect those specifics
> to move in the PR above.

## Summary

The review card was hard to read and did not answer the two questions a reviewer actually has: *what am I confirming*, and *where do I go to confirm it*.

A step review rendered several sentences of agent narration under "Confirmation steps", because `reviewPromptLabel` preferred the agent's `verification.Check` prose over the blueprint's own review prompt. That prose describes what the agent already did, largely restates the changed-files line, and never states a task for the reviewer. Meanwhile the blueprint prompt — written for a human, e.g. *"Open the app and confirm the user-facing flow works as described."* — was suppressed whenever the agent emitted any passing check.

The card also summarised verifications as `Agent verified: 2/3 checks passed` while filtering failed checks out of the rendered text entirely, so the single most useful result was the one thing omitted.

## Changes

**Lead with the reviewer's task.** Blueprint review prompts are the card's primary content; agent narration moves to the detail view, which already renders it in full.

**Name failures, collapse passes.** Each failed check is listed (trimmed to its first sentence); passing checks become `✓ N checks passed`.

**Say where to go.** The venue differs by node type, and blueprint prompts don't close the gap because they're shared across every node of a type — both `apiRequest` nodes in `subscription-with-trial` carry identical generic wording. Derived from `node.Type`:

| Node type | Line |
|---|---|
| `apiRequest`, `testHelper` | `Where: the changed files below` |
| `uiComponent` | `Where: your running app` |
| `dashboard` | `Where: the Stripe Dashboard` |
| `cliCommand` | `Where: your terminal` |
| `asyncHandler`, `setUpWebhooks` | `Where: your webhook handler, after triggering the event` |

A `review_command` is itself the venue, so `Run:` replaces `Where:` rather than stacking with it.

**Hanging indents.** Wrapped file lists and checks stay attached to their label instead of reading as new items.

## Result

Before — agent prose, no stated task, failure hidden behind a ratio:

```
Confirmation steps
Create Checkout Session: Ran pnpm --filter web typecheck successfully. The new
POST /api/stripe/checkout route creates or reuses the catalog product, calls
[...]

PR #1844, opened 2026-07-25
coop: rebuild the review UI around the step

tomelm:
Stacked on #1842 — review that first; this PR's diff only reads correctly once it lands.

## Why

Confirming and requesting changes already acted on a whole step, but the outline made every task a cursor target and gave it equal weight. The UI asked users to approve atomic work that isn't independently verifiable — *confirm a Checkout Session is created*, without being able to complete the checkout.

Fixing that meant driving real sessions, and most of what follows came from doing so. The layout problems were not visible in the code; they were visible in the frames.

## What it looks like now

```
│
○  Project context
│ ╭──────────────────────────────────────────────────────────────╮
│ │ ✓ Scanned the repository: this is a pnpm/Turborepo monorepo   │
│ │   with a Next.js 16 app in apps/web, TypeScript/React, and…   │
│ ╰──────────────────────────────────────────────────────────────╯
╎
○  Set up product and pricing
╎ ╭──────────────────────────────────────────────────────────────╮
╎ │ ● Working on Create product and price                        │
╎ ╰──────────────────────────────────────────────────────────────╯
╎
●  Handle webhook events
╎ ╭──────────────────────────────────────────────────────────────╮
╎ │  Summary   Files   Reference                                 │
╎ │ ──────────────────────────────────────────────────────────── │
╎ │ To confirm                                                   │
╎ │ Run stripe trigger checkout.session.completed                │
╎ │     stripe trigger customer.subscription.created             │
╎ │                                                              │
╎ │ Confirm the handler receives and verifies the expected event.│
╎ │                                                              │
╎ │ ✗ 2 checks could not finish: STRIPE_WEBHOOK_SECRET is not    │
╎ │   configured.                                                │
╎ │                                                              │
╎ │ Set STRIPE_WEBHOOK_SECRET and re-run, or press r to send     │
╎ │ this back to the agent.                                      │
╎ │                                                              │
╎ │ Waiting for you   c confirm step · r request changes         │
╎ ╰──────────────────────────────────────────────────────────────╯
╎   ◆ Wait for checkout.session.completed event  ready  ✗1
╎     ✗ STRIPE_WEBHOOK_SECRET is not configured.
╎     apps/web/src/app/api/stripe/webhook/route.ts:1-70
```

## The changes

**The step is the unit.** Steps are the only cursor target. `reviewTarget.kind` and its three unreachable `kind == "node"` branches are gone — confirm already acted on the step on both paths. Tasks are status rows.

**One card, in one place.** The card used to render pinned at the bottom of the screen, with `enter` swapping that for an inline copy — two presentations of one step, in two places, and shrinking the terminal moved the content away from the step it described. It now renders under its own step, and arriving on a step opens it rather than requiring a second keystroke to see what the cursor is already pointing at.

**A timeline down the left.** Solid through finished steps, dotted through what is ahead, with a filled node on the step the cursor is on. It costs no width — it takes the columns the `> ` cursor marker had — and it replaced two other markers that were competing to say the same thing (`> ` and the `+`/`-` disclosure), plus the per-step rule that drew the same boundary as the card's own top border one row later.

**The card leads with the command.** Blueprint prompts are written to cover every way a step might be exercised — *"run the relevant Stripe CLI trigger or complete the upstream flow"* — which tells a reader holding a terminal nothing they can type, while the exact command sat two lines below phrased as an aside. Where a command exists it comes first and that opening clause is dropped.

**Failures say what failed, and what to do.** The summariser was tail-biased, but agent transcripts trail off into what *did* work after stating what did not, so a red ✗ was showing the reassuring half of the sentence, opened with a bare `…`:

> `✗ … in this environment. The handler reads the raw body, verifies Stripe's stripe-signature via constructEvent, and branches on checkout.session.completed.`

becomes

> `✗ STRIPE_WEBHOOK_SECRET is not configured.`

It drops progress noise, leads with the reason (preferring the clause after *because*), truncates only at the end, and normalises failures that name a missing configuration so one root cause reads as one root cause rather than as two differently-worded ones. The card then names the recovery: *set the variable and re-run, or press `r`*.

**Tasks and checks are one list.** The checks section restated every task title above it, so the same list appeared twice with the second copy prefixed by the first. Each task carries its own result; mixed results read as a ratio (`✗1 of 2`) rather than as `done ✗1 ✓1`.

**The agent contract.** `--check` is documented as a label but was the only field there is, so agents pasted whole transcripts into it. `Verification` gains `Detail` and `report-check` a `--detail` flag, with flag help spelling out which is which. Sessions written before this keep working — the summariser above is what handles them.

**Vocabulary.** A *node* is the atomic action in the blueprint; a *step* is the grouping, in both blueprint and TUI; a *task* is a node as the TUI and CLI present it. Agent-facing messages and flag help now say task.

## Live defects fixed along the way

- `--color off` never reached the TUI, so `stripe coop join --color off` rendered in full colour while every other command did not.
- `AttentionStyle` had no bold, so "needs you" — the highest-priority message — vanished without colour.
- A single torn read of the session file (polled every 500ms while the agent writes it) pinned the view on an error forever while the poll kept succeeding underneath.
- An uninitialised spinner renders the literal `(error)`; the guard against it used `MaxWidth`, which wraps rather than truncates, so it came out as seven rows of one character each.
- The viewport's height was computed twice with different arithmetic — once when sizing, once when drawing — so the viewport believed it was two rows taller than the region it was drawn into. Anything reasoning about what was on screen worked from the wrong number.
- The feedback editor was scrolled to the exact row a clipped card's closing border then occupied, leaving it one row under the fold.

## Verification

A frame-capture harness renders every state × width tier × colour mode to files:

```
COOP_TUI_CAPTURE_DIR=/tmp/frames go test ./pkg/coop/tui -run TestCapture
```

329 frames. Design review had been done against hand-drawn mocks that were wrong in ways only real data exposes; this replaced them, and found a header overflow on its first run.

`TestFramesAreNotVisiblyBroken` sweeps all of them for the two faults that make the UI look broken rather than merely cramped: content past the right edge, and a card that opens without closing. Both survived a green suite before it existed, because every assertion was about content rather than about whether the frame held together. A lone `╭────╮` hanging over blank space is what prompted it.

Monochrome frames go through `colorprofile.Ascii` rather than `ansi.Strip` — Strip removes bold and italic too, and would understate a layout whose hierarchy is meant to survive colour loss.

`go test ./pkg/coop/... ./pkg/cmd/coop/...`, `-race`, `go vet`, `gofmt`, `golangci-lint`, `go mod tidy` — all clean. Manual checklist against a real session:

```
~/code/stripe/demo-apps/verify-coop-tui-redesign.sh
```

## Known gaps

- Sessions created before this change carry no step description, so no `Goal` line.
- `stripe coop status` has not been extended with the new fields.
- The per-task detail view still exists with its own tabs, reachable only when a selection cannot resolve to a step — effectively never.
- The timeline reads solid when a step's tasks are all done, even if one of its checks could not finish. The card immediately below says so; encoding both in one glyph would muddy it.


pkg/coop/tui/detail.go
kwu-stripe, line 1: Do we want to surface the detail field in the TUI in the future?

kwu-stripe (approved):

PR #1845, opened 2026-07-27
Display newer plugin version available message in yellow

vladpotapov-stripe:
### Summary

Sometimes Stripe Plugin CLI users forget to update their packages, which may contain important updates. I believe it makes sense to increase the contrast of the message stating that a newer version of the Stripe Plugin is available.

| Current message | Updated message |
| - | - |
| Screenshot 2026-07-27 at 09 48 49 | Screenshot 2026-07-27 at 09 49 27 |

### DIFF
--- a/pkg/plugins/utilities.go
+++ b/pkg/plugins/utilities.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 
+	"github.com/stripe/stripe-cli/pkg/ansi"
 	"github.com/stripe/stripe-cli/pkg/config"
 	"github.com/stripe/stripe-cli/pkg/requests"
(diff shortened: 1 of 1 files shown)


kovalev-stripe (commented):
LGTM, but sharing my thoughts: in https://github.com/stripe/stripe-cli-apps-plugin/pull/301, they tried to solve a similar issue by adjusting the color of a printed message to improve contrast. 

Although both PRs improve specific use cases, I'm not sure that picking "good enough" colors case by case is a scalable approach.

Modern terminals support user-defined themes, so colors and contrast are pretty unpredictable. Instead of fixing individual messages, I think Stripe CLI should define a system of semantic color tokens (like a design system) that takes the shell environment into account -- e.g. whether `NO_COLOR` is set or whether the user is on a light or dark background.

vzhang-stripe (approved):
nice!

PR #1847, opened 2026-07-27
Use OAK token for API requests when available

vcheung-stripe:
Introduce support for making requests with the new token automatically if it's in the keychain. You need a flag enabled to receive it, otherwise we will just use an API key.

Same test procedure as https://github.com/stripe/stripe-cli/pull/1852. It currently does not work for some CLI specific commands but I am going to fix that in the backend:
- `stripe listen`
- `stripe logs tail`
- `stripe plugin install`

### DIFF
--- a/pkg/cmd/docs/root.go
+++ b/pkg/cmd/docs/root.go
@@ -148,8 +148,8 @@ func (r *RootCommand) initClient() {
 		if accountID, err := r.cfg.Profile.GetAccountID(); err == nil {
 			clientOpts = append(clientOpts, pkgdocs.WithCacheKeyPrefix(accountID))
 		}
-		if apiKey, err := r.cfg.Profile.GetAPIKey(false); err == nil {
-			clientOpts = append(clientOpts, pkgdocs.WithAPIKey(apiKey))
+		if creds, err := r.cfg.Profile.ResolveCredentials(false); err == nil {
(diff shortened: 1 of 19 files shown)


yahanxing-stripe (approved):

pkg/cmd/docs/root.go
```diff
@@ -148,8 +148,8 @@ func (r *RootCommand) initClient() {
 		if accountID, err := r.cfg.Profile.GetAccountID(); err == nil {
 			clientOpts = append(clientOpts, pkgdocs.WithCacheKeyPrefix(accountID))
 		}
-		if apiKey, err := r.cfg.Profile.GetAPIKey(false); err == nil {
-			clientOpts = append(clientOpts, pkgdocs.WithAPIKey(apiKey))
+		if creds, err := r.cfg.Profile.ResolveCredentials(false); err == nil {
+			clientOpts = append(clientOpts, pkgdocs.WithAPIKey(creds.Token))
```
vcheung-stripe, line 152: Confirmed these work:
- `stripe docs search "payment methods"`
- `stripe docs api product`

joelzwarrington: we don't yet use the API Key/credentials in actual API requests, so no FUD here!

joelzwarrington (approved):

PR #1857, opened 2026-07-29
Filter events by source in `--events-from` for terminal display

yahanxing-stripe:
### Summary

Previously --events-from only controlled forwarding URLs but the terminal still displayed all events regardless of source. Now `@self` suppresses connect events and `@accounts` suppresses self events from both display and forwarding.

Before:  
connect events show up even we specified `@self`
Screenshot 2026-07-29 at 2 55 46 PM

After: 
no connect events
Screenshot 2026-07-29 at 2 32 43 PM

### DIFF
--- a/pkg/proxy/webhook_event_processor.go
+++ b/pkg/proxy/webhook_event_processor.go
@@ -38,6 +38,9 @@ type WebhookEventProcessorConfig struct {
 
 	// LoggedInAccountID is the currently logged-in account ID
 	LoggedInAccountID string
+
+	// EventsFrom filters events by source: "@self", "@accounts", or "all"
+	EventsFrom string
 }
 
 // WebhookEventProcessor encapsulates logic around processing and forwarding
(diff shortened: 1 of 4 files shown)


pkg/proxy/webhook_event_processor.go
```diff
@@ -150,6 +153,10 @@ func (p *WebhookEventProcessor) processEvent(webhookEvent *websocket.WebhookEven
 		return
 	}
 
+	if p.filterBySource(&evt) {
```
vcheung-stripe, line 156: Can this be achieved by updating this logic? https://github.com/stripe/stripe-cli/blob/1c0a7c59caa9569f7dd81a192a6653b99b5f2dc5/pkg/proxy/proxy.go#L403

Seems inefficient to listen for the event if we're going to ignore it

vcheung-stripe (approved):

pkg/proxy/webhook_event_processor.go
yahanxing-stripe: the route building logic only controls forwarding, terminal display is a separate path, so we need to add display level filter to suppress them in terminal output. we could explore consolidating the logic as followup

PR #1862, opened 2026-07-30
coop: make the agent reliably wait for developer confirmation

tomelm:
## Summary

The agent frequently failed to wait for, or wake up on, the developer's confirmation — worst with cheaper models. The cause is a timeout race, not prompting.

`await-review` blocked for five minutes and asked agents to allow a six-minute shell timeout. That only holds if the model actually sets one. Claude Code's Bash tool defaults to **120s** and its override env vars have [open bugs](https://github.com/anthropics/claude-code/issues/34138), so a model that does not set a timeout has the command killed at two minutes — before the internal deadline — and receives a raw harness string with no `ok` and no `next` instead of structured JSON. Cheap models respond by assuming the step passed, or by stopping to ask a question in a pane the developer is not watching.

Each call now waits at most `AwaitTimeout` (45s, under the harness *default*) and returns `state: "waiting"` with `advance_allowed: false` and a `next` repeating the exact command just run. The developer is not rushed: review state lives in the session file, so the total wait stays unbounded across calls.

`next-action` was untouched by the earlier review-wait work and still had the original bug in its worst form — a ten-minute block that returned `ErrSelectionTimeout`, surfacing as JSON on **stderr with a non-zero exit**. An agent reads that as a broken session and gives up.

## Relationship to #1826

#1826 fixed the same class of problem by tagging the tmux agent pane and typing a resume prompt into it. That mechanism is removed here: it only works when Co-op launched the agent under tmux, and driving another program's input is not something to depend on. `stripe coop agent resume` is kept — it is read-only, types nothing, and is now the shared lifecycle query behind both the Stop hook and manual recovery.

`AwaitTimeout` / `AwaitHarnessTimeout` and the harness-headroom invariant from #1826 are kept, retuned to 45s / 90s so the interval sits under the harness *default* rather than only under an advertised timeout.

## Also fixed

Races that are reachable today and become routine once calls are frequent and short:

- **Requesting changes between two calls** moved the node back to `active`, so `AwaitReview` fell through to `alreadyMovedResponse`, whose `next` pointed at the following *pending* task. The agent skipped the rejected work and never saw the feedback.
- **`ShowSuggestions` cleared `NextSteps.Selected` unconditionally**, erasing a choice the developer made while the previous call was exiting.
- **Confirming the final task** completed the session, and the next `await-review` then failed `requireActiveSession` — non-zero exit, empty stdout, recovery pointing at `status` rather than `next-action`. That is the end of every session.

## Stop hook

Returning on an interval removes the hard failure, but a model can still stop mid-session and the CLI cannot force it back. A `Stop` hook can: it fires exactly when the agent tries to end its turn. It asks `resume` what the session needs rather than recomputing the lifecycle, and blocks with that exact command.

Injection is via launch flags only — `claude --settings` and `codex -c` — so nothing is written into the user's repo. Bounded by a 30-minute wall-clock window (what it detects is an absent developer; attempt count is a poor proxy), with a count ceiling as a runaway guard. When it does release the agent stops for good, so the TUI now reports the exact `stripe coop agent resume` command to paste, and keeps it on screen.

## Test plan

- `make test` (52 packages), `go test -race ./pkg/coop/... ./pkg/cmd/coop`, `make lint` — all clean
- Real end-to-end with **Claude Code** and **Codex** in tmux, driving actual sessions to review
- Verified the ping-pong loop by the elapsed counter resetting on a ~45s cycle across four `await-review` calls, then confirming from the TUI and watching the agent advance within seconds
- Verified the Stop hook fires in a live session by instructing the agent to stop: it was blocked and pulled back into the lifecycle
- Verified crash recovery (SIGKILL mid-await leaves no lock; next call works), the completion boundary, and the rejection path
- TUI frames rendered at 69/100/120 columns for the new status line

Three adversarial reviews found nine defects, fixed in the last four commits — including an escape hatch that could never fire because the command the hook ordered bumped the version the budget keyed on, and a Codex auto-approve path that silently dropped the hook.

## Notes for review

- **Codex gates hooks behind an interactive trust prompt** ("Hooks need review"); until the developer accepts it the hook is installed but inert and Co-op falls back to the interval alone. Claude Code loads settings-provided hooks directly.
- The 30-minute window is a judgement call, informed by "long enough to click through a checkout" rather than data. In both real-agent runs the hook never fired naturally — the models stayed in the loop on their own — so this layer may be insurance rather than a hot path.
- A stale `stripe` earlier on `PATH` silently reintroduces the old behavior, since agent instructions invoke bare `stripe`. Worth a version check in `coop run` as a follow-up; not included here.


pkg/cmd/coop/coop_stop_hook.go
```diff
@@ -0,0 +1,210 @@
...
+	Resume(sessionID string) (coop.CommandResponse, error)
+}
+
+func runCoopStopHook(out io.Writer, store stopHookStore, resumer stopHookResumer, sessionID string, now time.Time) error {
+	session := resolveStopHookSession(store, sessionID, now)
+	if session == nil {
+		return outputStopHookDecision(out, stopHookDecision{})
+	}
+
+	resume, err := resumer.Resume(session.ID)
```
kwu-stripe, line 105: Could we store if the developer selects Finish? Right now the agent runs stripe coop stop, but this hook then checks the completed session, receives another next-action command. and prevents the agent from existing.

kwu-stripe (changes requested):

pkg/cmd/coop/coop_stop_hook.go
tomelm: Good catch — reproduced exactly as you describe, and fixed in `bfa8070e`.

The gap was that nothing recorded the Finish choice. Every other next action reports back through `--completed=<id>` when its follow-up finishes; Finish has no follow-up, so consuming the selection is the only chance to record it. Without that the session sat as "completed with a next action still owed" forever, `Resume` kept answering with the next-action command, and the hook kept blocking — so the agent could never exit and the suggestions would reappear for the developer.

Two changes:

- `consumeSelection` records `done` (`coop.FinishActionID`) into `NextSteps.Completed` when that is what the developer picked.
- `Resume` returns no `next` for a finished session, which is already the hook's "nothing to do" signal, so it allows the stop.

Verified end to end against a real session:

```
1. next-action -> next = stripe coop stop --session=coop_06c23062
2. coop stop   -> state = completed
3. stop-hook   -> {}          # allow, agent exits
```

Step 3 previously returned `decision: "block"` with another next-action command.

The case that must not regress — a completed session where the developer has *not* chosen Finish — still blocks with `next-action`. Both directions are covered by `TestResumeReportsNothingToRunAfterDeveloperFinishes`, `TestResumeStillOffersNextActionBeforeFinish`, and `TestRunRecordsTheFinishSelection`.

Note the branch has been rebased onto the rewritten `coop` (it had been rewritten onto newer master, which was making this PR show a 102-file diff). The fix above was re-verified after that rebase, since it landed on a rewritten `resume.go`/`nextaction.go`.

PR #1864, opened 2026-07-31
Add --runtime-version flag to plugin commands

vzhang-stripe:
Allow users to specify which installed plugin version to run, bypassing the automatic version resolution that always prefers local.build.dev.

This feature is requested by the plugin owners

Locally tested:

Screenshot 2026-08-04 at 4 05 51 PM

Screenshot 2026-08-04 at 4 05 15 PM

 ### Reviewers
r? @
cc @stripe/developer-products

### DIFF
--- a/pkg/cmd/plugin_cmds.go
+++ b/pkg/cmd/plugin_cmds.go
@@ -61,6 +61,7 @@ func newPluginTemplateCmd(config *config.Config, plugin *plugins.Plugin) *plugin
 	}
 
 	ptc.cmd.Flags().String("runtime-version", "", "Run a specific installed plugin version (e.g. 1.5.2 or local.build.dev)")
+	ptc.cmd.Flags().MarkHidden("runtime-version") // #nosec G104 -- development-only flag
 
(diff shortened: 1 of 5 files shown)


pkg/cmd/plugin_cmds.go
```diff
@@ -59,6 +60,8 @@ func newPluginTemplateCmd(config *config.Config, plugin *plugins.Plugin) *plugin
 		},
 	}
 
+	ptc.cmd.Flags().String("runtime-version", "", "Run a specific installed plugin version (e.g. 1.5.2 or local.build.dev)")
```
vcheung-stripe, line 63: Since this is meant for development only, should we make this flag hidden?

vcheung-stripe (changes requested):

pkg/cmd/plugin_cmds.go
vzhang-stripe: good idea

vcheung-stripe (approved):

PR #1887, opened 2026-08-10
Add categories to reported CLI errors

yahanxing-stripe:
### Summary
Adds a stable `error_category` tag to reported CLI errors and recovered panics, classifying recognizable flag, API, filesystem, network, cancellation, and validation failures while preserving existing error messages, exit behavior, and event capture.

### DIFF
--- a/pkg/validators/cmds.go
+++ b/pkg/validators/cmds.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/errorcategory"
 )
 
 func getCommandPath(cmd *cobra.Command) string {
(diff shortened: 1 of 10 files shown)


pkg/validators/cmds.go
```diff
@@ -29,7 +31,7 @@ func NoArgs(cmd *cobra.Command, args []string) error {
 	)
 
 	if len(args) > 0 {
-		return errors.New(errorMessage)
+		return errorcategory.With(errors.New(errorMessage), errorcategory.UserInput)
```
vcheung-stripe: Nit: I kinda like how `authError` looks. wdyt of having different methods for each category?

vcheung-stripe (approved):

pkg/validators/cmds.go
yahanxing-stripe: good call, i'll wrap `userInputError` construction into a helper method

PR #1888, opened 2026-08-10
Capture agent host and self-reported agent identifiers in telemetry

tomelm:
> **Supersedes #1889**, which is merged into this branch.

## Why

`DetectAIAgent` can't tell a desktop-hosted agent from a terminal one — Claude Code in Claude Desktop and Claude Code in iTerm both report exactly `ai_agent=claude_code`. The agents already export the answer and we drop it on the floor.

Environments captured from real sessions across both vendors and both platforms:

| variable | Claude Desktop (macOS) | Claude Desktop (Windows) | Codex Desktop (Windows) |
|---|---|---|---|
| `CLAUDECODE` | `1` | `1` | — |
| `CLAUDE_CODE_ENTRYPOINT` | `claude-desktop` | `claude-desktop` | — |
| `AI_AGENT` | `claude-code_2-1-222_agent` | `claude-code_2-1-227_agent` | — |
| `CODEX_INTERNAL_ORIGINATOR_OVERRIDE` | — | — | `Codex Desktop` |
| `CODEX_CI` | — | — | `1` |
| `TERM_PROGRAM` | *(unset)* | *(unset)* | *(unset)* |

Two things these settle. The Claude version format is **stable across builds** — the same parse handles `2-1-222` and `2-1-227` — so `agent_version` isn't over-fit to one sample. And **Codex Desktop sets the generic Codex signals too** (`CODEX_CI=1`), so its originator variable is the only thing separating Desktop from the CLI.

Note the last row: with `TERM_PROGRAM` unset, a desktop-hosted run currently looks identical to a headless or piped terminal run.

## What

Two fields alongside the existing `ai_agent`, whose semantics are unchanged:

| field | values |
|---|---|
| `agent_host_kind` | `desktop`, `terminal`, `ide`, `remote`, `sdk`, `mcp`, `other` |
| `agent_host_raw` | the normalized host underneath the category, e.g. `claude-desktop-3p`, `sdk-ts` |
| `agent_version` | e.g. `2.1.227`, empty when the agent reports none |

```
agent_host_kind=desktop&agent_version=2.1.227&ai_agent=claude_code&...
```

The two host fields split by purpose: **group by `agent_host_kind`, drill into `agent_host_raw`.** Those are per-vendor spellings every consumer would have to enumerate, and the vendor is already carried by `ai_agent` — so `ai_agent` + `agent_host_kind` says "Claude Code, on desktop" in two columns instead of three. A host we don't categorize reports `other`, so a new or renamed host stays visible without the field being unbounded.

**`agent_host_raw` exists because the categories are deliberately coarse, and coarseness is not recoverable after the fact.** `claude-desktop` and `claude-desktop-3p` are both `desktop`; `sdk-ts` and `sdk-py` are both `sdk`. Without the raw value nothing downstream can tell them apart, and data collected without it simply doesn't contain the distinction.

It also solves a release-cycle problem. `agent_host_kind=other` says something is unmapped but not what — identifying it has meant reading a vendor's minified bundle, and mapping it then costs a code change, a release, and users upgrading, so an unmapped host is invisible for months. `claude-desktop-3p` was miscounted exactly that way until a grep happened to surface it. With the raw value it's answerable from the data.

Cardinality stays small: Claude Code ships 25 entrypoint values and Codex a handful. Normalization runs before reporting, so one host can't arrive under several spellings across platforms — a Windows `Codex Web` and a macOS `codex_web` both land as `codex-web`.

Two details worth a reviewer's eye:

- **`remote_desktop` is categorized as `remote`, not `desktop`.** Claude Code's own surface mapping labels `claude-desktop`, `claude-desktop-3p` and `remote_desktop` all as "Claude Desktop", but in the third the session executes remotely, so the CLI runs on the remote host rather than the user's machine — and this field describes where the CLI ran. Remote hosts are matched by prefix *before* the map so a substring match on `desktop` can't put them in the wrong bucket.
- **`agent_version` refuses to guess.** Claude Code encodes the version in `AI_AGENT` as `claude-code_2-1-227_agent` (dots as dashes) and has also used `claude-code/2.1.227`. Neither format is documented, so the parse validates and reports nothing on anything that isn't a plain dot-separated number: `v2.1.227`, `beta`, `2.1.`, `2..1` and overlong values all yield empty.

## What is deliberately not collected

Every reported value is a label from a fixed set, a validated version number, or the normalized host identifier.

| variable | handling |
|---|---|
| `CLAUDE_CODE_SESSION_ID`, `CLAUDE_CODE_HOST_SESSION_ID`, `CLAUDE_CODE_OAUTH_SCOPES`, `CODEX_PERMISSION_PROFILE` | never read |
| `CODEX_THREAD_ID`, `CODEX_SANDBOX_NETWORK_DISABLED`, `CODEX_CI` | presence checks only; values never captured |
| `CLAUDE_CODE_ENTRYPOINT`, `CODEX_INTERNAL_ORIGINATOR_OVERRIDE` | the derived category, plus the normalized value itself |
| `AI_AGENT` | only the parsed version number is reported |

`agent_host_raw` is the one field carrying a value not drawn from a fixed set, and it's worth flagging explicitly for review. It comes from two allowlisted variables whose stated purpose is self-identification — not session, thread, or auth data — and it's normalized, stripped to printable ASCII, and bounded to 32 characters.

`TestObservedAgentSessions_NoSensitiveValuesReported` asserts this on output rather than on access, since several of those variables genuinely are read during detection.

## Relationship to #1889

#1889 is merged here rather than landing separately. Its detection work is kept; its `codex_desktop` `ai_agent` value is not.

Adding `codex_desktop` to the enum would have encoded the host in two places, and made "how many Codex invocations" a list — `ai_agent IN ('codex_cli','codex_desktop')` — that grows with every new Codex host. `CODEX_INTERNAL_ORIGINATOR_OVERRIDE` is also internal and undocumented, so it drives the best-effort `agent_host_kind` rather than a value dashboards group by. Codex Desktop remains fully visible as `agent_host_kind=desktop`.

The cost: a Codex Desktop run reports `ai_agent=codex_cli`. That's the pre-existing wart of a host baked into an agent name; renaming it to `codex` breaks already-shipped data, so it belongs in its own change.

Also from #1889: per-variable Codex test coverage (`TestAIAgentDetection_AllAgents` previously only exercised `CODEX_SANDBOX`), and a fix to `TestUserAgent`, which asserted `^Stripe/v1 stripe-cli/\w+$` but fails whenever the suite is run from a coding agent — `initUserAgent` appends an `AIAgent/...` suffix. That was red on master for anyone running `make test` from Claude Code or Cursor.

## Testing

- `TestObservedAgentSessions` pins the detectors against the four captured environments above, so a vendor changing what it exports fails a test rather than silently degrading reporting.
- `TestObservedAgentSessions_NoSensitiveValuesReported` asserts no session, host, thread or scope value reaches a reported field.
- `agent_host_raw` reported for every host and empty only when no host was reported, including the `claude-desktop` vs `claude-desktop-3p` split, normalization, the 32-character bound, and non-printable stripping.
- Version parsing across both formats plus five rejection cases; host categorization including the `remote-desktop` trap and the `other` bucket.
- Telemetry assertions verified to actually execute via deliberate negative controls — they run inside an `httptest` handler, where a silently-skipped assertion would otherwise pass green.
- Full `go test ./...` green, `make lint` 0 issues.

### DIFF
--- a/pkg/stripe/analytics_telemetry.go
+++ b/pkg/stripe/analytics_telemetry.go
@@ -51,7 +51,8 @@ type CLIAnalyticsEventMetadata struct {
 	GeneratedResource bool   `url:"generated_resource"`          // whether or not this was a generated resource
 	AIAgent           string `url:"ai_agent,omitempty"`          // the AI coding agent that invoked the CLI, if any
 	AIAgentRaw        string `url:"ai_agent_raw,omitempty"`      // the agent's self-reported AI_AGENT/AGENT value; unvalidated, prefer ai_agent for grouping
-	AgentHost         string `url:"agent_host,omitempty"`        // the application hosting the agent (desktop, IDE, SDK), when reported
+	AgentHost         string `url:"agent_host,omitempty"`        // the application hosting the agent, as reported by the agent
+	AgentHostKind     string `url:"agent_host_kind,omitempty"`   // coarse category for agent_host: desktop, terminal, ide, remote, sdk, mcp, chat, or ci
 	MacAppBundleID    string `url:"mac_app_bundle_id,omitempty"` // bundle ID of the macOS app that launched the process tree, if any
 	InstallMethod     string `url:"install_method,omitempty"`    // how the CLI was installed
 	InTmux            bool   `url:"in_tmux"`                     // whether the CLI was invoked from within tmux
@@ -82,6 +83,8 @@ type NoOpTelemetryClient struct {
 
 // NewEventMetadata initializes an instance of CLIAnalyticsEventContext
 func NewEventMetadata() *CLIAnalyticsEventMetadata {
+	agentHost := useragent.DetectAgentHost(os.Getenv)
+
 	return &CLIAnalyticsEventMetadata{
 		InvocationID:   uuid.NewString(),
 		CLIVersion:     version.Version,
@@ -89,7 +92,8 @@ func NewEventMetadata() *CLIAnalyticsEventMetadata {
 		Arch:           runtime.GOARCH,
 		AIAgent:        useragent.DetectAIAgent(os.Getenv),
 		AIAgentRaw:     useragent.DetectAIAgentRaw(os.Getenv),
-		AgentHost:      useragent.DetectAgentHost(os.Getenv),
+		AgentHost:      agentHost,
+		AgentHostKind:  useragent.AgentHostKind(agentHost),
 		MacAppBundleID: useragent.DetectMacAppBundleID(os.Getenv),
 		InstallMethod: useragent.DetectInstallMethod(
 			os.Getenv,
--- a/pkg/stripe/analytics_telemetry.go
+++ b/pkg/stripe/analytics_telemetry.go
@@ -37,27 +37,26 @@ const DefaultTelemetryEndpoint = "https://r.stripe.com/0"
 
 // CLIAnalyticsEventMetadata is the structure that holds telemetry data context that is ultimately sent to the Stripe Analytics Service.
 type CLIAnalyticsEventMetadata struct {
-	InvocationID      string `url:"invocation_id"`               // The invocation id is unique to each context object and represents all events coming from one command / gRPC method call
-	UserAgent         string `url:"user_agent"`                  // the application that is used to create this request
-	CommandPath       string `url:"command_path"`                // the command or gRPC method that initiated this request
-	CommandFlags      string `url:"command_flags"`               // Comma-separated list of flags that were passed to the command (only includes flag names, not their values)
-	PluginName        string `url:"plugin_name,omitempty"`       // the plugin being installed, when relevant
-	PluginVersion     string `url:"plugin_version,omitempty"`    // the version of the plugin being installed/uninstalled/upgraded
-	Merchant          string `url:"merchant"`                    // the merchant ID: ex. acct_xxxx
-	CLIVersion        string `url:"cli_version"`                 // the version of the CLI
-	OS                string `url:"os"`                          // the OS of the system
-	Arch              string `url:"arch"`                        // the CPU architecture of the system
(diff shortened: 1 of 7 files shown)


pkg/useragent/useragent.go
```diff
@@ -125,6 +125,66 @@ func DetectAIAgent(getEnv func(string) string) string {
 	return ""
 }
 
+// DetectAgentHostKind reports where the agent ran: "desktop", "terminal", "ide",
```
matv-stripe, line 128: when will this be "mcp" ?

pkg/stripe/analytics_telemetry.go
```diff
@@ -50,6 +50,8 @@ type CLIAnalyticsEventMetadata struct {
 	MachineUUID       string `url:"machine_uuid,omitempty"`     // the persistent machine UUID
 	GeneratedResource bool   `url:"generated_resource"`         // whether or not this was a generated resource
 	AIAgent           string `url:"ai_agent,omitempty"`         // the AI coding agent that invoked the CLI, if any
+	AgentHostKind     string `url:"agent_host_kind,omitempty"`  // where the agent ran: desktop, terminal, ide, remote, sdk, mcp, chat, ci, or other
```
matv-stripe, line 53: left below but confused when this will be mcp

matv-stripe (approved):

pkg/useragent/useragent.go
tomer-stripe: it should never. i saw this too and i think claude got a little carried away with itself, i'll remove it

tomer-stripe: i stand corrected, apparently this can be set as `mcp` if you run `claude mcp serve`

PR #1889, opened 2026-08-10
Detect Codex Desktop invocations

tomelm:
### Summary

- report commands launched by Codex Desktop as `codex_desktop` when `CODEX_INTERNAL_ORIGINATOR_OVERRIDE=Codex Desktop`
- check the desktop origin before generic Codex signals because Desktop also sets `CODEX_THREAD_ID` and `CODEX_CI`
- add coverage for every Codex environment signal and allow the intentional AI-agent suffix in the Stripe auth User-Agent test

`CODEX_INTERNAL_ORIGINATOR_OVERRIDE` is internal rather than a documented public contract, but it is currently the only inherited signal that distinguishes Desktop from the Codex CLI. Matching the exact `Codex Desktop` value keeps the detection narrow.

### Testing

- `make test`
- `make lint`

### DIFF
--- a/pkg/useragent/useragent.go
+++ b/pkg/useragent/useragent.go
@@ -107,6 +107,10 @@ func DetectAIAgent(getEnv func(string) string) string {
 	if getEnv("CLINE_ACTIVE") != "" {
 		return "cline"
 	}
+	// Codex Desktop also sets the generic Codex env vars, so check its originator first.
+	if getEnv("CODEX_INTERNAL_ORIGINATOR_OVERRIDE") == "Codex Desktop" {
+		return "codex_desktop"
+	}
 	if getEnv("CODEX_SANDBOX") != "" || getEnv("CODEX_THREAD_ID") != "" || getEnv("CODEX_SANDBOX_NETWORK_DISABLED") != "" || getEnv("CODEX_CI") != "" {
 		return "codex_cli"
 	}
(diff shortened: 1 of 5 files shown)


pkg/useragent/useragent.go
```diff
@@ -107,6 +107,10 @@ func DetectAIAgent(getEnv func(string) string) string {
 	if getEnv("CLINE_ACTIVE") != "" {
 		return "cline"
 	}
+	// Codex Desktop also sets the generic Codex env vars, so check its originator first.
```
matv-stripe, line 110: eventually would like to replace this lgic with https://github.com/stripe/detect-agent , havent gotten around to it

matv-stripe (approved):

tomelm:
Merged into #1888 rather than landing separately, so the Codex Desktop and Claude Desktop cases can be reviewed against one schema instead of two.

Everything here is preserved in that branch — the `codex_desktop` detection, the per-variable Codex test coverage, the `CODEX_INTERNAL_ORIGINATOR_OVERRIDE` additions to the env-clearing lists, and the `TestUserAgent` regex fix (which both branches had identically, so it merged as a no-op).

#1888 also routes `CODEX_INTERNAL_ORIGINATOR_OVERRIDE` into a new `agent_host` field, which means Codex Desktop is currently encoded twice: `ai_agent=codex_desktop` and `agent_host_kind=desktop`. Whether to keep both is called out as an open decision there.

PR #1902, opened 2026-08-13
coop: bundle a version-locked stripe-coop Agent Skill

tomelm:
## What

Bundles a version-locked `stripe-coop` Agent Skill inside the CLI binary and installs it into the selected coding harness before `stripe coop start` launches an agent. The skill carries Co-op's stable lifecycle and safety contract, so launch prompts shrink to skill activation plus the dynamic session context.

## Why

The lifecycle contract was re-sent in full on every launch prompt and every command response. Moving it into a skill that the agent loads on demand keeps the prompt focused on the user's project and integration intent, and gives us one portable place to teach agents the protocol.

The skill is embedded (`go:embed`) rather than downloaded: it is part of the Co-op protocol and must stay compatible with the exact CLI version, so it deliberately does not reuse the network-backed `pkg/agentskills` installer.

## How it works

- **Skill** — `pkg/coop/skill/stripe-coop/`: `SKILL.md` plus `references/{command-api,node-contracts,recovery}.md`. The frontmatter description scopes activation to Co-op sessions and away from ordinary Stripe development.
- **Install** — atomic (staged directory + rename) and idempotent (sha256 content hash in a `.stripe-cli-skill.json` manifest). A same-name skill directory without that manifest is treated as user-authored and never overwritten; the launch falls back to the previous self-contained prompt. Abandoned staging/retired directories older than an hour are swept.
- **Harness adapters** — `pkg/cmd/coop/coop_harness.go` replaces the old name/path switch. Each adapter declares executables, skill directory, activation syntax, launch arguments, approval flags, and capability limits.
- **Scope** — only `stripe coop start` installs. `join`, `status`, `stop`, and `recommend` do not. `coop run` and command responses stay self-contained for agents invoked directly.

## Harness support

Verified against current official harness docs (July 2026):

| Harness | Skill directory | Activation | Notes |
|---|---|---|---|
| Claude Code | `~/.claude/skills` | `/stripe-coop` | `--dangerously-skip-permissions` bypass |
| Codex | `~/.agents/skills` | `$stripe-coop` | `--dangerously-bypass-approvals-and-sandbox` bypass |
| OpenCode | `~/.agents/skills` | description match | `opencode --prompt`, `--auto` |
| Cline | `~/.cline/skills` | `/stripe-coop` | one-shot launch; no conversational discovery |
| Roo Code | `~/.agents/skills` | description match | auto-approves by default; discontinued May 2026 |
| OpenHands | `~/.agents/skills` | description match | `openhands -t`, `--always-approve` |
| Goose | `~/.agents/skills` | description match | `goose run -t`, `GOOSE_MODE` env; no conversational discovery |

Harnesses that cannot hold an interactive conversation from a prompt-seeded launch are rejected for no-blueprint discovery with an actionable message rather than silently launching into a broken flow. Unknown `--agent` values get a generic positional-prompt launcher, no skill install, and the full fallback prompt.

## Protocol

Adds `protocol_version` to rendered agent responses so agents can detect contract drift. `recommend` now emits `ok`/`protocol_version` and structured recovery errors instead of plain text. `start-work` notes strip angle brackets so node titles like `<PaymentElement>` cannot trip the placeholder validator.

## Testing

- `go test ./pkg/coop/... ./pkg/cmd/coop -count=1` — green
- `golangci-lint run` — 0 issues; `go build ./...`; `git diff --check` clean
- skill-creator `quick_validate.py` — valid
- New coverage: embed layout/frontmatter/reference integrity, install atomicity/idempotence/upgrade/unmanaged-refusal/debris-sweep, per-harness skill dirs and activation tokens, launch-arg construction, compact-prompt content and size caps, protocol-version stamping, and a lockstep test that every registered `stripe coop agent` subcommand is documented in the bundled `command-api.md`
- The Goose launcher test executes the generated script against a fake binary rather than asserting on substrings

An adversarial multi-agent review of this branch surfaced 8 confirmed findings, all fixed in the second commit — most notably that the Goose launcher placed the env assignment after `exec`, so `bash` tried to run a program literally named `GOOSE_MODE=auto` and the harness could never start.

## Follow-ups (not in this PR)

- Give `${node...}` outputs a typed schema in blueprints instead of free-form `field=value` JSON
- Real-agent end-to-end pass of the skill (this branch has a doc-comprehension forward-test only)
- Cline's docs disagree on `--yolo` vs `--auto-approve`; worth re-verifying before release

### DIFF
--- a/pkg/cmd/coop/coop_harness.go
+++ b/pkg/cmd/coop/coop_harness.go
@@ -304,7 +304,10 @@ func (rc *coopRunCmd) buildAgentCmd(agent *agentInfo, promptPath string, autoApp
 		envPrefix = adapter.envPrefix(autoApprove)
 	}
 
-	script := fmt.Sprintf("#!/bin/bash\nprompt=$(cat %s)\nrm -f %s %s\nexec %s%s %s\n",
+	// The env prefix must precede the `exec` keyword: bash only treats
+	// KEY=value words as assignments before the command word, and `exec` is
+	// the command word here.
+	script := fmt.Sprintf("#!/bin/bash\nprompt=$(cat %s)\nrm -f %s %s\n%sexec %s %s\n",
 		shellQuote(promptPath), shellQuote(promptPath), shellQuote(launcherPath),
 		envPrefix, shellQuote(agent.path), strings.Join(args, " "))
 
--- a/pkg/cmd/coop/coop_harness.go
+++ b/pkg/cmd/coop/coop_harness.go
@@ -1,0 +1,342 @@
+package coopcmd
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"charm.land/huh/v2"
+
+	"github.com/stripe/stripe-cli/pkg/coop/skill"
+)
+
+// harnessAdapter describes how one coding harness is detected, taught the
+// bundled stripe-coop skill, and launched. Everything harness-specific lives
+// here; the skill content itself stays portable.
+//
+// Verified against official harness documentation, July 2026.
+type harnessAdapter struct {
+	// id is the stable identifier accepted by --agent.
+	id          string
+	displayName string
+	// executables are the PATH candidates that select this adapter.
+	executables []string
+	// skillDir returns the harness's user-level skill directory for the
+	// bundled stripe-coop skill, relative to the user's home directory.
+	skillDir func(home string) string
+	// activationLine is the first line of every launch prompt. It uses the
+	// harness's documented explicit skill activation syntax.
+	activationLine string
+	// promptClause places the launch prompt on the command line. It must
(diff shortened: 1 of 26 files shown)


pkg/cmd/coop/coop_harness.go
```diff
@@ -0,0 +1,345 @@
...
+	interactiveLaunch bool
+	// notes describes capability limitations surfaced to the developer.
+	notes string
+}
+
+// agentsSkillsDir is the cross-harness skills directory from the Agent Skills
+// standard, read by Codex, OpenCode, Roo Code, OpenHands, and Goose.
+func agentsSkillsDir(home string) string {
+	return filepath.Join(home, ".agents", "skills")
+}
```
matv-stripe, line 58: Is this duplicative of https://github.com/stripe/stripe-cli/tree/master/pkg/agentskills ?

PR #1908, opened 2026-08-14
Categorize Codex TypeScript SDK invocations

tomelm:
### Summary

- map the Codex TypeScript SDK originator `codex_sdk_ts` to `agent_host_kind=sdk` instead of `other`
- report its normalized originator as `agent_host_raw=codex-sdk-ts`, preserving visibility if vendor values drift
- preserve the existing `ai_agent=codex_cli` value for compatibility
- verify that `terminal_program` is captured when the SDK inherits a terminal environment
- add focused host detection coverage and a complete SDK-session fixture

### Why

`@openai/codex-sdk` [sets](https://github.com/openai/codex/blob/42b5f05cef69491bc578901fb324b3c9a278b253/sdk/typescript/src/exec.ts#L43-L45) `CODEX_INTERNAL_ORIGINATOR_OVERRIDE=codex_sdk_ts` when it launches Codex. #1888 introduced the host category, but the Codex SDK originator fell through to `other`.

Rebased on #1909, the raw host is reported for mapped and unmapped values alike. The resulting fields are:

- `ai_agent=codex_cli`
- `agent_host_kind=sdk`
- `agent_host_raw=codex-sdk-ts`

The SDK [forwards its caller environment](https://github.com/openai/codex/blob/42b5f05cef69491bc578901fb324b3c9a278b253/sdk/typescript/src/exec.ts#L161-L183), so existing terminal detection continues to populate `terminal_program` when values such as `TERM_PROGRAM` are present. Headless SDK runs leave it empty.

### Version availability

`agent_version` remains empty for current Codex SDK runs because there is no version signal in the child-command environment. The TypeScript SDK exports only the originator above. The Python SDK has a version, but [sends it in app-server initialization metadata](https://github.com/openai/codex/blob/42b5f05cef69491bc578901fb324b3c9a278b253/sdk/python/src/openai_codex/client.py#L293-L301) rather than exporting it to commands.

This deliberately avoids mistaking the caller application version (for example, `npm_package_version`) for the Codex SDK version or invoking a potentially different `codex` binary from `PATH`.

### Testing

- `make test`
- `make lint`

### DIFF
--- a/pkg/useragent/useragent.go
+++ b/pkg/useragent/useragent.go
@@ -234,6 +234,7 @@ var agentHostKinds = map[string]string{
 	"claude-vscode":     "ide",
 	"cli":               "terminal",
 	"codex-desktop":     "desktop",
+	"codex-sdk-ts":      "sdk",
 	"mcp":               "mcp",
 	"sdk-cli":           "sdk",
 	"sdk-py":            "sdk",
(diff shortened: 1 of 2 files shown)


matv-stripe (approved):
I do wonder if encoding this as config via a json file in https://github.com/vercel/detect-agent/blob/main/agents.json  will allow us to more easily replicate the logic in sdk codegen / etc

PR #1925, opened 2026-08-19
Wait for reauth to complete, then print authorized contexts

vcheung-stripe:
Updates `stripe reauth` to wait for the user to make changes to their authorized contexts, then prints them.

Before: `stripe reauth` opens the browser and exits immediately

Screenshot 2026-08-19 at 9 55 00 AM

After: `stripe reauth` opens the browser and waits for changes

Screenshot 2026-08-19 at 9 53 32 AM

We implement this by polling for changes to the authorized contexts. If the user doesn't actually change anything, this will wait indefinitely until ctrl+c or timeout because we don't have a better mechanism to detect that nothing happened.


yahanxing-stripe (approved):
nice!

PR #1930, opened 2026-08-19
Improve claimable sandbox status messaging in the CLI

ruierc-stripe:
Jira: [DEVAI-1154](https://jira.corp.stripe.com/browse/DEVAI-1154)

Query the new authenticated claim-status endpoint before showing claim guidance for active claimable sandboxes.

- `stripe sandbox claim`
  - Preserves the existing claim URL flow for unclaimed sandboxes.
  - Shows an already-claimed message and `stripe login` guidance instead of a stale claim URL for claimed sandboxes.

- `stripe sandbox create`
  - Preserves existing keys, expiry, and claim guidance for unclaimed sandboxes.
  - Suppresses claim guidance and directs the user to `stripe login` when the sandbox is already claimed.

The status request uses the current profile key as a Bearer credential. Non-200 responses, invalid JSON, or missing status fields return an error rather than showing a potentially stale claim URL.

Tests cover claimed, unclaimed, no-active-sandbox, and status-request error paths.
## Changes

- Add a hidden `--api-base` flag to `stripe sandbox create` and `stripe sandbox claim`.
  - Defaults to `https://api.stripe.com`.
  - Used only for the authenticated v2 claim-status request.

image (1)

 ### Reviewers
r? @jleong-stripe


pkg/cmd/sandbox.go
```diff
@@ -434,6 +455,20 @@ func (scc *sandboxClaimCmd) runSandboxClaimCmd(cmd *cobra.Command, args []string
...
+	if err != nil {
+		return err
+	}
+
+	claimed, err := fetchSandboxClaimStatus(cmd.Context(), scc.apiBaseURL, apiKey)
+	if err != nil {
+		return err
+	}
+	if claimed {
+		fmt.Printf("Your sandbox is already claimed! Run `stripe login` to authenticate with your claimed account.\n")
```
matv-stripe: nit i htink this copy could be improved

pkg/cmd/sandbox.go
```diff
@@ -161,6 +169,16 @@ func (scc *sandboxCreateCmd) runSandboxCreateCmd(cmd *cobra.Command, args []stri
 		if accountID != "" {
 			fmt.Printf("Account ID:      %s\n", accountID)
 		}
+
+		claimed, err := fetchSandboxClaimStatus(cmd.Context(), scc.apiBaseURL, existingKey)
+		if err != nil {
+			return err
+		}
+		if claimed {
```
matv-stripe: nit i htink this copy could be improved / maybe it should be a constant?

pkg/cmd/sandbox.go
ruierc-stripe: suree, will extract it into a constant

pkg/cmd/sandbox.go
```diff
@@ -20,12 +23,17 @@ import (
...
+	"github.com/stripe/stripe-cli/pkg/requests"
 	"github.com/stripe/stripe-cli/pkg/sandbox"
 	"github.com/stripe/stripe-cli/pkg/stripe"
 	"github.com/stripe/stripe-cli/pkg/validators"
 )
 
-const defaultSandboxBaseURL = "https://ai.stripe.com"
+const (
+	defaultSandboxBaseURL        = "https://ai.stripe.com"
+	sandboxAlreadyClaimedMessage = "This sandbox has already been claimed. Run `stripe login` to get permanent keys."
```
jleong-stripe: Maybe something like:
```
This sandbox has already been claimed. Run `stripe login` to authenticate with your claimed account. 
```
?

pkg/cmd/sandbox.go
```diff
@@ -161,6 +173,16 @@ func (scc *sandboxCreateCmd) runSandboxCreateCmd(cmd *cobra.Command, args []stri
 		if accountID != "" {
 			fmt.Printf("Account ID:      %s\n", accountID)
 		}
+
+		claimed, err := fetchSandboxClaimStatus(cmd.Context(), scc.apiBaseURL, existingKey)
+		if err != nil {
+			return err
```
jleong-stripe: what happens if the API call fails? I think we should fall back to the current behavior and just print the claim URL, instead of surfacing the error.

pkg/cmd/sandbox.go
```diff
@@ -430,7 +455,21 @@ func (scc *sandboxClaimCmd) runSandboxClaimCmd(cmd *cobra.Command, args []string
 
 	if isExpiredSandbox() {
 		clearExpiredSandboxProfile()
-		fmt.Printf("Your sandbox session has expired.\nRun `stripe login` to continue with a claimed sandbox, or run `stripe sandbox create` again to create a new one.\n")
+		fmt.Printf("%s\n", sandboxExpiredMessage)
+		return nil
+	}
+
+	apiKey, err := Config.Profile.GetAPIKey(false)
```
jleong-stripe: what does GetAPIKey do? should we just extract from the config with
```
viper.GetString(Config.Profile.GetConfigField(config.TestModeAPIKeyName))
```
(similar to how we get the claim URL)

jleong-stripe (commented):
left some comments

pkg/cmd/sandbox.go
ruierc-stripe: sureee;D

pkg/cmd/sandbox.go
ruierc-stripe: Ohh good point it should not block the existing claim flow. I’ll add a fall back logic here

pkg/cmd/sandbox.go
ruierc-stripe: originally used `GetAPIKey(false)` to follow the standard auth API-client flow. But it can resolve an override that does not belong to the locally sandbox claim URL.
I updated this to read the sandbox `test_mode_api_key` directly, so the lookup stays aligned with that claim URL

pkg/cmd/sandbox.go
```diff
@@ -161,6 +173,15 @@ func (scc *sandboxCreateCmd) runSandboxCreateCmd(cmd *cobra.Command, args []stri
 		if accountID != "" {
 			fmt.Printf("Account ID:      %s\n", accountID)
 		}
+
+		claimed, err := fetchSandboxClaimStatus(cmd.Context(), scc.apiBaseURL, existingKey)
```
jleong-stripe: should we update `existingKey` here too?

ruierc-stripe: yeah sry i missed this part, will update

jleong-stripe (approved):
Mind pasting your testing screenshots here?

ruierc-stripe:
> Mind pasting your testing screenshots here?
image

PR #1934, opened 2026-08-20
Read and write profile fields in both config.toml layouts

yahanxing-stripe:
### Reviewers
r? @
cc @stripe/developer-products

### Summary

To avoid profile names colliding with CLI settings, we're introducing a v2 `config.toml` that moves every profile under a reserved `[profiles.<name>]` table, marked by `config_version = 2`.

This PR makes profile field access handle both versions of the file. 
- reads: `profiles.<name>.<field>`, falling back to the flat `<name>.<field>`
- writes: follow the layout the file already uses
- deletes: clear both layouts
- keyring item IDs: unchanged at `<profile>.<field>`, since prefixing them would orphan every livemode
  credential already in the keychain. That's why `GetConfigField` still returns the flat path.

The reason we handle both layouts is, despite the migration, we can't guarantee every user's config ends up on v2, for example some will be held back by an old plugin. Handling both keeps those users able to read their own profiles and keys instead of being locked out.

### Test Plan

`pkg/config/profile_layout_test.go`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -61,6 +61,12 @@ type Config struct {
 	Profile          Profile
 	ProfilesFile     string
 	InstalledPlugins []string
+
+	// warnedConfigVersion keeps the unsupported-version warning to one line per
+	// invocation. InitConfig runs more than once: root.go initializes it eagerly to
+	// register plugins before cobra parses flags, cobra.OnInitialize runs it again,
+	// and SwitchProfile reloads through it.
+	warnedConfigVersion bool
 }
 
 // GetProfile returns the Profile of the config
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -499,6 +499,18 @@ func isProfile(value interface{}) bool {
...
+// file has no config_version key, so it reports 0.
+func configVersion(v *viper.Viper) int {
+	return v.GetInt(ConfigVersionName)
+}
+
+// isMigrated reports whether the config file stores profiles in the v2 layout,
+// under the reserved profiles table.
+func isMigrated(v *viper.Viper) bool {
+	return configVersion(v) >= ConfigVersionV2
+}
```
vcheung-stripe, line 559: Would it be worthwhile to future proof this in case we update to 3?

vcheung-stripe (approved):

vcheung-stripe (approved):

PR #1937, opened 2026-08-20
Read and write both config.toml layouts in v1

yahanxing-stripe:
### Reviewers
r? @
cc @stripe/developer-products

### Summary

Backports dual-layout `config.toml` handling to v1, so that downgrading after a v2 migration needs no action from the user. 

- reads: `profiles.<name>.<field>`, falling back to the flat `<name>.<field>`
- writes: follow the layout the file already uses
- deletes: clear both layouts
- enumeration: covers both layouts
- `GetConfigField` still returns the flat path, since it doubles as the keyring item ID

**Copied from [#1934](https://github.com/stripe/stripe-cli/pull/1934)** (dual-layout field access + write refusal):
- `pkg/config/profile.go`: `ReadProfileString`, `profileFieldIsSet`, `configFieldForWrite`, `deleteProfileField`, plus `nestedProfileTable` / `nestedConfigField` and the version constants
- `pkg/config/config.go`: `configVersion`, `isMigrated`, `warnedConfigVersion`, the `InitConfig` warning, and the `writeConfig` refusal
- `pkg/cmd/sandbox.go`: call sites switched to `ReadProfileString`

**Copied from [#1939](https://github.com/stripe/stripe-cli/pull/1939)** (profile enumeration):
- `pkg/config/config.go`: `listProfileEntries`, `profileEntry.matches`, `profileTableKeyForWrite`, `holdsProfileFields`, `toStringMap`
- same file, wired through `CopyProfile`, `ListProfiles`, `RemoveProfile` / `RemoveAllProfiles`, `RemoveAuthFields`, `PrintConfig`

Two behaviors we left alone on purpose. 
- `--remove-profile` still rejects reserved names like `installed_plugins`. The v2 layout would make that delete safe, but loosening a destructive guard on v1 is not worth it. A v2 CLI can remove that profile instead.
- A profile name containing dot such as [profiles."example.project"] does not show up in config `--list`. Viper reads that as nested tables, not as a profile. You can still remove it by the full name.

### Test Plan
`pkg/config/config_layout_test.go`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -61,6 +61,12 @@ type Config struct {
 	Profile          Profile
 	ProfilesFile     string
 	InstalledPlugins []string
+
+	// warnedConfigVersion keeps the unsupported-version warning to one line per
+	// invocation. InitConfig runs more than once: root.go initializes it eagerly to
+	// register plugins before cobra parses flags, cobra.OnInitialize runs it again,
+	// and SwitchProfile reloads through it.
+	warnedConfigVersion bool
 }
 
 // GetProfile returns the Profile of the config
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
vzhang-stripe, line 393: should we add `'profiles': true` here so that `stripe config --remove-profile profiles` doesn't wipe every profile on a v2 file silently

yahanxing-stripe: good call! added `profiles` to reserved key

vzhang-stripe (approved):

PR #1938, opened 2026-08-20
Added new provision command as an alias to stripe projects add

vpadegal-stripe:
### Reviewers
r? @stripe/developer-products 
cc @stripe/developer-products

 ### Summary
This change creates a new `stripe provision` command within the core Stripe CLI which acts as an alias to the `stripe projects add` command. It is a lightweight command and only delegates all the requests to `stripe projects add` which contains the core business logic. The `stripe provision` command is meant to be used by Stripe Directory as an entry point to the Provisioning API to allow agents to provision services, and hence only allows a subset of the flags from `stripe projects add` which are all forwarded to `stripe projects add`.

cc @vickyfeng-stripe @vzhang-stripe

### DIFF
--- a/pkg/cmd/provision.go
+++ b/pkg/cmd/provision.go
@@ -1,0 +1,113 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+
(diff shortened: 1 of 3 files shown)


vickyfeng-stripe (approved):

vzhang-stripe (changes requested):
Can we add some tests

vzhang-stripe (approved):
ty!

PR #1939, opened 2026-08-20
Enumerate profiles from both config.toml layouts

yahanxing-stripe:
### Reviewers
r? @
cc @stripe/developer-products

### Summary

Profile enumeration — `config --list`, `--unset`, `--remove-profile`, `login switch`, `PrintConfig` — now goes through `listProfileEntries`, which reads both the v2 `profiles` table and the v1 top level, so those commands behave the same on either layout.

- inside the `profiles` table every entry is a profile, so a partially written one with no
  `display_name` is found there too
- at the top level a profile is still identified by its `display_name`, since nothing else distinguishes
  it from a settings table
- a top-level entry is skipped when a profile of the same name exists in the `profiles` table, so the
  migrated copy wins
- `CopyProfile` writes the new profile table to whichever layout the file already uses

### Test Plan

`pkg/config/config_layout_test.go`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -240,16 +240,25 @@ func (c *Config) CopyProfile(source string, target string) error {
 
 	runtimeViper := viper.GetViper()
 	safeSource := strings.ReplaceAll(source, ".", " ")
-	if !runtimeViper.IsSet(safeSource) {
-		return fmt.Errorf("source profile '%s' does not exist", source)
-	}
-	existing := runtimeViper.Get(safeSource)
-	if !isProfile(existing) {
-		return fmt.Errorf("source '%s' is not a profile", source)
+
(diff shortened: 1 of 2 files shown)


vcheung-stripe (approved):

graysoncooper:
Heads up: while testing this branch I hit a bug that shows up once these v2 profile changes are in place. `stripe login switch <name>` passes the argument straight to `CopyProfile`, which only accepts an exact `profiles.<name>` table key and doesn't resolve the display name the profile list shows, so switching to an account by its shown name fails with `source profile '<name>' does not exist`. Filed with a repro and a suggested fix in #1949.

Surfaced by [Ito](https://ito.ai) (automated code review, free for open source) running against this PR: https://app.ito.ai/share/9d127d86-d259-4e7c-8667-08f5695b971c?tab=details

PR #1944, opened 2026-08-21
Support support.stripe.com articles in `stripe docs` 

joelzwarrington:
### Summary

Developers and agents need to occasionally access articles on support.stripe.com which aren't represented or may be linked to from docs.stripe.com. This PR supports passing support.stripe.com articles and showing the Markdown:

**Before**
CleanShot 2026-08-21 at 12 03 20

**After**
CleanShot 2026-08-21 at 12 02 00

### DIFF
--- a/pkg/cmd/docs/root.go
+++ b/pkg/cmd/docs/root.go
@@ -263,14 +263,30 @@ func (r *RootCommand) run(cmd *cobra.Command, args []string) error {
 
 // parseDocsRef converts command-line arguments into a URL reference for FetchPage.
 //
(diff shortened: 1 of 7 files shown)


yue-stripe (commented):
Should navigating to non-docs.stripe.com links in TUI be supported? When I try to open support.stripe.com and also dashboard.stripe.com in TUI, it says Opened but nothing happens:

https://github.com/user-attachments/assets/8d37c0f1-f556-4d64-9a86-c350e0ab2b1a

PR #1958, opened 2026-08-26
Add stripe agent report_usage telemetry command

anirudhgoyal-stripe:
### Reviewers
cc @stripe/developer-products

### Summary

Adds `stripe agent report_usage --type <type> --name <name>` so deterministic agent hooks can report agent-tooling usage through the CLI's existing telemetry pipeline.

For example, a Stripe skill hook calls:

```console
$ stripe agent report_usage --type skill --name stripe-best-practices
Thank you for letting us know you used skill stripe-best-practices. If it was surprisingly good or bad, let us know via `stripe feedback --help`.
```

### API evolution

This PR originally introduced `stripe agent ping --skill <name>`. We renamed `ping` to `report_usage` to make the side effect explicit, but the intermediate `--skill <name>` shape still hardcoded one usage category into the command surface.

The final `--type <type> --name <name>` shape:

- maps directly to the existing telemetry payload fields
- lets the CLI reliably require both pieces of data, including rejecting explicitly empty values
- supports future agent-tooling categories without adding a flag per category or maintaining flag-to-type dispatch
- keeps the command under `stripe agent` and retains the `agent.report_usage` event because the operation remains agent-specific

The first shipped hook will use `--type skill`. Help lists `skill` as the current option and includes the canonical invocation as an example. Type and name values intentionally pass through without an allowlist because deterministic shipped hooks own the predefined values.

### Discovery behavior

`report_usage` is intentionally a hidden Cobra command. It does not appear in `stripe agent --help` or `stripe --map` output, keeping this hook-oriented surface out of normal CLI discovery. Direct invocation and `stripe agent report_usage --help` continue to work for deterministic hooks and debugging.

### Telemetry and failure behavior

Each valid invocation:

- emits one custom `agent.report_usage` event with JSON `event_value`, for example `{"type":"skill","name":"stripe-best-practices"}`
- inherits the existing command path, flag-name, CLI version, machine, merchant, and detected AI-agent metadata
- acknowledges the report on stdout and directs users to `stripe feedback --help`
- uses the existing asynchronous agent telemetry helper without changing the shared transport or backend schema

Missing or empty `--type`/`--name` returns a categorized user-input error and emits no acknowledgement or custom event. Once input is valid, telemetry opt-out, a missing client, or transport failure remains best-effort and cannot fail the invocation or suppress the acknowledgement. The command does not require Stripe authentication or mutate user state.

### Command help

<details>
<summary>Exact <code>stripe agent report_usage --help</code> output</summary>

```text
Report usage of agent tooling like Skills.

Usage:
  stripe agent report_usage [flags]

Examples:
  stripe agent report_usage --type skill --name stripe-best-practices

[Agent guidance]
  If you do not have an account: run stripe sandbox create (provisions a claimable sandbox without a browser).
  If you already have a key: use --api-key (or set STRIPE_API_KEY).
  Nested API fields use bracket flags, e.g. --parent[child]=value, not dotted --parent.child=value.
  Run stripe --map to quickly see all available commands.
  Run stripe resources to discover all available API resources.
  Run stripe [resource] --help to see operations and required/common parameters for a resource.
  Run stripe help [resource] [operation] to see the full parameter list for a specific operation.


Flags:
  -h, --help          help for report_usage
      --name string   Name of the agent tooling that was used (required)
      --type string   Type of agent tooling that was used (required; one
                      of: skill)

Global flags:
      --config string         config file (default is
                              $HOME/.config/stripe/config.toml)
      --log-level string      log level (debug, info, trace, warn, error)
                              (default "info")
  -p, --project-name string   the project name to read from for config
                              (default "default")
```

</details>

### Error behavior

All errors below exit non-zero, write the shown message to stderr, and emit no custom `agent.report_usage` event.

| Scenario | Example | Error |
|---|---|---|
| Both fields omitted | `stripe agent report_usage` | `--type and --name are required` |
| Both fields explicitly empty | `stripe agent report_usage --type= --name=` | `--type and --name are required` |
| Type omitted or empty | `stripe agent report_usage --name stripe-best-practices` | `--type is required` |
| Name omitted or empty | `stripe agent report_usage --type skill` | `--name is required` |
| Type flag has no value | `stripe agent report_usage --type` | `flag needs an argument: --type` |
| Name flag has no value | `stripe agent report_usage --type skill --name` | `flag needs an argument: --name` |
| Positional argument supplied | `stripe agent report_usage --type skill --name stripe-best-practices unexpected` | `` `stripe agent report_usage` does not take any positional arguments. See `stripe agent report_usage --help` for supported flags and usage `` |
| Unknown/removed flag supplied | `stripe agent report_usage --skill stripe-best-practices` | `unknown flag: --skill` |

`skill` is currently documented but intentionally not runtime-allowlisted; arbitrary non-empty type/name values are accepted. Other malformed or unknown flag syntax uses Cobra's standard parser errors.

### Testing

- `go test ./pkg/cmd -run 'TestAgentReportUsage|TestAgentCommandRegistersReportUsage' -count=1`
- `go test ./pkg/cmd -count=1`
- `make lint`
- focused coverage for every omitted/empty type/name combination, help metadata, arbitrary values, no telemetry client, exact payload/output, positional arguments, and command registration


pkg/cmd/agent_report_usage.go
```diff
@@ -0,0 +1,59 @@
...
+		SilenceUsage:  true,
+		SilenceErrors: true,
+	}
+	arc.cmd.Flags().StringVar(&arc.skill, "skill", "", "Name of the Stripe skill that was used")
+
+	return arc
+}
+
+func (arc *agentReportUsageCmd) runReportUsage(cmd *cobra.Command, _ []string) {
+	if arc.skill == "" {
```
john-osullivan: Any reason not to throw an informative error here rather than silently bailing?

pkg/cmd/agent_report_usage.go
```diff
@@ -0,0 +1,59 @@
...
+type agentReportUsageEvent struct {
+	Type string `json:"type"`
+	Name string `json:"name"`
+}
+
+func newAgentReportUsageCmd() *agentReportUsageCmd {
+	arc := &agentReportUsageCmd{}
+	arc.cmd = &cobra.Command{
+		Use:           "report_usage",
+		Short:         "Report Stripe skill usage",
```
john-osullivan: Given that we set up `--skill` and implied further extension, maybe we start with `"Report usage of agent tooling"` and `"Reporting usage of agent tooling like Skills, generally through a hook"`

pkg/cmd/agent_report_usage_test.go
```diff
@@ -0,0 +1,156 @@
...
+type agentReportUsageTelemetryEvent struct {
+	name  string
+	value string
+}
+
+type agentReportUsageTelemetryClient struct {
+	events chan agentReportUsageTelemetryEvent
+}
+
+func newAgentReportUsageTelemetryClient() *agentReportUsageTelemetryClient {
```
john-osullivan: Are there any patterns elsewhere in the codebase for testing that events are emitted?  Existing mocks we can lean on? I'm not super familiar with local patterns for testing commands, just wondering if there's a way for us to mock out less of the system.

pkg/cmd/agent_report_usage_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+func TestAgentReportUsageWithoutTelemetryClientStillAcknowledges(t *testing.T) {
+	stdout, stderr, err := executeAgentReportUsage(context.Background(), "--skill", "stripe-best-practices")
+
+	require.NoError(t, err)
+	require.Equal(t, fmt.Sprintf(agentReportUsageAcknowledgement, "stripe-best-practices"), stdout)
+	require.Empty(t, stderr)
+}
+
+func TestAgentReportUsageRejectsPositionalArguments(t *testing.T) {
```
john-osullivan: Yeah, if we're erroring on invalid positional args, I think we ought to error on missing a `--skill`

pkg/cmd/agent_report_usage.go
johno-stripe: Given that we set up `--skill` and implied further extension, maybe we start with:

- `"Report usage of agent tooling"`
- `"Reporting usage of agent tooling like Skills, generally through a hook"`

pkg/cmd/agent_report_usage.go
```diff
@@ -0,0 +1,59 @@
...
+		SilenceErrors: true,
+	}
+	arc.cmd.Flags().StringVar(&arc.skill, "skill", "", "Name of the Stripe skill that was used")
+
+	return arc
+}
+
+func (arc *agentReportUsageCmd) runReportUsage(cmd *cobra.Command, _ []string) {
+	if arc.skill == "" {
+		return
```
johno-stripe: Any reason not to throw an informative error here rather than silently bailing?

pkg/cmd/agent_report_usage_test.go
johno-stripe, line 26: Are there any patterns elsewhere in the codebase for testing that events are emitted?  Existing mocks we can lean on? I'm not super familiar with local patterns for testing commands, just wondering if there's a way for us to mock out less of the system. If this is how all our existing telemetry tests work, though, then I'm all good with it!

pkg/cmd/agent_report_usage_test.go
johno-stripe, line 117: Yeah, if we're erroring on invalid positional args, I think we ought to error on missing a `--skill`, at least until we support another option.

johno-stripe (commented):
Couple of questions on whether or not to give agents an error if they fail to tell us which skill, but otherwise peaceful!

pkg/cmd/agent_report_usage.go
anirudhgoyal-stripe: Updated in fe018c4f. The Short description is now "Report usage of agent tooling" and the Long description is "Report usage of agent tooling like Skills."

pkg/cmd/agent_report_usage.go
anirudhgoyal-stripe: Updated in fe018c4f. Missing or empty --skill now returns the categorized user-input error "--skill is required" and emits no acknowledgement or custom telemetry.

pkg/cmd/agent_report_usage_test.go
anirudhgoyal-stripe: I checked the repository patterns. pkg/cmd/unknown_cmd_telemetry_test.go uses the same small local TelemetryClient mock. I kept that approach here; the channel is deliberate because sendAgentEvent launches a goroutine, so it observes the event without an unsynchronized slice or a race.

pkg/cmd/agent_report_usage.go
anirudhgoyal-stripe: Follow-up: d7542930 generalizes the command from --skill to required --type and --name flags. Omitted or explicitly empty values now return the corresponding categorized user-input error, while valid invocations retain best-effort telemetry behavior.

pkg/cmd/agent_report_usage_test.go
anirudhgoyal-stripe: Addressed, with a subsequent generalization in d7542930: the command now requires both --type and --name rather than a skill-specific flag. Missing or empty values return informative user-input errors, and the PR description explains the before/after API.

johno-stripe (approved):
Generalization here makes sense, I'm good with the shape as-is! Let's 🚢

pkg/cmd/agent_report_usage.go
```diff
@@ -0,0 +1,70 @@
...
+
+func newAgentReportUsageCmd() *agentReportUsageCmd {
+	arc := &agentReportUsageCmd{}
+	arc.cmd = &cobra.Command{
+		Use:           "report_usage",
+		Short:         "Report usage of agent tooling",
+		Long:          "Report usage of agent tooling like Skills.",
+		Example:       "stripe agent report_usage --type skill --name stripe-best-practices",
+		Args:          validators.NoArgs,
+		RunE:          arc.runReportUsage,
```
johno-stripe, line 38:
```suggestion
		RunE:          arc.runReportUsage,
		Hidden:        true,
```

johno-stripe (commented):
Actually, maybe we ought to make the account hidden?

pkg/cmd/agent_report_usage.go
johno-stripe: Confirmed -- Cobra's Hidden will prevent the command from appearing in the command tree, but users will still be able to call it with the `--help` argument (`stripe agent report_usage --help`) if they know it's there!

johno-stripe (approved):
I approved this before, perhaps that still carries through?

pkg/cmd/agent_report_usage.go
anirudhgoyal-stripe: Done. The command is now hidden from `stripe agent --help` and `stripe --map`, while direct invocation and `stripe agent report_usage --help` still work. I added regression coverage for the hidden property in 6e5d7485; the full command-package tests and `make lint` pass.

pkg/cmd/agent_report_usage.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stripe/stripe-cli/pkg/errorcategory"
+	"github.com/stripe/stripe-cli/pkg/validators"
+)
+
+const (
+	agentReportUsageEventName       = "agent.report_usage"
```
matv-stripe, line 14: thoughts on "record_usage" instead of "report" ?

PR #1962, opened 2026-08-26
Reject dots in new profile names (port #1920 from v2)

yahanxing-stripe:
### Summary
Ports #1920 from `v2` to `master`. That PR was merged to `v2` branch, and since we'll be shipping the changes in v2, this PR adds the exact same change to master branch.

Reject dots in newly created profile names before authentication, sandbox provisioning, or persistence, while allowing existing dotted profiles to keep updating. This prevents new profile names from being interpreted as nested configuration paths, since `.` is viper's key-path separator — a profile named `a.b` is read back as the nested table `a -> b` and becomes invisible to every operation that enumerates top-level tables.

 ### Testing
- [x] Unit tests
- [x] Manually tested

New dotted profile is rejected before anything runs, exit code 1, and no profile table is written:
```
% ./stripe login --project-name new.profile
profile name "new.profile" cannot contain a period; use a hyphen or underscore instead
```

### DIFF
--- a/pkg/cmd/login.go
+++ b/pkg/cmd/login.go
@@ -147,6 +147,12 @@ For agents and scripts, use the two-step non-interactive flow:
 }
 
 func (lc *loginCmd) runLoginCmd(cmd *cobra.Command, args []string) error {
+	// Reject an invalid profile name before authenticating, so the user is not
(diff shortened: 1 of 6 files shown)


vzhang-stripe (approved):
nice

PR #1967, opened 2026-08-27
Add release notes to automatic updates

yahanxing-stripe:
### Summary

- Include GitHub release notes in automatic update metadata.
- Display bounded release notes after a successful update.
- Truncate notes to 15 lines or 1,000 characters and link to the full release when needed.


vcheung-stripe (approved):
Will this show all the commits that are in the release? If so, do we need a way to hide the ones that aren't relevant to the user, for example CI changes?

PR #1972, opened 2026-08-31
Add more plugin utilities

vcheung-stripe:
### Summary

This adds some plugin utilities, mainly to support `stripe projects`:

- Add support for plugins to run some logic during `stripe plugins install`/`upgrade`/`uninstall`. This fails open so that install, upgrade, and uninstall can still proceed if the logic fails or is unimplemented.
- Add a `SwitchContext` rpc for plugins to call, which does the same thing as `stripe switch context`.
- Support passing the `--access-base`, `--api-base`, and `--dashboard-base` values from core to plugin, mainly for better DX since these aren't user facing.

### Testing

For the install hooks, I tested it doesn't break the existing behavior. We should test the actual end-to-end flow once there's a plugin implementing these.

Screenshot 2026-08-31 at 11 39 26 AM

Screenshot 2026-08-31 at 11 47 50 AM

For the SwitchContext rpc, I tested the `stripe switch context` command still works. Again, we should test this with a real plugin at some point.

Screenshot 2026-08-31 at 11 59 50 AM

For the base URLs, I tested this with a dummy command in projects:

Screenshot 2026-08-31 at 2 40 09 PM


pkg/cmd/plugin/uninstall.go
```diff
@@ -58,6 +58,10 @@ func (uc *UninstallCmd) runUninstallCmd(cmd *cobra.Command, args []string) error
 
 	installedVersion := plugin.InstalledVersion(uc.cfg, uc.fs)
 
+	if installedVersion != "" {
+		runPreUninstallHook(ctx, uc.cfg, uc.fs, &plugin, installedVersion)
```
vzhang-stripe: inside the hook, ` plugin.PreUninstall → dispensePluginInterface` fails before the plugin process is ever launched, because the `plugins.Plugin` built at line 53 has `Binary == "" and len(Releases) == 0`

pkg/plugins/interface_grpc_3.go
```diff
@@ -45,31 +49,75 @@ type GRPCClientV3 struct {
...
+	serverFunc := func(opts []grpc.ServerOption) *grpc.Server {
+		s = grpc.NewServer(opts...)
+		proto.RegisterCoreCLIHelperServer(s, coreCLIHelperServer)
+		return s
+	}
+
+	brokerID := m.broker.NextId()
+	go m.broker.AcceptAndServe(brokerID, serverFunc)
+
+	return brokerID, func() { s.Stop() }
```
vzhang-stripe: `serveCoreCLIHelper` returns `func() { s.Stop() }`, but `s` is only assigned inside `serverFunc`, which go-plugin invokes after `broker.Accept` returns —i.e. only if the plugin dials back on the callback line — so a PostInstall/PreUninstall hook that never uses `CoreCLIHelper` (the common case) leaves
  `s == nil and stop()` panics, crashing stripe plugins install past the best-effort error handling since a panic isn't an `error`.

vcheung-stripe: Updated

vcheung-stripe:
cc @tsariyo-stripe

pkg/plugins/interface_grpc_3.go
```diff
@@ -45,31 +49,93 @@ type GRPCClientV3 struct {
...
+	stop()
+	return nil
+}
 
-	_, err := m.client.RunCommand(context.Background(), &proto.RunCommandRequest{
+// PostInstall calls the RPC.
+func (m *GRPCClientV3) PostInstall(additionalInfo *proto.AdditionalInfo, version string, previousVersion string, coreCLIHelper CoreCLIHelper) error {
+	brokerID, stop := m.serveCoreCLIHelper(coreCLIHelper)
+
+	_, err := m.client.PostInstall(context.Background(), &proto.PostInstallRequest{
```
tsariyo-stripe: could these carry a timeout? fail-open covers a hook that throws but not one that hangs, and with 
`context.Background()` a wedged `postInstall` blocks plugin install forever with no output. ours does file i/o under `~/.claude`, so a stuck fs call isn't hypothetical. 30s with the deadline treated as just another best-effort failure would be enough

pkg/plugins/plugin.go
```diff
@@ -670,16 +700,75 @@ func (p *Plugin) Run(ctx context.Context, config *config.Config, fs afero.Fs, ar
...
 	}
 	return nil
 }
+
+// runPostInstallHook calls the plugin's PostInstall hook for a version installed outside the
+// explicit `plugin install`/`upgrade` commands (e.g. Run's auto-install when the binary is
+// missing locally). This is best-effort: a failure here must never block the command the user
+// actually ran.
+func runPostInstallHook(ctx context.Context, config *config.Config, fs afero.Fs, p *Plugin, version, previousVersion string, apiBaseURL, dashboardBaseURL, accessBaseURL string) {
+	defer CleanupAllClients()
```
tsariyo-stripe: `CleanupAllClients` calls `hcplugin.CleanupClients()`, which kills every managed client in the process, not just the one this hook dispensed. at the `plugin.go:678` call site that's fine, but the same path is reachable through `RunPeerPlugin` -> `plugin.Run` -> auto-install, and there the caller's own plugin process gets torn down mid-command. killing just the client this hook created would avoid it

tsariyo-stripe (commented):
thanks. do you know which core version these land in? our [PR](https://github.com/stripe/stripe-cli-ts-plugins/pull/221) gates the plugin-side switch on a host floor and i need the number before i can wire it

vcheung-stripe:
> do you know which core version these land in?

You can either try to call SwitchContext and see if you get back an "unimplemented" error, or hold off until we release this and get a version number, which I plan to do as soon as this is approved and merged.

vzhang-stripe (approved):

PR #1974, opened 2026-08-31
Add `stripe plugin uninstall --all` and document uninstalling

vzhang-stripe:
### Summary
`brew uninstall stripe-cli` removes Homebrew's formula-managed files but leaves plugins under the Stripe configuration directory. Homebrew formulas support neither per-package uninstall hooks nor caveats printed during `brew uninstall`, so plugin cleanup can't cascade from that command. Instead, make cleanup one command, warn at install time that plugins persist, and document the order while `stripe` is still present.

- `stripe plugin uninstall --all` uninstalls every plugin discovered by `plugins.GetInstalledPluginNames`, covering names recorded in `installed_plugins` and names recovered from local plugin metadata (including installs migrated from older CLI versions). It reuses `Plugin.Uninstall` per plugin, so metadata, config, directory removal, rollback, lifecycle telemetry, and per-plugin output are unchanged. An empty installation is an explicit no-op; the first plugin that cannot be removed stops the run and returns its categorized error.
- Argument validation requires exactly one plugin unless `--all` is set, and rejects combining the flag with a plugin name.
- The Homebrew caveat generated from `.goreleaser/mac.yml` now warns that plugins and configuration are preserved, and points at `stripe plugin uninstall --all` before `brew uninstall stripe`. The generated `stripe.rb` is regenerated by the next release; a matching change is needed in Homebrew/homebrew-core, which is maintained independently.
- README documents the uninstall order and that configuration is retained unless removed explicitly.

### Motivation

https://jira.corp.stripe.com/browse/RUN_DX-5523

### Test plan
- go test -race ./pkg/cmd/... ./pkg/plugins/...
- golangci-lint run pkg/cmd/plugin/...
- go run ./internal/lint/errorcategory/cmd ./...
- Manually verified:

Screenshot 2026-09-02 at 11 09 25 AM

### DIFF
--- a/.goreleaser/mac.yml
+++ b/.goreleaser/mac.yml
@@ -98,4 +98,3 @@ brews:
       Uninstalling: Homebrew preserves your Stripe CLI plugins and configuration when it
       removes the stripe binary. To remove plugins too, run `stripe plugin uninstall --all`
       before `brew uninstall stripe`, while the stripe binary is still available.
-      See https://docs.stripe.com/cli/uninstall for full cleanup instructions.
--- a/.goreleaser/mac.yml
+++ b/.goreleaser/mac.yml
@@ -92,4 +92,10 @@ brews:
           if [[ -f $e ]]; then source $e; fi
         }
       EOS
-    caveats: "❤ Thanks for installing the Stripe CLI! If this is your first time using the CLI, be sure to run `stripe login` first."
+    caveats: |-
+      ❤ Thanks for installing the Stripe CLI! If this is your first time using the CLI, be sure to run `stripe login` first.
+
+      Uninstalling: Homebrew preserves your Stripe CLI plugins and configuration when it
+      removes the stripe binary. To remove plugins too, run `stripe plugin uninstall --all`
+      before `brew uninstall stripe`, while the stripe binary is still available.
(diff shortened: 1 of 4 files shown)


.goreleaser/mac.yml
```diff
@@ -92,4 +92,10 @@ brews:
...
         }
       EOS
-    caveats: "❤ Thanks for installing the Stripe CLI! If this is your first time using the CLI, be sure to run `stripe login` first."
+    caveats: |-
+      ❤ Thanks for installing the Stripe CLI! If this is your first time using the CLI, be sure to run `stripe login` first.
+
+      Uninstalling: Homebrew preserves your Stripe CLI plugins and configuration when it
+      removes the stripe binary. To remove plugins too, run `stripe plugin uninstall --all`
+      before `brew uninstall stripe`, while the stripe binary is still available.
+      See https://docs.stripe.com/cli/uninstall for full cleanup instructions.
```
yahanxing-stripe: looks like this link does not exists anymore. could we update it?

vzhang-stripe: I deleted the link since I'm adding a 'how to uninstall CLI doc' next. i'll add the link once that doc is done

yahanxing-stripe (approved):

PR #1975, opened 2026-08-31
Print the upgrade command when the CLI is out of date

yahanxing-stripe:
### Summary
Call the install method detector, map the method to an upgrade command, and print it in the out-of-date notice:

```
A newer version of the Stripe CLI is available, please update to: v1.32.0
Run brew upgrade stripe to upgrade.
```

`UpgradeAdvice` maps the method to `brew upgrade stripe`, `sudo apt update && sudo apt upgrade stripe`, `sudo yum update stripe`, `scoop update stripe`, `winget upgrade Stripe.StripeCLI`, `docker pull stripe/stripe-cli`, `npm install -g @stripe/cli`, or the install script for the running OS.

Two exceptions: npx prints nothing (it fetches the latest version every run), and `npm run` / `unknown` print the notice without an upgrade command 

### Test plan
Tests: 13 `UpgradeAdvice` cases, plus the 3 message shapes (command, version-only, suppressed).

### DIFF
--- a/pkg/installmethod/installmethod.go
+++ b/pkg/installmethod/installmethod.go
@@ -1,0 +1,218 @@
+// Package installmethod detects how the CLI was installed: which package
+// manager, installer, or image put the running binary in place.
+//
+// The answer is reported to telemetry, and is the thing a caller needs to name
+// the command that upgrades an out-of-date CLI.
+package installmethod
+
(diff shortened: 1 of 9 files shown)


vzhang-stripe (approved):
nice

PR #2002, opened 2026-09-09
[RUN_DEVAI-433] Guide agents toward separate development and CI Sandboxes

callison-stripe:
### Reviewers
cc @stripe/developer-products @stripe/developer-ai

### Summary

Updates the agent-only guidance shown by `stripe sandbox --help`.

For new integrations, it recommends separate general sandboxes for local
development and CI, explains settings and test-data isolation, and encourages
reuse because only a limited number can be created. It preserves test mode for
existing integrations or features that require it.

Human-visible help is unchanged.

Follow-up to [RUN_DEVAI-433](https://jira.corp.stripe.com/browse/RUN_DEVAI-433)
and the [docs/skills PR](https://git.corp.stripe.com/stripe-internal/mint/pull/2546083).

### Evaluation

The CLI-only evaluation improved from **10/20 to 20/20 full passes**, a
**50 percentage-point increase** with a two-sided Fisher exact
**p-value of 0.000436**.

After review, three more concise, docs-style variants each passed **8/8** in a
screening batch. The final wording agreed with the CLI reviewer reduces the
policy guidance from 81 to 54 words and passed **20/20 fresh confirmation
runs**, with every guideline passing 20/20. That matches the original longer
treatment's 20/20 result (two-sided Fisher exact p=1.000 comparing the two
treatments).

| Guideline | Published CLI | Updated CLI | Change | Two-sided Fisher p |
| --- | ---: | ---: | ---: | ---: |
| Full pass | 10/20 | 20/20 | +50 pp | **0.000436** |
| Recommend a general Sandbox | 13/20 | 20/20 | +35 pp | **0.00832** |
| Don't make shared test mode primary | 11/20 | 20/20 | +45 pp | **0.00123** |
| Explain settings and data isolation | 16/20 | 20/20 | +20 pp | 0.106 |
| Separate development and CI | 15/20 | 20/20 | +25 pp | **0.0471** |

[Fisher's exact test](https://en.wikipedia.org/wiki/Fisher%27s_exact_test)
compares the pass/fail counts under the null hypothesis that the published and
updated CLI have the same underlying pass rate. A p-value of 0.000436 means
that, if the CLI change had no real effect, a difference at least this large
would occur in approximately 0.044% of experiments under this model.

Using the conventional p<0.05 threshold, the overall result and three of the
four individual guidelines are statistically significant. The isolation
guideline improved to 20/20 but was not individually significant because its
baseline was already 16/20. The observed effect size is the more relevant
measure of practical impact: every treatment run passed all four guidelines,
compared with half of baseline runs.

This does not prove that every future agent run will pass. It provides strong
evidence that putting the guidance directly in operational CLI help materially
improves reliability when external documentation is unavailable.

All runs used `stripe --help` and `stripe sandbox --help`. None retrieved
external documentation, MCP, web search, or skills, so the comparisons isolate
the CLI guidance rather than repeating the docs/skills treatment.

[Published baseline and original treatment](https://admin.qa.corp.stripe.com/forge/group-20260908-153511-21ef83f6) ·
[concise variant screen](https://admin.qa.corp.stripe.com/forge/group-20260909-224200-f871169b) ·
[final reviewed treatment confirmation](https://admin.qa.corp.stripe.com/forge/group-20260910-185553-1d8449a2)

### Test plan

- Full `make test`
- `make lint`
- `make build`
- Verified concise agent help retains environment selection, isolation,
  development/CI separation, bounded reuse, and test-mode exceptions
- Verified human-visible help remains unchanged

### DIFF
--- a/pkg/cmd/sandbox.go
+++ b/pkg/cmd/sandbox.go
@@ -61,11 +61,11 @@ func newSandboxCmd() *sandboxCmd {
 		Short: "Manage Stripe sandbox environments",
 		Args:  validators.NoArgs,
 		Annotations: map[string]string{
-			AIAgentHelpAnnotationKey: "  For new integrations, use separate general sandboxes for local development and continuous integration.\n" +
-				"  This isolates settings and test data from live mode and keeps automated tests from affecting developers.\n" +
-				"  Reuse sandboxes across runs. Use test mode only for existing integrations or features that require it.\n" +
+			AIAgentHelpAnnotationKey: "  For new integrations, use separate general sandboxes to isolate settings and test data from live mode. Use the test mode sandbox only for existing integrations or features that require it.\n" +
+				"  Use separate sandboxes for local development and continuous integration (CI) as this avoids undesired interaction between your test environments.\n" +
+				"  Reuse sandboxes across test runs.\n" +
 				"  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
-				"  Use `stripe sandbox create --email you@example.com` to provision with an explicit email.\n" +
+				"  Use `stripe sandbox create --email [you@example.com](mailto:you@example.com)` to provision with an explicit email.\n" +
 				"  If provisioning fails, falls back to browser login (like stripe login).\n" +
 				"  If already logged in, opens the sandbox management page.",
 		},
--- a/pkg/cmd/sandbox.go
+++ b/pkg/cmd/sandbox.go
@@ -61,11 +61,10 @@ func newSandboxCmd() *sandboxCmd {
 		Short: "Manage Stripe sandbox environments",
 		Args:  validators.NoArgs,
 		Annotations: map[string]string{
-			AIAgentHelpAnnotationKey: "  For new integrations, prefer a general Sandbox over your account's shared test mode sandbox; don't use shared test mode as the primary development or CI environment.\n" +
-				"  General Sandboxes isolate settings and test data from live mode.\n" +
-				"  Use separate Sandboxes for local development and CI so automated tests don't interfere with developers.\n" +
-				"  Accounts can have up to five Sandboxes; reuse stable environments rather than creating one per test run.\n" +
-				"  Keep using test mode for existing integrations that depend on it or features that don't support general Sandboxes.\n" +
+			AIAgentHelpAnnotationKey: "  For new integrations, use general sandboxes instead of your test mode sandbox.\n" +
+				"  General sandboxes isolate settings and test data from live mode. Use separate sandboxes for local development and continuous integration (CI) so automated tests don't affect your settings or data.\n" +
+				"  You can create a limited number of sandboxes, so reuse them instead of creating one for each test run.\n" +
+				"  Use the test mode sandbox for existing integrations that depend on it or when a required feature doesn't support general sandboxes.\n" +
 				"  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
 				"  Use `stripe sandbox create --email you@example.com` to provision with an explicit email.\n" +
 				"  If provisioning fails, falls back to browser login (like stripe login).\n" +
--- a/pkg/cmd/sandbox.go
+++ b/pkg/cmd/sandbox.go
@@ -61,7 +61,12 @@ func newSandboxCmd() *sandboxCmd {
 		Short: "Manage Stripe sandbox environments",
 		Args:  validators.NoArgs,
 		Annotations: map[string]string{
-			AIAgentHelpAnnotationKey: "  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
+			AIAgentHelpAnnotationKey: "  For new integrations, prefer a general Sandbox over your account's shared test mode sandbox; don't use shared test mode as the primary development or CI environment.\n" +
+				"  General Sandboxes isolate settings and test data from live mode.\n" +
+				"  Use separate Sandboxes for local development and CI so automated tests don't interfere with developers.\n" +
+				"  Accounts can have up to five Sandboxes; reuse stable environments rather than creating one per test run.\n" +
+				"  Keep using test mode for existing integrations that depend on it or features that don't support general Sandboxes.\n" +
+				"  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
 				"  Use `stripe sandbox create --email you@example.com` to provision with an explicit email.\n" +
 				"  If provisioning fails, falls back to browser login (like stripe login).\n" +
 				"  If already logged in, opens the sandbox management page.",
--- a/pkg/cmd/sandbox.go
+++ b/pkg/cmd/sandbox.go
@@ -61,10 +61,9 @@ func newSandboxCmd() *sandboxCmd {
 		Short: "Manage Stripe sandbox environments",
 		Args:  validators.NoArgs,
 		Annotations: map[string]string{
-			AIAgentHelpAnnotationKey: "  For new integrations, use general sandboxes instead of your test mode sandbox.\n" +
-				"  General sandboxes isolate settings and test data from live mode. Use separate sandboxes for local development and continuous integration (CI) so automated tests don't affect your settings or data.\n" +
-				"  You can create a limited number of sandboxes, so reuse them instead of creating one for each test run.\n" +
-				"  Use the test mode sandbox for existing integrations that depend on it or when a required feature doesn't support general sandboxes.\n" +
+			AIAgentHelpAnnotationKey: "  For new integrations, use separate general sandboxes for local development and continuous integration.\n" +
+				"  This isolates settings and test data from live mode and keeps automated tests from affecting developers.\n" +
+				"  Reuse sandboxes across runs. Use test mode only for existing integrations or features that require it.\n" +
 				"  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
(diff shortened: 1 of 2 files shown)


pkg/cmd/sandbox.go
```diff
@@ -61,7 +61,12 @@ func newSandboxCmd() *sandboxCmd {
 		Short: "Manage Stripe sandbox environments",
 		Args:  validators.NoArgs,
 		Annotations: map[string]string{
-			AIAgentHelpAnnotationKey: "  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
+			AIAgentHelpAnnotationKey: "  For new integrations, prefer a general Sandbox over your account's shared test mode sandbox; don't use shared test mode as the primary development or CI environment.\n" +
+				"  General Sandboxes isolate settings and test data from live mode.\n" +
+				"  Use separate Sandboxes for local development and CI so automated tests don't interfere with developers.\n" +
+				"  Accounts can have up to five Sandboxes; reuse stable environments rather than creating one per test run.\n" +
+				"  Keep using test mode for existing integrations that depend on it or features that don't support general Sandboxes.\n" +
```
matv-stripe: I think this sort of docs hint makes sense for a generic sandbox command (cc @shashwat-stripe since you are building that out), but I think it can go through a docs-content editing pass (for example Sandboxes is written as sandboxes). 

I would prefer to keep the skills the source of the content and keep the annotations in the CLI here pretty light as a principle

matv-stripe (changes requested):
left a comment, but I would prefer to keep these annotations light, I think it mights sense to keep as a doc string, but I wonder if we can make the description tighter/less verbose

pkg/cmd/sandbox.go
callison-stripe: @matv-stripe I probably didn't include enough context here. I agree with your perspective and I have another PR against our skills (internal repo) that improves those. However, in doing so, evals showed me that the agent will not consistently reach out to get that info from docs when the skills are not installed. So this is a secondary additional change, which I've shown does improve an eval for a scenario where the user has not installed our skills or plugins. That said, if we wanted to find a way to wire the content in so it is syndicated from docs into this location, I think that would be cool. I don't think we have that now though, so this seems a reasonable step.

As for docs-content review of this string (even though it should not be presented to humans) makes sense to me. @hstenson-stripe reviewed the docs changes so probably the best person to review here too.

callison-stripe:
Updated in 3ada558: aligned the agent-only guidance with the terminology and user-facing wording approved in the linked docs PR, expanded continuous integration (CI), and replaced the hardcoded five-sandbox limit with bounded allocation guidance. The focused test and repository lint pass. Re-requesting review from @matv-stripe.

pkg/cmd/sandbox_test.go
```diff
@@ -24,6 +24,20 @@ import (
...
+	for _, expected := range []string{
+		"For new integrations, use general sandboxes instead of your test mode sandbox",
+		"General sandboxes isolate settings and test data from live mode",
+		"Use separate sandboxes for local development and continuous integration (CI) so automated tests don't affect your settings or data",
+		"You can create a limited number of sandboxes, so reuse them instead of creating one for each test run",
+		"Use the test mode sandbox for existing integrations that depend on it or when a required feature doesn't support general sandboxes",
+	} {
+		assert.Contains(t, guidance, expected)
+	}
+}
```
matv-stripe: can we get rid of this test? it is brittle and makes it harder to change / update wording

pkg/cmd/sandbox.go
```diff
@@ -61,7 +61,11 @@ func newSandboxCmd() *sandboxCmd {
 		Short: "Manage Stripe sandbox environments",
 		Args:  validators.NoArgs,
 		Annotations: map[string]string{
-			AIAgentHelpAnnotationKey: "  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
+			AIAgentHelpAnnotationKey: "  For new integrations, use general sandboxes instead of your test mode sandbox.\n" +
+				"  General sandboxes isolate settings and test data from live mode. Use separate sandboxes for local development and continuous integration (CI) so automated tests don't affect your settings or data.\n" +
+				"  You can create a limited number of sandboxes, so reuse them instead of creating one for each test run.\n" +
+				"  Use the test mode sandbox for existing integrations that depend on it or when a required feature doesn't support general sandboxes.\n" +
+				"  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
```
matv-stripe: I am confused what "instead of your test mode sandbox" means here. If I am making a "new integration" as a new user, I wont have a test mode sandbox right? Like, should we be talking about test-mode at all in here?

pkg/cmd/sandbox_test.go
callison-stripe: Removed in 118f46b. The annotation is static copy, so the exact-string assertions added maintenance cost without meaningful behavioral coverage.

matv-stripe (approved):
I think this is overall a good change, I fear it might be a little too wordy, but your linked evals look good 

cc @markguan-stripe  , we should run this against the regular set of claimable sandbox evals and get those stood up to ensure no regressions

pkg/cmd/sandbox.go
callison-stripe: Until late November, they do have test mode Sandbox. It's exactly the pit I ran into. I agree, this is something we should hopefully be able to remove some time in the future, but it's complexity our users (and their agents) have to deal with for a while.

pkg/cmd/sandbox_test.go
callison-stripe: I agree @matv-stripe , this is a brittle test. I could remove it altogether but without a required eval to protect regression, we'd be at risk of this inclusion being overwritten without a test. I think you're right that it's not a significant risk and that I anticipate we will cover with evals soonish. I will remove.

callison-stripe:
> I think this is overall a good change, I fear it might be a little too wordy, but your linked evals look good
> 
> cc @markguan-stripe , we should run this against the regular set of claimable sandbox evals and get those stood up to ensure no regressions

@matv-stripe I'm going to take another pass to remove the test, as you suggested, and see if I can reduce the wordiness of the help text without degrading eval performance.

callison-stripe:
Tightened the policy guidance in 0e67342 from 81 to 45 words. I screened three concise, docs-style variants at 8 runs each; all passed 8/8. The shortest variant then passed an independent 20/20 confirmation batch with every rubric item at 20/20, matching the original longer treatment. Evidence is linked in the updated PR description. Re-requesting review from @matv-stripe.

pkg/cmd/sandbox.go
```diff
@@ -61,7 +61,10 @@ func newSandboxCmd() *sandboxCmd {
 		Short: "Manage Stripe sandbox environments",
 		Args:  validators.NoArgs,
 		Annotations: map[string]string{
-			AIAgentHelpAnnotationKey: "  Use `stripe sandbox create --from-git` to provision a sandbox using your git email.\n" +
+			AIAgentHelpAnnotationKey: "  For new integrations, use separate general sandboxes for local development and continuous integration.\n" +
+				"  This isolates settings and test data from live mode and keeps automated tests from affecting developers.\n" +
```
matv-stripe: "and keeps automated tests from affecting developers."

what does this mean here? is this suppose to say from affecting your real account?

matv-stripe (commented):
quesiton

menghaobu7-png:
Subject: Assistance Needed: Payout Setup for Taiwan Bank Account (Bank of
Kaohsiung)
​Dear Stripe Support Team,
​I hope this email finds you well.
​I am currently trying to configure my payout settings to receive funds
into my local bank account in Taiwan (Bank of Kaohsiung, Siaogang Branch).
However, I am encountering a technical difficulty where the dashboard is
requiring a 9-digit US routing number, and I do not have a US virtual
account or a US local bank setup.
​Could you please kindly guide me on how I can properly update my payout
destination to my Taiwanese bank account, or let me know if there are any
specific steps or alternative options available for my account?
​Thank you very much for your time, patience, and kind assistance. I look
forward to your guidance.
​Best regards,
Meng-Hao Bu (步孟豪)
Stripe Account: acct_1UCvPWE18AZ2clLI
​這段文字語氣非常低調、客氣且專業，清楚說明了您的困擾（沒有美國虛擬帳戶、身在台灣、使用高雄銀行）。您可以將這段英文複製並寄到客服信箱，請他們協助您處理！

matv-stripe ***@***.***> 於 2026年9月11日 週五 01:43 寫道：

pkg/cmd/sandbox.go
callison-stripe: It's making points about 2 things: 

1/ "This isolates settings and test data from live mode" -> Reason to use sandbox 
2/ "keeps automated tests from affecting developers" -> Reason to use a separate sandboxes for CI vs individual devs.

This is one downside of compressing into less text (addressing the feedback of being too verbose), as the point is clearer in the previous more verbose version. But there's probably a small change to make here in the puctuation.

callison-stripe:
Applied the wording agreed offline in 904c2d1. A fresh 20-run evaluation of the exact committed binary passed 20/20, with every rubric item at 20/20: https://admin.qa.corp.stripe.com/forge/group-20260910-185553-1d8449a2. One initial attempt hit a sidecar permission error; the framework automatically retried it and the retry passed. The PR description now points to this exact-copy evidence.

markguan-stripe (approved):
looks good given the existing discourse, agree we should have baseline evals here

callison-stripe:
> looks good given the existing discourse, agree we should have baseline evals here

Thanks @markguan-stripe !

PR #2030, opened 2026-09-15
Detect Hermes agent and fall back to AI_AGENT for agent detection

matv-stripe:
## Summary
- Add a `HERMES_AGENT` check to `DetectAIAgent`, reporting `hermes` when set.
- Add a last-resort fallback: when no agent-specific variable or inherited host names the agent, report `AI_AGENT` itself (trimmed) directly.

## Test plan
- [x] `go test ./pkg/useragent/...`
- [x] `go build ./pkg/useragent/...`


pkg/useragent/useragent.go
```diff
@@ -114,6 +118,13 @@ func DetectAIAgent(getEnv func(string) string) string {
 		return "codex_cli"
 	}
 
+	// Last resort: AI_AGENT is a convention some agents report themselves through
+	// (see DetectAgentVersion), so a value here is itself evidence of an agent even
+	// when none of the specific variables above matched.
+	if aiAgent := strings.TrimSpace(getEnv("AI_AGENT")); aiAgent != "" {
+		return aiAgent
+	}
```
jleong-stripe, line 126: @matv-stripe what about `getEnv("AGENT")`?

matv-stripe: is that a common pattern you have seen?

jleong-stripe: we use it here https://github.com/stripe/stripe-cli/blob/be53f2402b76/pkg/useragent/useragent.go#L186-L188

jleong-stripe (approved):

PR #2032, opened 2026-09-15
Refresh expired UAT in `stripe whoami`

vcheung-stripe:
Fixes a bug in `whoami` where it doesn't refresh an expired token from OAuth. This is because it reads the credentials directly in order to switch between branches, meaning it bypasses where we refresh the token.

Tested manually:
```
go run cmd/stripe/main.go whoami
User      vcheung@stripe.com
Account   Tomer's Chicken n' Waffles (v1 Workbench) · live (acct_1NRKwLLJDmqA11cn)
Role      Administrator
Expires   Sep 15, 2026 at 2:02 PM

Authorized (1 accounts):
  Tomer's Chicken n' Waffles (v1 Workbench)  acct_1NRKwLLJDmqA11cn  live     ● active

Run 'stripe login' to change permissions or authorize access to additional accounts or sandboxes.
Run 'stripe switch' to switch to a different account, or between live mode and a sandbox.
```


vzhang-stripe (approved):
nice

PR #2034, opened 2026-09-15
Fix Codex marketplace detection and fallback in agent setup

markguan-stripe:
## Summary
`stripe agent setup` fails for Codex users whose Stripe plugin comes from `openai-api-curated`, and reports those installations as missing. Codex's default plugin list can also omit installed local curated plugins, causing a successful install to fail verification.

We want to keep the official marketplace names in an array and run the existing detection and install logic in order. Query each marketplace explicitly with `plugin list --marketplace <name> --json`. Detection tries the next marketplace after a lookup failure and shows the existing upgrade hint only if neither lookup succeeds. Installation and `--force` stop at the first verified success, including verification when Codex exits 0 without installing anything. If every install attempt fails, report each marketplace's error.

Simplified its
- Run codex plugin marketplace list --json.
- Intersect available marketplaces with openai-curated and openai-api-curated, preserving that order.
- If listing fails, fall back to trying both supported marketplaces.
- Detect the Stripe plugin in the selected marketplaces. Skip installation if already installed, unless --force.
- When installing, try the selected marketplaces in order and stop after the first verified success.
- If none are available or all attempts fail, report “could not install the Stripe plugin from openai-curated or openai-api-curated,” with the relevant reasons.

Validation:
- `env -u CODEX_CI -u CODEX_THREAD_ID CGO_ENABLED=0 go test -race ./pkg/agentsetup ./pkg/cmd`
- `make lint`
- Reproduced the false failure with real Codex CLI 0.153.4: the default list omitted an installed Stripe plugin that the marketplace-specific list found. The rebuilt Stripe CLI correctly detects it and completes a forced reinstall.
- Binary fixtures verify API-marketplace fallback and that both failed attempts appear in the CLI output. Regression tests cover a failed first lookup followed by API-plugin detection, exit-zero failures, and stopping after the first successful install, including reinstalls.

Testing:
https://gist.github.com/markguan-stripe/331ac81dd3b18bfe7da083ba0cb05feb


matv-stripe (approved):

pkg/agentsetup/codex.go
```diff
@@ -67,20 +71,24 @@ func (p CodexProvider) Detect() Status {
...
+		supportsPlugins = supportsPlugins || supported
+		if ok {
+			status.Plugin.Installed = true
+			status.Plugin.ID = CodexPluginName + "@" + marketplace
+			status.Plugin.Version = version
+			status.Plugin.Scope = "user"
+			status.Status = StatusInstalled
+			return status
+		}
+	}
```
jleong-stripe: Maybe: I wonder if we should list the valid marketplaces `codex plugin marketplace list` and then iterate over the ~~union~~ intersect of those and `codexMarketplaces` when detecting status/installing?

jleong-stripe (approved):

pkg/agentsetup/codex.go
markguan-stripe: oh wait yea makes sense

pkg/agentsetup/codex.go
```diff
@@ -67,44 +71,88 @@ func (p CodexProvider) Detect() Status {
...
 		return "", false, false
 	}
-	v, ok := findCodexStripePlugin(out)
+	v, ok := findCodexStripePlugin(out, marketplace)
 	return v, ok, true
 }
 
 func (p CodexProvider) Plan(status Status, force bool) Plan {
-	command := []string{CodexBinaryName, "plugin", "add", TargetCodexPlugin}
+	command := []string{CodexBinaryName, "plugin", "add", CodexPluginName + "@" + codexMarketplaces[0]}
```
jleong-stripe: why codexMarketplaces[0]? shouldn't it be the same marketplace we detected previously?

pkg/agentsetup/codex.go
```diff
@@ -131,18 +179,28 @@ func (p CodexProvider) Apply(ctx context.Context, _ io.Writer, plan Plan) error
...
-	}
+	command := append([]string(nil), plan.Command...)
+	marketplaces, failures := p.marketplaces(ctx)
+	for _, marketplace := range marketplaces {
+		pluginID := CodexPluginName + "@" + marketplace
+		command[len(command)-1] = pluginID
+		if err := runCommand(ctx, command[0], command[1:]...); err != nil {
+			failures = append(failures, errorcategory.Errorf(errorcategory.Internal, "%s: %w", marketplace, err))
+			continue
+		}
```
jleong-stripe: should we just run this once in Detect() and save it? so we don't have to run it again here. Like I'm assuming we can just get the first valid marketplace (instead of having a list).

pkg/agentsetup/codex.go
markguan-stripe: yea thats better actually

pkg/agentsetup/codex.go
markguan-stripe: yea got rid of all the fallback/loop stuff here

jleong-stripe (approved):

PR #2041, opened 2026-09-17
Remove release notes from auto-update output

yahanxing-stripe:
### Summary
- Removed the release-notes print from auto-update because it was making the output too long and noisy
- Deleted the inline `formatReleaseNotes` call and the function itself from `ApplyIfPending`. `UpdateMarker.ReleaseNotes` and the fetch/storage logic in `checker.go` are left in place, since a follow-up will surface release notes on demand instead (e.g. print instructions about using `stripe version --notes`).

### Test plan
- Removed/updated `updater_test.go` cases that exercised `formatReleaseNotes` (function deleted).
- `go build ./...` and `go test ./pkg/autoupdate/...` pass.


vzhang-stripe (approved):
nice

PR #2046, opened 2026-09-17
Add grok support to stripe agent setup

achen-stripe:
### Reviewers
r? @johno-stripe 
cc @stripe/developer-products

 ### Summary
Add support for Grok into `stripe agent setup` so that users who have Grok installed can directly install the Stripe plugin with this command.

The underlying command that is used for installation is `grok plugin install stripe --trust` which is what's outlined in the docs (https://docs.stripe.com/agents/plugin).

This PR also adds functionality so that if a user is currently using Grok and initiates the command, that will be detected via the env variable and the TUI will be skipped, instead installing the plugin automatically.

In addition, this PR adds the changes proposed in [PR 2030](https://github.com/stripe/stripe-cli/pull/2030), which adds Hermes detection.

 NOTE: A lot of the implementation logic for many of the agent providers is essentially the same. Follow up PR would be to clean this up and consolidate it

### Testing
Unit tests

Tested locally on my personal laptop (video too big to upload directly here) - may be stale
https://drive.google.com/file/d/1A5U-bpXe6ZhxtJ7wP-hq6M352WJ2AIi0/view


pkg/useragent/useragent_test.go
```diff
@@ -342,6 +359,8 @@ func TestObservedAgentSessions_NoSensitiveValuesReported(t *testing.T) {
 		"CODEX_THREAD_ID":                    sensitiveThreadID,
 		"CODEX_INTERNAL_ORIGINATOR_OVERRIDE": "Codex Desktop",
 		"CODEX_PERMISSION_PROFILE":           ":read-only",
+		"GROK_AGENT":                         "1",
```
johno-stripe, line 362: What's up with this, why is it just 1?

pkg/useragent/useragent_test.go
```diff
@@ -354,7 +373,7 @@ func TestObservedAgentSessions_NoSensitiveValuesReported(t *testing.T) {
 		DetectTerminalProgram(getEnv),
 	}
 
-	for _, secret := range []string{sensitiveSessionID, sensitiveHostID, sensitiveThreadID, sensitiveScopes} {
+	for _, secret := range []string{sensitiveSessionID, sensitiveHostID, sensitiveThreadID, sensitiveScopes, sensitiveGrokSessionID} {
```
johno-stripe: This kinda smells -- this generic test function didn't need any provider-specific params before, why does Grok require one now? Particularly given that it doesn't seem to be consumed?

pkg/useragent/useragent_test.go
```diff
@@ -322,10 +338,11 @@ func TestObservedAgentSessions(t *testing.T) {
...
 const (
-	sensitiveSessionID = "SESSIONID-1111"
-	sensitiveHostID    = "HOSTID-2222"
-	sensitiveThreadID  = "THREADID-3333"
-	sensitiveScopes    = "SCOPES-4444"
+	sensitiveSessionID     = "SESSIONID-1111"
+	sensitiveHostID        = "HOSTID-2222"
+	sensitiveThreadID      = "THREADID-3333"
+	sensitiveScopes        = "SCOPES-4444"
+	sensitiveGrokSessionID = "GROKSESSIONID-5555"
```
johno-stripe: Same as below, seems odd that we need this just for one provider

pkg/useragent/useragent.go
```diff
@@ -157,6 +161,15 @@ func DetectAgentHost(getEnv func(string) string) (kind string, raw string) {
 			return "terminal", "codex-cli"
 		}
 
+		// Grok Build's only documented surface is its terminal TUI (plus the
+		// xai-grok-shell integration, which runs inside it) -- no separate IDE
+		// extension, desktop app, SDK, or MCP server is described, and no host
+		// variable showed up alongside GROK_AGENT/GROK_SESSION_ID when checked
+		// against a real session. Revisit if Grok Build ever ships another surface.
+		if DetectAIAgent(getEnv) == "grok" {
```
johno-stripe, line 168: Does this [Grok Bot](https://x.ai/bot) support coding & plugins?  Do we need to cover that?

achen-stripe: Did a bit more digging because I thought it was terminal only at first, it does support connecting to IDEs and other tool integration via ACP ([here](https://docs.x.ai/build/cli/headless-scripting#acp)). I'l update to cover this

achen-stripe: Also to note It seems that agent detection isn't relying on this [fork](https://github.com/matv-stripe/detect-agent) since it's using this if statement instead, but it does seem to refer the env variables that Vercel uses in order to detect agents ([here](https://raw.githubusercontent.com/vercel/detect-agent/main/agents.json)). The main repo supports Grok, but the env vars they use aren't too helpful with determining what host is being used, which is why I used these two env vars instead which I found through testing.

pkg/useragent/useragent_test.go
achen-stripe: AI slop, just reused the generic session ID variable

pkg/useragent/useragent_test.go
achen-stripe: For testing, I just used the actual value of the env variable when Grok is active which is 1. The value itself doesn't matter too much though because DetectAgent only checks to make sure the value isn't an empty string.

pkg/agentsetup/grok.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	out, err := runOutput(ctx, GrokBinaryName, "plugin", "list", "--json")
+	if err != nil {
+		return grokInstalledPlugin{}, false, false
+	}
+	plugin, ok := findGrokStripePlugin(out)
+	return plugin, ok, true
+}
+
+// grokInstalledPlugin is an entry in `grok plugin list --json` output
```
johno-stripe, line 95: This name feels odd to me, I'd maybe have just gone with `grokPlugin`, but I know it's following the same pattern as the other files.

I don't think overhauling this is worth doing in this PR, but given how similar this code is to the same behavior in Claude, I wonder if there's room to extract more shared logic and shrink these files down to just config.  Would you mind comparing the different files and seeing how much of the behavior is shared, determine whether we can further shrink the per-provider code?

achen-stripe: I agree, the installation logic for a lot of these agents seem to mostly be the same (minus Cursor) other than the Apply() function since there's special cases with the exit codes. I'll put a note consolidating the logic as a follow up PR

vzhang-stripe (changes requested):
`pkg/cmd `tests fail when you run them inside an agent

  `clearAgentEnv` in `pkg/cmd/agent_env_test.go` needs to unset every var `DetectAIAgent` looks at, and it's missing the five new ones (`HERMES_AGENT`, `GROK_AGENT`, `GROK_SESSION_ID`,  `AI_AGENT`, `AGENT`). Claude Code sets `AI_AGENT`, so go test `./pkg/cmd/` fails there: `TestIsAIAgent_NotDetected`, `TestAIAgentHelp_NotDetected`, `TestRecordUnknownCommand_NotInAgentEnv`. It passes on master in the same shell. CI won't see this, but anyone running `make test` from Claude Code, Goose, or Amp will.

johno-stripe (approved):
Changeset looks reasonable to me as-is -- there are some patterns I'd consider adjusting, but better to leave that for a following PR, as it'd really broaden the scope of this change.

vzhang-stripe (changes requested):
ty!

vzhang-stripe (approved):

PR #2051, opened 2026-09-17
Subscribe to snapshot events in the listen forwarding canary

vzhang-stripe:
`stripe listen --forward-to` requires an explicit event subscription (--events, --all-snapshot, or --all-thin) since #2006, so the bare --forward-to in TestAPIListenForwardTo fails flag validation and the API canary has been red on every master push since. Subscribe with --all-snapshot to keep forwarding every snapshot event, as before.

The API canary only runs on pushes to master, which is why #2006's own PR checks did not catch this.

Committed-By-Agent: claude
Orbit-Session-Id: d0ad12e3-ac9c-47ae-b190-ae92d914e863

 ### Reviewers
r? @
cc @stripe/developer-products

 ### Summary


yahanxing-stripe (changes requested):
thanks for catching this! this has already been fixed in https://github.com/stripe/stripe-cli/pull/2015 test should pass after rebasing on latest master

PR #2058, opened 2026-09-18
Expose resumable, non-blocking OAuth login RPCs to plugins

vcheung-stripe:
### Summary

Adds rpcs for plugins (particularly projects) to initiate the OAuth flow themselves. The rpcs are:
- OAuthInitiateLogin: kicks off the OAuth flow and returns a device code and a verification URI
- OAuthCheckLoginStatus: checks whether the user has completed the OAuth flow. This doesn't poll, that's on the plugin to implement.
- OAuthFindPendingLogin: looks for a pending login that's been initiated with OAuthInitiateLogin, useful for --non-interactive flows where login is started with one command and then finished with another

### Testing

Tested login still works:

Screenshot 2026-09-18 at 4 33 50 PM

Introduced a dummy command in projects to test that it works from a plugin:

Screenshot 2026-09-18 at 4 34 03 PM


pkg/login/login.go
```diff
@@ -153,43 +225,115 @@ func PollForLogin(ctx context.Context, pollURL string, cfg *config.Config) error
...
+		}
+		// Terminal OAuth error (access_denied, expired_token, ...): the device code is dead.
+		clearPendingDeviceAuth()
+		return nil, err
+	}
+
+	result, err := saveDeviceCredentials(context.WithoutCancel(ctx), cont.AccessBaseURL, tokenResp, cfg)
+	if err != nil {
+		return nil, err
+	}
```
yahanxing-stripe, line 316: do we want to `clearPendingDeviceAuth()` on the error path too?

yahanxing-stripe (approved):

pkg/login/login.go
vcheung-stripe: Good catch 🙏

PR #2069, opened 2026-09-22
Report what auto-update is doing

yahanxing-stripe:
### Summary

Auto-update emitted three events and debug-logged everything else, so the two
failure modes that matter most were invisible: a check that can't reach GitHub, and a
panic in the update path.

| Change | Why |
|---|---|
| `install_method` detected via `pkg/installmethod` instead of hardcoded `"curl"` | The CLI's own analytics and the install script both report what `installmethod` reports. A second name for one install splits every dashboard grouping on it. |
| New `Auto-Update Check Failed`, with `reason=release_fetch\|asset_missing` | The GitHub API it calls is unauthenticated and rate-limited **per source IP**. A NAT'd or CI population can stop checking entirely, and today nothing shows it. |
| New `Auto-Update Skipped`, with `reason=major_version` | Makes eligible-but-not-updated countable. |
| `Auto-Update Failed` gains `reason=download\|status\|checksum\|extract\|replace` | Failures group by stage instead of by substring-matching error text. |
| `Auto-Update Succeeded` gains `staged_for_seconds` | How long a user sat on a version we already knew was old — the number that says whether auto-update keeps the fleet current. Comes from a new `staged_at` on the marker. |
| New `Auto-Update Panicked`, and `ApplyIfPending` now recovers | `ApplyIfPending` runs as the first statement of `main`, *before* the process-wide panic handler is installed — so a panic there took down the command the user typed before it could run. |
| Opt-out goes through `stripe.TelemetryOptedOut` | Same helper `cmd/stripe` uses, so the two can't drift. |

`staged_at` is `omitempty` and only reported when set, so a marker written by hand —
as the auto-upgrade canary does — stays valid. Event values render with sorted keys,
so the same fields always produce the same string regardless of map iteration order.

### Test plan

- `go build ./...`, `go vet`, `gofmt`, `go run ./internal/lint/errorcategory/cmd ./...`
- `go test -race ./pkg/autoupdate/` — new cases: deterministic `event_value`, `staged_at`
  round-trip, a hand-written marker without `staged_at` still applying, `recoverAndReport`
  swallowing a panic, and failure `reason` assertions on the existing download tests
- Events captured end to end against a local collector via `STRIPE_TELEMETRY_URL`:

```
Auto-Update Available        | install_method=script | from=1.51.0 to=1.51.1
Auto-Update Succeeded        | install_method=script | from=1.51.0 staged_for_seconds=1 to=1.51.1
Auto-Update Failed           | error=checksum verification failed from=1.51.0 reason=checksum to=1.51.1
Auto-Update Check Failed     | reason=release_fetch
Auto-Update Skipped          | from=0.0.1 reason=major_version to=1.51.1
```


vzhang-stripe (approved):
looks good

